PR #58, opened 2026-04-30
feat(retrieval): port adjacent_neighbors lever from benchmark to core

jaylfc:
## Summary

The dominant LoCoMo lever — `adjacent_turns` (+0.089 on same-tier per `docs/benchmarks.md`) — lived only in `benchmarks/locomo_runner.py`. A fresh \`pip install\` of taosmd never got the benefit. This PR lifts the same idea into the public \`retrieve()\` API as \`adjacent_neighbors\`, default off, fully backwards-compatible.

## What changes

- **\`VectorMemory.get_by_position()\`** — looks up rows by \`metadata[position_key]\` via SQLite \`json_extract\`, with optional \`group_key\`/\`group_value\` to confine neighbours to the same conversation/session/document.
- **\`retrieve()\`** gains \`adjacent_neighbors\` / \`position_key\` / \`group_key\` parameters. When \`N>0\`, after rerank/verification each surviving hit gets a \`neighbors\` list attached, populated by ±N positional lookups against the vector source. Primary hits and already-attached rows are de-duplicated.
- **Consumers opt in** by populating an integer position field at ingest time, e.g. \`metadata={\"position\": turn_idx, \"session\": \"conv1\"}\`.
- **README** gains a short Quick Start example surfacing the new lever.
- **\`docs/benchmarks.md\`** gets a one-line pointer to the new API in the architectural-findings section.

## What does NOT change

- Default \`adjacent_neighbors=0\` — opt-in, zero behaviour change for existing users.
- \`benchmarks/locomo_runner.py\` still uses its own \`_build_adjacent_map\`, so the LoCoMo leader recipe (0.557 same-tier) reproduces identically. A follow-up can swap the benchmark to the core path after re-measuring parity.
- The README's 97.0% LongMemEval-S headline is on a different runner path and isn't affected by this lever.

## Why ship now

Original gate was \"wait for qwen9b + MoE runs\". qwen9b confirmed +0.089 stable across 5B and 9B; MoE (qwen3.6-35B-A3B) is still pending. Decided to ship behind an opt-in flag rather than block on MoE because the surface is purely additive and doesn't affect anyone who doesn't enable it.

## Test plan

- [x] \`pytest tests/ -q\` — 109 passing (12 existing retrieval + 8 new + 89 elsewhere)
- [x] SQLite path smoke-tested against a real \`VectorMemory\` instance with multi-session data — group filtering, out-of-range, and missing-match all behave correctly
- [x] Existing benchmark code unchanged — LMEKU Phase A and the future TurboQuant runs are unaffected
- [ ] Follow-up: swap \`benchmarks/locomo_runner.py\` to the core path and verify 0.557 reproduces

## Out of scope

Default-on for \`adjacent_neighbors\` and the broader hardware-tier default-model matrix (4 GB / 8 GB / 12 GB / Pi NPU / CPU-only × generator / embedder / reranker / extractor / librarian) are deferred until both the 1050 Ti benchmark row and the qwen3.6-MoE row are in \`docs/benchmarks.md\`.

## Summary by CodeRabbit

* **New Features**
  * Opt-in positional neighbor retrieval for multi-turn conversations: store integer turn positions at ingest and request adjacent neighbors with optional grouping.

* **Documentation**
  * README updated with usage examples for positional retrieval; benchmarks updated to show few-shot variant and impact of adjacent-turns.

* **Tests**
  * Added comprehensive unit and integration tests covering neighbor attachment, grouping, boundaries, and end-to-end retrieval.


jaylfc:
Thanks for the review. Pushed `d45a1e0` addressing the two real comments:

**1. README wording (line ~260)** — softened "Each hit gains" → "Each hit may include", with a parenthetical noting the conditions under which neighbours are skipped (boundaries, primary-hit overlap, missing position/group keys).

**2. Cross-group leakage when `group_key` is set but missing on a hit** — confirmed bug. `_attach_neighbors` was building primary keys of `(None, pos)` and calling `get_by_position` with `group_value=None`; that helper drops the group filter when `group_value is None`, so a group-less hit was quietly pulling same-position rows from any group. Now: when `group_key` is configured, hits without that key are skipped at both the primary-key gather and the neighbour scan, mirroring the existing `position_key`-missing behaviour. Added a regression test (two sessions × shared positions × one stray hit lacking the session key) that asserts no neighbours get attached. 110 tests pass.

**3. SQLite expression indexes for `get_by_position` (nitpick)** — deferring. Fair point on full-scan cost, but: (a) `metadata_json` is a JSON blob, so an expression index is per-position-key and per-group-key — adding them blindly bloats the schema; (b) the lever is opt-in and we don't yet have a profile showing the JSON path is hot. Happy to revisit in a follow-up PR if it shows up in real workloads.

**4. Docstring coverage warning (72.73%)** — all newly added functions have docstrings; the gap is in pre-existing helpers in the modified files. Will tidy in a separate sweep so this PR stays focused on the lever port.

PR #61, opened 2026-05-01
fix(rescore): handle uncategorised results without dividing by zero

jaylfc:
## Summary

\`benchmarks/locomo_rescore_streaming.py\` crashed at the very end of every LongMemEval-KU rescore with \`ZeroDivisionError: division by zero\`. The script bucketed results by LoCoMo's integer category (1-4) and accumulated \`tot[\"count\"] += len(rs)\` inside the per-category loop. LMEKU results have no category field (always \`None\`), so the loop ran zero iterations, \`tot[\"count\"]\` stayed at 0, and \`all_f1 = tot[\"f1\"] / n\` raised — after all the heavy external-judge work was already done.

This bit LMEKU Phase A Cell 1 (\`lmeku_baseline-vector_gemma4_e2b\`) earlier today: bench finished, all 78 \`judge_rejudged\` values landed in the JSON, but no scorecard printed.

## What changes

- **Fallback bucket** when no LoCoMo categories match: print one Overall row using all results so LMEKU and any future non-LoCoMo benchmark using the same rescore path still produces a usable number.
- **Empty results** print \`(no results)\` instead of crashing.
- Extracted the per-bucket computation into \`_row_stats()\` so the LoCoMo and fallback paths share the same code.

## What does NOT change

- LoCoMo per-category breakdown is byte-for-byte unchanged (all 4 category lines plus Overall, same numbers, same formatting).
- The actual rescore work (LLM calls + checkpoint writes) was already correct — only the final scorecard print path is touched.

## Recovered number

Cell 1 already had all 78 rejudge values in its \`rescored_v2.json\`, so its number is recoverable today:

> **LongMemEval-KU baseline-vector / gemma4:e2b → 0.5385 ext-judge (qwen3:4b)**

The other 5 cells in Phase A weren't run — the chain was killed because the \`vector+kg\` and \`full+supersede\` configs were taking ~45-100 min/QA on ingest (the KG-extraction LLM-call volume), which would push Phase A to ~80h+ instead of the original ~4.5h estimate. A redesigned Phase A is for a separate session.

## Test plan

- [x] 4 new regression tests in \`tests/test_locomo_rescore_streaming.py\`: LoCoMo with categories (positive), LMEKU shape with no categories (regression), empty results, partial rescore coverage
- [x] Full suite: 127 passing (123 prior + 4 new)
- [x] Verified Cell 1 number: \`statistics.mean\` of \`judge_rejudged\` over the existing \`rescored_v2.json\` = 0.5385

## Summary by CodeRabbit

* **Bug Fixes**
  * Scorecard output now includes uncategorized rows in overall totals, avoids errors on empty datasets, and shows correct per-category statistics and coverage counts (e.g., "1/2") even with partial rescoring.
* **Refactor**
  * Centralized per-set metric computation to ensure consistent and accurate displayed values and totals.
* **Tests**
  * Added tests verifying scorecard formatting and robustness across categorized, uncategorized, mixed, partial-rescore, and empty-result scenarios.

### DIFF
--- a/benchmarks/locomo_rescore_streaming.py
+++ b/benchmarks/locomo_rescore_streaming.py
@@ -167,7 +167,7 @@ def print_scorecard(path: Path, data: dict, judge_model: str) -> None:
     results = data["results"]
     print()
     print("=" * 88)
-    print(f"LoCoMo Rescore (streaming) — {path.name}  [judge: {judge_model}]")
+    print(f"Rescore (streaming) — {path.name}  [judge: {judge_model}]")
     print("=" * 88)
     print(f"{'Category':<16} {'Count':>6} {'F1':>8} {'Orig':>8} {'Tol':>8} {'Rejudge':>10} {'Delta':>8}  Covered")
     print("-" * 88)
--- a/benchmarks/locomo_rescore_streaming.py
+++ b/benchmarks/locomo_rescore_streaming.py
@@ -145,7 +145,25 @@ def write_checkpoint(path: Path, data: dict) -> None:
     tmp.replace(path)
 
 
+def _row_stats(rs: list[dict]) -> dict:
+    """Compute per-bucket metrics from a list of rescored results."""
+    f1 = statistics.mean(r["f1"] for r in rs)
+    orig = statistics.mean(r["judge"] for r in rs)
+    tol = statistics.mean(r.get("judge_tolerant", r["judge"]) for r in rs)
+    rv = [r["judge_rejudged"] for r in rs if r.get("judge_rejudged") is not None]
+    rj = statistics.mean(rv) if rv else float("nan")
+    delta = (rj - orig) if rv else 0.0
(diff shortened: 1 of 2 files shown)


jaylfc:
Pushed \`d11b5f5\` addressing the mixed-categories finding. Real edge case — when the dataset has some rows with \`category in {1,2,3,4}\` and some without, the previous fix silently dropped the uncategorised rows from Overall. Now folds them in before computing the totals, with a regression test (2 categorised + 2 uncategorised → Overall count = 4, not 2). Pure-LMEKU and pure-LoCoMo paths are unchanged.

Note: a flaky failure in \`test_agents.py::test_reflect_gate_disabled_returns_empty\` showed up once when running the full suite (passes on isolated re-run, passes on full re-run). That's pre-existing on master — \`tests/test_api.py\` from #60 holds onto SQLite handles in its fixture cleanup that occasionally leak into \`test_agents.py\`. Not introduced here. Worth a separate small PR.

PR #62, opened 2026-05-02
feat(bench): --llm-backend ollama|llama-server for TurboQuant runs

jaylfc:
## Summary

Phase 1 of the [TurboQuant benchmark spec](docs/specs/2026-04-29-qwen36-turboquant-benchmark-design.md). Wires \`locomo_runner.py\` so a single \`--llm-backend\` flag swaps the generator between Ollama (default, today) and llama-server (TurboQuant fork, for Qwen3.6-35B-A3B with CPU-MoE offload).

The qwen3.6-35B-A3B model only fits on the 3060 12 GB card with TurboQuant KV-cache offload, which lives in the \`TheTom/llama-cpp-turboquant\` fork — not in Ollama. That fork serves OpenAI-compatible \`/v1/chat/completions\`; this PR teaches the runner to speak it.

## What changed

- **\`_llama_server_generate()\`** — new function. POSTs to \`{url}/v1/chat/completions\` with \`messages=[{\"role\":\"user\",\"content\":prompt}]\` and reads \`choices[0].message.content\`. Honours \`temperature\`, \`thinking_mode\`, and \`no_think_prefix\` (the same \`/no_think\` prefix works inside the llama-server chat template).
- **\`_generate(backend, ...)\`** — dispatcher. Routes to \`_ollama_generate\` (default) or \`_llama_server_generate\`.
- **\`_judge\`** — now goes through the dispatcher, defaults to ollama. The in-runner self-judge stays on the same backend as the generator so a llama-server-only model isn't required to also serve via Ollama just for self-judging.
- **\`_process_qa\`** — gains \`llm_backend\` and \`llama_server_url\` params; \`gen_url\` resolved once at the top so both retrieval and full-context branches reach a defined value (the naïve draft put it inside the \`else\` and would NameError on \`--full-context\`).
- **HyDE and main answer callsites** — switched from \`_ollama_generate\` to \`_generate(llm_backend, client, gen_url, ...)\`.
- **New CLI flags**: \`--llm-backend {ollama,llama-server}\` (default \`ollama\`), \`--llm-server-url\` (default \`http://localhost:8085\`).
- **JSON meta** records \`llm_backend\` and \`llm_server_url\` for reproducibility.

## What does NOT change

- Default behaviour: every existing chain script that doesn't set \`--llm-backend\` keeps hitting Ollama exactly as before.
- \`_decompose_query\` still calls \`_ollama_generate\` directly — uses \`gemma4:e2b\` as a small utility model, kept on Ollama regardless of the main generator's backend.
- The external rescore (\`locomo_rescore_streaming.py\`) is untouched; \`qwen3:4b\` external judge always runs on Ollama.

## Status of the surrounding TurboQuant work

- ✅ Disk freed for it (~198 GB reclaimed in two cleanup passes today)
- ✅ Qwen3.6-35B-A3B-UD-Q4_K_M GGUF downloaded (21 GB)
- 🔄 \`llama-cpp-turboquant\` Docker build in progress on Fedora (at 31% — \`turbo2_0/turbo3_0/turbo4_0\` template instances compiling, so the TurboQuant cache types are still in this fork)
- ⏳ Once build finishes: verify \`--cache-type-k turbo3\` in \`--help\`, smoke test \`--limit 5 --llm-backend llama-server\` against a running llama-server, then wait-and-run the full bench after thinking_on subset frees the GPU

## Test plan

- [x] \`python3 -c \"import ast; ast.parse(...)\"\` — syntax OK
- [x] \`pytest tests/ -q\` — 128 passing (no regression on the retrieval/api/rescore surface)
- [x] \`python3 benchmarks/locomo_runner.py --help\` shows both new flags
- [ ] Smoke test \`--limit 5 --llm-backend llama-server\` once llama-server is built + running
- [ ] Phase 2 baseline + Phase 3 full-stack runs land in a follow-up commit with the actual numbers

## Summary by CodeRabbit

* **New Features**
  * Added support for an alternate LLM backend (llama-server) alongside the existing Ollama backend for generation and evaluation
  * New CLI option to select an expansion model via --expansion-model

* **Improvements**
  * CLI now exposes backend selection and server URL; run metadata records backend, server URL, and expansion model
  * Self-judging and generation consistently use the configured backend

* **Bug Fixes**
  * Multihop decomposition failures now emit an error message instead of failing silently

### DIFF
--- a/benchmarks/locomo_runner.py
+++ b/benchmarks/locomo_runner.py
@@ -193,11 +193,69 @@ async def _ollama_generate(client: httpx.AsyncClient, url: str, model: str,
     return (resp.json().get("response") or "").strip()
 
 
+async def _llama_server_generate(client: httpx.AsyncClient, url: str, model: str,
+                                 prompt: str, temperature: float = 0.2,
+                                 thinking_mode: bool = False,
+                                 no_think_prefix: bool = False) -> str:
+    """OpenAI-compatible generator for llama-server (TurboQuant fork).
+
+    Used when ``--llm-backend llama-server`` is set so Qwen3.6-35B-A3B can
+    serve answers via llama.cpp's chat completions endpoint (the model
+    doesn't fit in Ollama at this quant on a 12 GB card without TurboQuant
+    KV-cache offload). ``thinking_mode`` and ``no_think_prefix`` keep their
+    semantics: with ``no_think_prefix`` we prepend ``/no_think`` so the
+    chat template can suppress reasoning tokens; with ``thinking_mode`` we
+    leave both off and let the generator emit CoT.
+    """
+    payload_prompt = ("/no_think\n\n" + prompt) if no_think_prefix else prompt
+    payload = {
+        "model": model,
+        "messages": [{"role": "user", "content": payload_prompt}],
+        "temperature": temperature,
+        "stream": False,
+    }
+    resp = await client.post(f"{url}/v1/chat/completions", json=payload)
+    resp.raise_for_status()
+    body = resp.json()
+    choices = body.get("choices") or []
+    if not choices:
+        return ""
+    msg = choices[0].get("message") or {}
+    return (msg.get("content") or "").strip()
+
+
+async def _generate(backend: str, client: httpx.AsyncClient, url: str, model: str,
+                    prompt: str, temperature: float = 0.2,
+                    thinking_mode: bool = False,
+                    no_think_prefix: bool = False) -> str:
+    """Dispatch a generation call to the configured backend.
+
+    ``backend`` is either ``"ollama"`` (default — POSTs to ``{url}/api/generate``)
+    or ``"llama-server"`` (TurboQuant fork — POSTs to ``{url}/v1/chat/completions``).
+    Caller supplies the right URL for the chosen backend; the runner picks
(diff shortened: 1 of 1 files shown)


jaylfc:
Pushed \`a53c47c\` addressing the major finding. Real bug — confirmed:

- With \`--llm-backend=llama-server --llm-query-expansion\`, the expansion call would have hit Ollama with the GGUF model name (e.g. \`Qwen3.6-35B-A3B-UD-Q4_K_M\`) which Ollama doesn't have, then failed inside \`expand_query_llm\`, then been swallowed by the bare \`except: pass\`. The bench would have run with **no expansion** and the user would have had no signal.

Fixes:

- New \`--expansion-model\` arg (default empty → falls back to \`--model\` for back-compat). Required when \`--llm-backend=llama-server\` since \`expand_query_llm\` only speaks Ollama's \`/api/generate\`. Recorded in JSON meta.
- Both bare \`except Exception: pass\` blocks (query expansion + multihop decompose) replaced with explicit \`print(..., file=sys.stderr)\` including the offending URL/model so misconfigurations are visible in the bench log instead of silently degraded.

Pre-existing chains that don't set the new flag get **identical** Ollama call shape as before. 128 tests still pass.

PR #162, opened 2026-06-13
feat(embed): arctic-embed-s as the low-tier dense default (ships F-010)

jaylfc:
## What

Ships the confirmed E-007/F-010 win: the low-tier dense embedder default moves all-MiniLM-L6-v2 to Snowflake arctic-embed-s. Full-1540 judged +0.0565 (0.7305 vs 0.6740), +0.157 R@K, same 384 dim and latency.

## Changes (the 5-place ship-the-win, via the recipe-aware store seam from #161)

- `recipes.py`: new `embed_model` field; `lite-pi` and `fast-8b` set `arctic-embed-s`. Higher GPU tiers keep MiniLM pending their own recipe-level tri-judge re-run (honest scope: E-007 confirmed the dense path, not the rerank/late-interaction stacks).
- `config.vector_memory.embed_model` is the store-level embedder; `_ensure_stores` + `_resolve_onnx_path` honor it (configurable model dir, finds arctic's `onnx/` subdir layout); `auto_setup` seeds it from `recommend()`.
- Store-mode guard now includes the **embedder identity** in the signature: switching embedder on an existing store (different vector space) fails loud with a re-embed instruction. Backward-compatible with #161 markers (component-wise compare, legacy markers upgraded, never spuriously rejected).
- `scripts/setup.sh` fetches arctic-embed-s ONNX (mirrors the MiniLM fetch).
- `docs/benchmarks.md`: low-tier default section with the F-010 numbers.

## Coordination

@taOS asked to register arctic-embed-s in the model store and verify the download path end to end (bus msg 417): a model taOSmd recommends is not shipped until its store download path is verified.

## Remaining before merge

- Small Pi-CPU arctic speed sample (latency data point, gentle per the PSU rule).
- @taOS store-path verification.

## Tests

20 new/updated across store-guard + recipe-aware + embed_model wiring; full suite 862 green; my diff ruff-clean (pre-existing E402s in recipes.py untouched). Dense MiniLM installs unchanged (default embed_model stays minilm-onnx unless a recipe seeds arctic).

## Release Notes

* **New Features**
  * Low-tier recipes now use an improved embedder with enhanced recall performance
  * Added support for configurable embedding models per data directory

* **Documentation**
  * Documented the updated low-tier dense embedder and associated performance gains

* **Tests**
  * Enhanced test coverage for embedding model configuration and recipe defaults


jaylfc:
Pi-CPU speed sample (gentle, n=200, niced, aarch64 Orange Pi 5 Plus, no GPU): arctic-embed-s **13.0ms p50 / 20.6ms p95 / 13.9ms mean per embed**, 384-dim. Realtime-viable: embedding is negligible against the sub-300ms search SLA, so the low-tier default swap costs no meaningful latency on the Pi. That closes the Pi-speed-sample item; remaining before merge is @taOS's store-path verification (acked, in progress).

PR #172, opened 2026-06-21
Explorer: time scrubber (memory as of T) - Phase 2c

jaylfc:
Drag a time slider in the Memory Explorer and watch the knowledge graph reconstruct itself as it stood at any past instant. This completes the temporal-provenance story the Explorer started: facts appear when first learned, and facts that were later replaced fade out of the snapshot.

**Backend:** `kg.graph(limit, as_of=None)` gains an optional `as_of` (unix time). When set, the degree ranking, edge set, and totals are computed over only the triples whose validity window contains that instant (`valid_from <= as_of AND (valid_to IS NULL OR valid_to >= as_of)`) - the same point-in-time predicate `query_entity` already uses, so no new storage or events. The `active` flag keeps meaning current-today, so a fact that was live at `as_of` but has since been superseded is included and rendered faded, which keeps the All/Current toggle meaningful in scrubber mode. `graph()` now also returns `t_min`/`t_max` (earliest `valid_from` and the latest temporal point) so the client builds the slider range without a second call. Threaded through api/service/remote and `GET /graph?as_of=` (400 on a non-numeric value).

**Frontend:** a debounced time slider under the All/Current toggle (only shown when the history spans more than one instant), a localized date label, an 'as of <date>' header badge while scrubbing, and a Now button to return to the live view. Native `<input type=range>` with `aria-valuetext` for accessibility; dragging fully right returns to live.

Builds on the merged Memory Explorer (Phase 2a/2b). Full suite 973 pass; new tests cover the as-of snapshot reconstruction, the active-flag semantics, t_min/t_max bounds, and the endpoint plus its 400. Spec: docs/superpowers/specs/2026-06-21-dashboard-phase2c-time-scrubber.md.

Out of scope (later polish if wanted): timeline auto-play, per-edge valid-from/to tooltips, two-instant diff.


jaylfc:
Superseded by the temporal memory view merged in 94a896c (PR #195 era work, branch feat/temporal-memory-view).

That change landed the same core design this PR proposed: `kg.graph(limit, as_of=None)` filtering on the existing valid-time window, threaded through api/service/remote, exposed as `GET /graph?as_of=`, with an Explorer time scrubber and a return-to-now affordance. It also covers the two items listed here as out of scope: timeline auto-play (a playback control that steps earliest to now, disabled under prefers-reduced-motion) and a memories-per-day density backdrop under the track. Accessibility went a bit further too, with a full `role=slider` implementation (arrow keys step a day, PageUp/PageDown a week, Home/End to the ends) rather than a native range input.

Closing as superseded rather than merging, since the two implementations touch the same files and the merged one is a superset.

One thing this PR did better and worth carrying over: it returns 400 on a non-numeric `as_of`, where the merged version silently falls back to the live graph. An adversarial review of the merged code flagged the same gap from the other direction (nan and inf are numeric, so they pass the float parse and reach SQL). Following up with a finite-value guard so the endpoint matches its documented intent.

Thanks for the design write-up in docs/superpowers/specs/2026-06-21-dashboard-phase2c-time-scrubber.md, it stays useful as the record of how this was reasoned through.

PR #185, opened 2026-07-06
Add task edge read endpoint

021-lab:
## Summary
- add GET /tasks/edges for active task edge reads
- support from_id, to_id, type, and limit filters
- add service/remote wrappers and focused task/http tests

## Tests
- .venv/bin/python -m pytest tests/test_tasks.py tests/test_http_server.py -q

## Summary by CodeRabbit

* **New Features**
  * Added support for viewing task-graph edges through a new `GET /tasks/edges` endpoint.
  * Edge queries now support filtering by source, destination, edge type, and result limit.

* **Bug Fixes**
  * Active edges are returned consistently, excluding removed entries from results.

* **Chores**
  * Updated ignore rules to exclude the Python virtual environment directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ benchmarks/data/*
 models/minilm-onnx/model.onnx
 models/cross-encoder-onnx/
 .env
+.venv/
 *.log
 eval/reports/
 .superpowers/
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -88,6 +88,7 @@ Endpoints
 ``GET  /a2a/members``      ``?channel=<name>``             -> ``{"members": [...]}``
 ``POST /tasks``            ``{"title", "body"?, "project"?, "assignee"?, "priority"?, "depends_on"?: [...], "created_by"}`` -> task object
 ``GET  /tasks``            ``?status=&project=&assignee=&limit=``  -> ``{"tasks": [...]}``
+``GET  /tasks/edges``      ``?from_id=&to_id=&type=&limit=``       -> ``{"edges": [...]}``
 ``GET  /tasks/ready``      ``?project=&assignee=&limit=``  -> ``{"tasks": [...]}``
 ``GET  /tasks/prime``      ``?project=&assignee=``         -> ``{"text": ..., "tasks": [...]}``
 ``POST /tasks/{id}``       ``{"status"?, "assignee"?, "priority"?, "body"?}`` -> updated task object
@@ -826,6 +827,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     self._handle_task_create()
                 elif method == "GET" and path == "/tasks":
                     self._handle_task_list(query)
+                elif method == "GET" and path == "/tasks/edges":
+                    self._handle_task_list_edges(query)
                 elif method == "GET" and path == "/tasks/ready":
                     self._handle_task_ready(query)
                 elif method == "GET" and path == "/tasks/prime":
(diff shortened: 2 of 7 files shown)


jaylfc:
Thanks for this, it's a genuinely tidy PR. The layering matches how the rest of the codebase is built, the SQL is parameterized with the type allowlist checked up front, and the tombstone handling lines up with remove_edge, so nice work reading the conventions.

One blocker before I can merge: the handler needs to go through _apply_token_binding the way _handle_task_list does, so a verified registry token gets its grant checked and a project-bound token can't read edges belonging to other projects. Since task_edges has no project column you'll want to join to tasks to scope by the bound project. The edge write endpoints have the same gap on our side and a fix is planned, so I'd rather not add another instance via a read path.

Could you also add a short CHANGELOG entry, and ideally a bounds check on limit (negative means unlimited in SQLite) plus an HTTP test for an invalid type? The .venv ignore line is fine to keep.

jaylfc:
Quick update: the scoping infrastructure just landed on master (#186 for the edge write endpoints, #187 for task update), so there's now a ready-made path for this PR. After a rebase you can route the handler through _apply_token_binding and use service.task_projects to filter the returned edges to the token's bound project (see _enforce_edge_project_scope in http_server.py for the non-enumerating pattern). Happy to merge once that's in along with the CHANGELOG line.

jaylfc:
Thank you for this, and apologies for the long wait — the delay is on us, not on the contribution. The endpoint is a genuine gap and the implementation is clean: sensible filters, an explicit `limit` with a proper 400 on a non-integer, service and remote wrappers kept in step, and focused tests. I want it.

**One blocking issue, and it is a matter of timing rather than of your code.**

This PR was opened on 2026-07-06. Two security fixes landed shortly afterwards that changed the contract for exactly this family of endpoints:

- `25b7d22` — *fix(tasks): apply token binding and project scoping to edge endpoints (#186)*
- `53ed486` — *fix(tasks): apply project scoping to the task update endpoint (#187)*

Since those, every task-edge surface derives its scope **from the token**, and a token bound to a project may only see that project's tasks. Foreign and nonexistent ids deliberately return the **identical 403**, so that task existence cannot be probed by an unauthorised caller. On master today that is enforced through `self._token_project` and `_enforce_edge_project_scope`.

`_handle_task_list_edges` as written does none of that. It reads `from_id`, `to_id`, `type` and `limit` and calls straight through to `service.task_list_edges`, passing no token or project context at all — so no scoping is possible downstream either. The practical effect is that a token bound to one project can enumerate the **entire** task graph across every project, including ids it must not learn exist. That is precisely the hole #186 closed, reintroduced through a new door.

To be clear about blame: nothing here was careless. The endpoint simply predates the posture, and there was no way for you to write against a contract that did not exist yet.

**What it needs**

1. Derive scope from `self._token_project`, matching the sibling edge handlers.
2. When the token is project-bound, restrict results to that project's tasks. Return the same non-enumerating `403` for foreign or unknown ids rather than an empty list or a `404` — an empty result is itself an existence oracle when the caller can compare it against a known-good id.
3. Leave the tokenless / standalone / global-token path (`token_project is None`) untouched, exactly as the existing handlers do.
4. A test asserting a project-bound token **cannot** see another project's edges, and that a foreign id and a nonexistent id are indistinguishable in the response.

`taosmd/http_server.py` around the `_handle_task_*` edge handlers is the pattern to copy; the comment above the `depends_on` scoping in the create path explains the reasoning in full, including why scope keys on the token-bound project rather than a body-supplied tag.

**How to proceed**

Please open this as a **new PR from a fresh branch** rather than pushing to this one. Our build tooling constructs a fresh worktree and cannot operate on an existing branch, so a follow-up on this branch would be unworkable on our side. Reference this PR number so the history stays followable, and I will close this one with a pointer once the replacement is up.

Happy to review promptly this time — and if you would rather not carry the scoping work, say so and I will card it internally with full credit to you for the endpoint design and the tests.

jaylfc:
**Correction to my review above — please disregard the "open a new PR" request. Keep this branch.**

I asked you to abandon this branch and open a fresh PR, explaining that "our build tooling constructs a fresh worktree and cannot operate on an existing branch." That limitation is real, but it applies only to **our internal automated build lanes**. You are pushing your own commits from your own branch and never touch that tooling, so none of it constrains you.

I imposed an internal constraint on an outside contributor for no benefit. Push the scoping changes to this branch as you normally would, and this PR merges as-is once they land.

Everything else in the review stands unchanged — the token-derived scoping, the non-enumerating `403` for foreign and unknown ids alike, leaving the `token_project is None` path untouched, and the test that a project-bound token cannot see another project's edges.

My thanks for the patience; you have now waited on us twice.

jaylfc:
Checked this against master again today so the trail doesn't go cold. Still want it.

Current state: the branch now conflicts with master in three places, but all three are the boring kind where both sides added adjacent lines. The endpoint list in the http_server.py help text, `__all__` in service.py, and `list_edges` landing next to the new `get_task_projects` in tasks.py. A few minutes of rebase, nothing structural. Every name from the July review is unchanged on master (`_apply_token_binding`, `_token_project`, `_enforce_edge_project_scope`, `service.task_projects`), so that guidance still applies as written.

You have "Allow edits from maintainers" enabled, which gives us a clean way to finish this. If you're short on time, I'll push the rebase and the scoping commit to this branch myself and merge, with your commits left exactly as they are. If I haven't heard from you by September 1st I'll go ahead and do that rather than let the PR sit through a third stretch, since the waiting has been on our side both times. If you'd rather finish it yourself, say so and it's all yours.

jaylfc:
Thanks for this, and sorry it sat so long. GET /tasks/edges ended up landing on master through #415, which also scopes it by project, so I'm closing this one rather than leaving it open with conflicts. Your PR was the first to put the endpoint on the table, and it's appreciated.

PR #195, opened 2026-07-19
feat(collections): Phase 1 docs collections (create/index/link/grants, wires loaders + gitignore walker)

jaylfc:
## What

Phase 1 of collections (the docs MVP from `docs/specs/codebase-indexing-collections-design.md`): named containers of content indexed from a folder, queryable by granted agents alongside conversation memory. This wires the previously-unwired loader framework (`taosmd/loaders/`) into a real ingest path and adds the full container lifecycle, access control, and query surface. **HELD FOR REVIEW: do not merge without sign-off.**

## Why

Codebase indexing is table stakes for coding agents; docs collections ship first (per the Jul 19 decision) to get real usage on the container model before the Phase 2 code path. Everything is additive and off by default: with no `collections.allowed_roots` configured, nothing changes for existing installs.

## Binding decisions implemented (Jul 19)

- No tree-sitter, no new dependencies; docs-shaped files only (DocLoader md/txt/markdown/rst + other registered loaders where they claim a file); zero-dep paragraph chunker with `# upgrade-path:` comments.
- Admin-token gating (fail closed) on `POST /collections` and `POST /collections/{id}/index`; `DELETE /collections/{id}` (archive) is admin too. Data plane (list/get/link/unlink/grants/query) uses the normal token + grants.
- Per-collection `embedder` field: stored at create, returned in GET, read by the index path; Phase 1 always indexes with the global default.
- Grants table `(canonical_id, scope='collection', collection_id)` UNIQUE together, stored in taosmd, enforced on search per requesting agent; no transitivity (project links never grant access).
- Typed link rows `{type: taos|git, id}`, multiple per collection, link/unlink metadata-only.
- Zero-loss: DELETE = archive (reversible status change, content hidden from query, nothing destroyed); re-index supersedes changed/deleted-file rows via the existing `valid_to` machinery.
- Filesystem safety: `collections.allowed_roots` config list (default EMPTY = feature off), `resolve_within` containment at create and at every index, symlink escapes rejected, `check_size` per file.

## How it works

- `taosmd/collections.py`: sqlite store (`collections.db`) for rows/links/grants + per-file content hashes; gitignore-aware walker (simplified stdlib fnmatch rules, VCS/dependency/hidden dirs, binary sniff, size cap); `ingest_folder` chunks and routes through `api.ingest_batch` under the collection id as agent namespace, so per-chunk content-hash ids dedup unchanged files and the archive keeps every version.
- Content rows live under the collection id (which matches the agent-name grammar by construction), so search scoping is the existing `search_agents` mechanism: `api.search` gains `collections`/`collections_only` and extends the agent set with granted, non-archived collections only.
- Indexing over HTTP is async: 202 + `{"status": "indexing", "job": id}`, poll `GET /collections/{id}` until `ready|error`; the walk runs via a new `_ServiceLoop.spawn` so store access stays on the single service-loop thread and HTTP never blocks.

## Endpoints

```
POST   /collections                      (admin)  create
GET    /collections [?project=]                   list (project matches links of either type)
GET    /collections/{id}                          full row: status/stats/links/grants
POST   /collections/{id}/index           (admin)  202 + poll
POST   /collections/{id}/link|unlink              {"type": "taos"|"git", "id"}
POST   /collections/{id}/grants                   {"agent"}
DELETE /collections/{id}/grants/{agent}           revoke
DELETE /collections/{id}                 (admin)  archive (reversible)
POST/GET /search                                  + collection / collections / collections_only
```

Plus CLI `taosmd collections list|create|index|link|unlink|grant|revoke` and MCP `memory_list_collections` + `collection` on `memory_search`.

## Tests

- Full suite green: **1178 passed** (1110 before this branch; **68 new tests** across config roots, store, walker/chunker/ingest, HTTP, CLI, MCP). The known `test_admin_surface` random-order flake did not trip.

## Eval (pre-registered, section 5 of the spec)

`benchmarks/collections_eval.py` + a 20-question gold-file set over the repo's own `docs/` (tracked in `benchmarks/data/collections_eval_questions.json`, written before the index was built). No local ONNX model exists on this dev host, so the smoke ran the **lexical (BM25) fallback arm**, which stubs the embedder only so rows land and retrieves through the engine's `mode="bm25"` path (full walk/chunk/ingest/grant/scope path exercised, vectors never consulted):

- **Lexical smoke: file-level Recall@5 = 19/20 = 0.950** (47 files / 493 chunks indexed in 3.1 s). The one miss is the LoCoMo-scorecards question ("LoCoMo" appears across many docs).
- The semantic arm (real ONNX embedder) is the number the kill bar (>= 0.8) judges; run `--mode semantic` on the Fedora bench host.

## Spec deviations (flagged)

1. `source_path` is **required** on create (spec body marked it optional). The docs MVP has nothing to do with a source-less collection; making it required keeps `create` from minting rows `index` must then reject.
2. Create takes no `project_id` body field; project attachment is exclusively via the typed `link` endpoint (one mechanism instead of two, matches the typed-links decision).
3. A fourth table, `collection_files` (per-file content hashes), backs incremental re-index; storing the hash map in `stats` JSON would have made stats writes racy with indexing.
4. Statuses are `created|indexing|ready|error|archived` (decisions list) rather than the spec section-2 `empty|...` wording.
5. Service-layer collections ops are local-only (no `RemoteClient` forwarding), matching the shelf admin wrappers: the server that owns the indexed filesystem runs the ops. Remote-configured CLI installs would need a follow-up to proxy them.
6. `api._format_hit` now unwraps to the innermost user metadata so the semantic path exposes the same metadata contract as the BM25 path for batch-ingested rows (previously the full path surfaced the row envelope). No test relied on the old double-nested shape.

Docs: `docs/collections.md` user page, endpoint docstring table + serve() summary, CHANGELOG entry, spec section 9 (Decisions 2026-07-19).

## Summary by CodeRabbit

* **New Features**
  * Added Collections (Phase 1) to index/search approved documentation folders with collection-scoped querying and per-collection access via links and per-agent grants.
  * Added collections management via HTTP, CLI, and MCP (including listing, linking/unlinking, granting/revoking, indexing, and reversible archiving).
  * Added incremental indexing with deduplication, allowed-root enforcement, and safe handling of symlink escapes; supports async indexing with status polling.
  * Added a per-collection `embedder` field (Phase 1 uses the global default).
* **Documentation**
  * Added a Collections user guide and Phase 1 design decisions; updated Unreleased changelog.
* **Tests**
  * Added comprehensive end-to-end coverage for collections store, ingestion behavior, config, HTTP/CLI/MCP, grants/search enforcement, and indexing semantics.
* **Chores**
  * Updated ignore rules to explicitly track the collections evaluation question set.


jaylfc:
Adversarial review findings addressed. Fixes pushed as five commits, each with a regression test written first and confirmed failing before the fix.

**1. Claims gate went blind for batch rows (high)** - `_format_hit`'s unwrap-to-innermost loop stripped the provenance envelope on the semantic path (retrieval envelope -> row meta -> user metadata), so `archive_span_id` vanished from formatted hits and `prefer_verified` stopped dropping contradicted-claim rows. Fix: capture `archive_span_id`/`agent`/`project` while descending and re-attach them to a copy of the innermost metadata (user keys never clobbered). Collection hits keep exposing `file_path`. Three regression tests: span survives on the semantic and bm25 paths, and a contradicted batch row is dropped again. (`0a24e8c`)

**2. Async index blocked the service loop (medium)** - `collect_files` (full os.walk + per-file stat + 1KB null sniff) and the per-file hash+chunk step now run via `asyncio.to_thread`, so `/search` and `/ingest` stay responsive during an index; store writes stay on the loop with the thread-affine sqlite connections. Test pins the walk to a worker thread and end-to-end completion. (`a30ea3b`)

**4. No concurrent-index guard (low)** - `collections_index_start` now raises `CollectionBusyError` when the status is already `indexing`; the endpoint returns 409 with a poll hint. `ready` and `error` re-arm it. Tested at the service and HTTP layers. (`93574a3`)

**5. No tree cap on the walker (low)** - `collect_files` errors past 20000 ingestable files (constant with an upgrade-path note for making it configurable) and the index fails cleanly naming the offending path. (`fabfe4c`)

**3. Grants on the data plane** - intended Jul 19 decision, left as is; docs/collections.md now states the trust model explicitly (grants scope collections per agent inside the server-token boundary; any holder of the server token can manage grants; collection access is as strong as the server token). Finding 6's gitignore simplification already carries its upgrade-path marker. (`5edf175`)

Suite: 1186 passed (was 1178; +8 tests).

jaylfc:
Semantic eval result (the pre-registered kill-criterion arm), run on the Fedora bench host with the real MiniLM ONNX embedder rather than the lexical fallback:

```
mode=semantic   file-level Recall@5: 20/20 = 1.000
```

Protocol as specced: the repo's own `docs/` tree indexed into a temp collection, 20 gold-file questions, file-level Recall@5. This clears the ship bar (>= 0.8). For comparison the lexical/BM25 fallback arm scored 0.950 on the same question set; the one miss there was a term collision on the LoCoMo-scorecards question, which the semantic arm resolves.

Ship-bar status for Phase 1:
- file-level Recall@5 >= 0.8: PASS (1.000 semantic, 0.950 lexical)
- index the repo docs tree well under 5 min: PASS (47 files / 493 chunks in ~3s)
- zero-loss verified (re-index supersedes, delete archives, no hard-delete path): PASS (adversarial review confirmed)

Held for review.

jaylfc:
Thanks for the review. Both majors were genuine bugs, not false positives. Fixed in 5a9c0b6, 5621882, b63277e.

**1. Emptied file left stale rows searchable (zero-loss contract violation)**

Correct call. A previously-indexed file whose content became blank hit the `continue` before the supersede step, but `rel` was already in `seen`, so it counted as "still present": its old active rows were never superseded and its `collection_files` hash row was never cleared. The old content stayed in active recall indefinitely, which contradicts the incremental/zero-loss behaviour the docstring promises.

Fix: an emptied file is now dropped from `seen`, so it falls through to the existing deleted path. Its rows get superseded through `_supersede_collection_rows` (stamped `valid_to` with the `hidden_by: collection-reindex:<ts>` marker, never hard-deleted) and its file state is cleared, so a later refill re-indexes cleanly. A file that was already blank is absent from the prior state too, so it stays a no-op and does not churn on every re-index.

Regression tests: `test_reindex_emptied_file_supersedes_rows` indexes the fixture tree, asserts the content is searchable, truncates `readme.md` to whitespace, re-indexes, then asserts the old text is gone from active recall, the hash state is cleared, and the rows survive physically with the supersede marker. `test_reindex_already_empty_file_is_a_no_op` pins the no-churn half. The first test failed on the old code (`chunks_superseded == 0`, old text still returned); the second passed before and after, which is the point.

**2. Store connection leaked in `ingest_folder`**

Also correct. `CollectionStore(resolved_dir)` was never closed on any path, including the `except` that sets `status=error`. As you noted, the `_collection_store` service wrappers all close in a `finally`, so this was the one outlier, and a server doing periodic re-indexing accumulates a connection per run.

Fix: the whole body is wrapped in `try`/`finally` with `store.close()`, matching the service-layer pattern. Kept the archived-collection `ValueError` outside the inner status-setting `try` so an archived collection still raises without having its status overwritten to `error`.

Regression tests: `test_ingest_folder_closes_the_store_on_success` and `test_ingest_folder_closes_the_store_on_error` patch `CollectionStore` with a spy subclass that records `close()`, then assert every instance opened during the run was closed (the success test also asserts the underlying sqlite connection actually rejects a query afterwards). The error test induces the failure by patching `collect_files` to raise mid-index. Both failed on the old code.

**Nitpick**

`docs/collections.md`: added `text` as the language on the two HTTP surface fenced blocks (MD040).

Full suite green at 1190 passed (was 1186, plus the 4 new tests).

PR #200, opened 2026-07-21
feat(api): /version endpoint with contract-identifier capabilities list

jaylfc:
**HELD FOR REVIEW. Do not merge.**

## Why

A taOS builder wired an integration to the wrong service and "verified" the route existed by checking an HTTP status code. That check was worthless: `taosmd serve` renders the dashboard SPA on unknown non-API paths, so `GET /collections` returns `200 text/html` on a build that has no collections code at all. The check passes no matter what the server is.

Separately, a production Pi ran a month-stale build and nobody noticed, even though `GET /health` already returns a version. Semver cannot answer "does this box actually speak collections": features land continuously between bumps, so the version number and the feature set are only loosely coupled. A capabilities list answers it directly.

taOS will consume this for a drift probe and a user-facing update-alert feature.

## What

`GET /version` (unauthenticated, cheap, cacheable):

```json
{
  "version": "0.4.0",
  "commit": "76f72ffef139a9cc08c76d7348b9b25849c845a6",
  "commit_source": "git",
  "built_at": "2026-07-21T11:38:52Z",
  "built_at_source": "install",
  "capabilities": [
    "a2a.v1", "collections.v1", "grants.v1", "graph.v1",
    "ingest.v1", "search.v1", "shelves.v1", "tasks.v1", "temporal.v1"
  ]
}
```

`GET /health` gains the same `capabilities` list. Its existing `status` and `version` keys are untouched, and a test asserts that old contract explicitly so a future change cannot silently break taOS or the dashboard.

Both endpoints join `/health` in `_PUBLIC_PATHS`. They are public by design: a monitoring or drift probe must keep working on a token-secured box, and the payload is build identity plus capability identifiers only. A test asserts the response key sets are exactly the allowlist and that no path, token, or config value appears, so the public surface cannot grow sensitive fields by accident.

## The naming contract: why `.v1` suffixes

The identifiers are **stable contract identifiers**, not feature names. The suffix is the entire point.

If the list said `collections`, a consumer that checks `"collections" in caps` would keep passing after the collections wire contract changed underneath it, and would fail at runtime somewhere far from the check. With `collections.v1`, a breaking change to the contract is published as `collections.v2`: the consumer pinned to `collections.v1` sees the capability *disappear*, which is a visible, actionable break it can handle at the probe rather than in production. Additive, backwards-compatible changes keep the same identifier, which is exactly when a consumer should not have to care.

A build may advertise several versions of one contract at once during a migration window, which gives clients a window to move rather than a flag day.

The right client check is membership, not equality or prefix matching:

```python
caps = set(get(f"{base}/version").json()["capabilities"])
if "collections.v1" not in caps:
    raise RuntimeError("this build does not speak collections.v1")
```

## How the list is kept honest

The requirement was that it must be impossible for the list to claim a capability the build lacks. A hardcoded constant cannot give that, since it drifts the moment someone deletes code without editing the constant.

So the list is **derived by probing the running build**. `taosmd/capabilities.py` declares each identifier as a `CapabilityProbe` bound to the module and the symbols that implement it. A capability is advertised only if the module imports and every symbol resolves. Deleting or renaming an implementation therefore deletes the claim rather than leaving a stale boast.

Three tests defend this:

- `test_capability_is_dropped_when_its_backing_symbol_is_missing` deletes a real backing symbol and asserts `collections.v1` vanishes from the list while unrelated capabilities survive. This is the anti-drift property, tested rather than asserted in a comment.
- `test_capability_declarations_do_not_diverge_from_the_http_surface` checks every declared capability's route markers actually appear in the HTTP dispatcher. Adding a capability without wiring its routes, or renaming a route out from under a capability, fails here. This is the divergence test for the one part that cannot be probed at runtime (the handler class is built inside `_make_handler`, so its routes are not introspectable from module scope).
- `test_capabilities_are_contract_identifiers_with_version_suffix` enforces the `.vN` naming on every entry, so a bare feature name cannot slip into the contract.

The declaration table sits in one place with a comment binding it to the probe mechanism and to the divergence test, per the requirement that a declared constant live adjacent to what it describes.

## Commit resolution

Resolved once on first call and cached, so no request pays for it, and it **never shells out**. `git rev-parse` in a request path can block on an index lock, a slow or unmounted filesystem, or a missing git binary, and a monitoring endpoint must not be able to hang. The plumbing we need is plain file reads, so the files are read directly:

1. A packaged build stamp (`taosmd/_build_info.py` exporting `COMMIT` / `BUILT_AT`) wins if present, for wheel and container builds.
2. Otherwise `.git/HEAD`, resolved through a loose ref or `packed-refs`, including the `gitdir:` indirection used by worktrees and submodules (this was verified against a real linked worktree, which is how the example response above was produced).
3. Otherwise `null`.

Every step is wrapped so a corrupt or partial `.git` degrades to `null` instead of turning `/version` into a 500. `built_at` falls back to the dist-info mtime (install date) when there is no build stamp, since "when did this copy get installed" is what an operator chasing a stale box wants. `commit_source` and `built_at_source` tell the consumer how much to trust the values, which the drift probe needs.

Tests cover the loose-ref, packed-refs, detached-HEAD, build-stamp, corrupt-`.git`, and not-a-checkout cases. The not-a-checkout test builds its own throwaway directory rather than relying on the dev machine not being a repo.

## Docs

`http_server` endpoint docstring table, the README API table plus a new "Version and capability discovery" section with the client-side check, `docs/serve-service.md` (public paths and the port-is-live check), `docs/collections.md` (a "check the server actually speaks collections first" section at the top of the HTTP surface, since that is where integrators look and where the original incident started), and CHANGELOG.

## Tests

21 new tests in `tests/test_version_capabilities.py`. Full suite: **1210 passed, 10 failed**.

All 10 failures are pre-existing on `origin/master` in this environment, confirmed by running the same files on a clean `origin/master` worktree with the same venv (identical 10 failures). They are torch/CUDA related on this box (GTX 1050 Ti, `sm_61`, against a torch build supporting `sm_75`+), affecting the embedder-backed project-scoping, BM25 cache, reindex, and catalog-pipeline tests. Nothing in this branch touches those paths. 1210 - 21 new = 1189, consistent with the baseline.

## Summary by CodeRabbit

* **New Features**
  * Added a public `GET /version` endpoint with version, build metadata, and supported capabilities.
  * Added capability discovery to `GET /health`.
  * Kept `/version` and `/health` accessible without authentication, including when token security is enabled.

* **Documentation**
  * Documented endpoint response formats, capability identifiers, authentication behavior, and compatibility checks.
  * Added guidance for verifying collections and grants API support.

* **Bug Fixes**
  * Version metadata gracefully falls back to `null` when unavailable without causing endpoint errors.


jaylfc:
Superseded by #213, now merged: the identical feature content (capabilities.py and tests byte-identical, verified in the #213 review) rebased onto current master. This PR's review history carries over: CodeRabbit's completed pass and the Jay+lead walkthrough approval of 2026-07-26 are referenced in the #213 review comment. Closing per the one-PR-per-task retry rule; branch kept for history.

PR #201, opened 2026-07-21
feat(storage): schema migration framework (per-db user_version, legacy stamping, subsumes hand-rolled ALTERs)

jaylfc:
**HELD FOR REVIEW. Do not merge without Jay's sign-off.**

## Why

taOSmd had no migration mechanism at all. No `PRAGMA user_version`, no version stamp, no upgrade path. Schemas were created with `CREATE TABLE IF NOT EXISTS` and evolved by three hand-rolled `try/except ALTER TABLE` guards (archive `project`, session-catalog taxonomy columns, vector-memory `valid_to`).

That worked only because every change so far happened to be additive. `CREATE TABLE IF NOT EXISTS` does nothing to a database that already exists, so the first column added to an existing table reaches new installs and no existing one, and every query naming it fails at runtime on exactly the stores holding real data. `collections.db` had no upgrade path whatsoever, so that break was one column away from every collections store in the field. The recent 265-commit upgrade of the live Pi store was only safe because it happened to be additive-only. taOS is now shipping taosmd to users, so this stops being an internal problem.

## The design

`taosmd/migrations.py`, dependency-free. Per-database schema version in the database file itself via SQLite's native `PRAGMA user_version` (no extra table). An ordered tuple of `Migration` steps per database in `REGISTRY`. The runner applies only steps above the stored version, each inside its own `BEGIN IMMEDIATE`, writing the version stamp in the same transaction as the change, so a failure rolls back both. When a database is current it is a single PRAGMA read, so every store calls it on open, after its schema constant and before its first query.

Two rules bind a step, both enforced by tests: it must be idempotent, and it must not commit, roll back, or call `executescript()` (which issues an implicit COMMIT and would silently defeat the rollback guarantee, hence the `exec_script()` helper).

## The zero-version problem

Databases in the field report `user_version = 0` while carrying a fully modern schema, because nothing ever stamped them. A runner that trusted that zero would replay every step against a live store.

So each migration carries a `detect` probe: a cheap `PRAGMA table_info`-style read answering "is this change already present?". On an unstamped database the runner does not migrate first, it looks first. It walks the list in order and stamps to the longest leading run of steps that report themselves already present, **executing none of them**, then applies only the remainder. The walk stops at the first step that does not detect itself rather than skipping over it; a later step that happens to be present is re-applied, which is safe under the idempotency rule and never assumes a step ran.

One mechanism covers every case:

| database state | detected | result |
| --- | --- | --- |
| new empty file | nothing | built from step 1 to latest |
| older build, modern schema | everything | stamped to latest, zero schema writes |
| older build, mid-era schema | a prefix | stamped to the prefix, tail applied |
| already stamped and current | not consulted | one PRAGMA read |

## What it subsumes

- **archive-index.db** v2: the `project` back-fill, previously a `SELECT project ... LIMIT 1` probe inside a bare `except`.
- **session-catalog.db** v2/v3: the five taxonomy columns and `idx_sessions_path`, then `agent_name`. The index is still created after the columns it indexes, inside the same step.
- **vector-memory.db** v2: `VectorMemory._migrate`'s `valid_to` column plus the deferred `idx_vm_valid`. `_migrate` is deleted; the ordering hazard its comment warned about (CREATE INDEX in the schema constant firing before the ALTER on a legacy table) is preserved by construction.
- **collections.db** v1: the Phase 1 schema, previously an inline script with no versioning, now a module constant used as the baseline.
- **knowledge-graph.db**, **claims.db** v1: registered at their current schema so future changes have somewhere to go.

The schema constants stay and still run on open, and version 1 of each database runs the same constant. The line that matters: adding a whole new table can stay in the schema constant (`IF NOT EXISTS` genuinely reaches existing databases), changing an existing table must be a migration. `docs/migrations.md` states that contract.

## Test story

37 new tests in `tests/test_migrations.py`.

The ones that carry the value seed genuine legacy databases at `user_version = 0` and assert both the resulting version and, through a spy wrapping every step's `apply`, that **nothing was re-run**:

- `test_legacy_modern_archive_index_is_stamped_without_reapplying` is the field case and the one that would have caught a real break: modern schema, version 0, real rows. Asserts `calls == []`, `stamped_baseline is True`, version at latest, data intact.
- `test_legacy_collections_db_is_stamped_to_phase1_baseline`, same for the store that had no path at all.
- `test_genuinely_old_archive_index_gains_project_column`, `test_legacy_session_catalog_without_taxonomy_columns`, `test_legacy_vector_memory_gains_valid_to_and_its_index`: pre-change shapes with real rows, schema advanced, every value round-tripped afterwards.
- `test_partially_migrated_database_advances_only_the_missing_steps`: taxonomy present, `agent_name` dropped, only the tail runs.

Plus: fresh empty file lands at latest (parametrized over all six databases), idempotent second and third runs apply nothing (parametrized, spy-verified), a failing migration rolls back its schema change *and* its version stamp with no half-applied state and no open transaction, status/check reporting, and end-to-end tests that open the real `ArchiveStore`, `SessionCatalog`, `VectorMemory`, `TemporalKnowledgeGraph`, `ClaimStore`, and `CollectionStore` over both fresh and legacy files.

The demonstration: `test_new_column_on_existing_table_reaches_an_existing_store` first proves the failure mode (re-running the `CREATE TABLE` script with a new column in it changes nothing on an existing database), then adds the same column as an appended migration and shows the already-current store gains it, at the bumped version, with the pre-existing row intact and the next open a no-op.

## Suite

1161 passed, 5 skipped on this branch, against a 1124-passed master baseline in the same environment: exactly +37, no regressions. The same 9 failures appear on master and on this branch and are environment-dependent (missing ONNX embedder and BM25 extras): `test_project_storage` (4), `test_hermes_audit_batch2` (3), `test_reindex` (1), `test_catalog_pipeline` (1).

## New CLI

```
taosmd migrate --status     # per-database version, latest, pending steps
taosmd migrate --check      # dry run, exits 1 if anything is pending
taosmd migrate              # apply pending steps out of band
taosmd migrate --status --json
```

`--check` is the deploy pre-flight for the taOS drift/deploy slice: writes nothing, exits non-zero when any database in the data dir is behind.

## Notes for review

- Scope is the six databases with real schema surface. `crystals.db`, `insights.db`, `browsing-history.db`, `access-tracker.db`, `tasks.db` and `memory-settings.db` are not registered yet; adding them is a `REGISTRY` entry each and no code change.
- `CollectionStore` still opens with a bare `sqlite3.connect` rather than `_db.connect`, so it alone is not in WAL mode. Left alone deliberately, out of scope for this PR, but worth a follow-up.
- `pending_decisions.py` opens `knowledge-graph.db` as a second owner and creates its own tables through its own schema constant. It works, but two modules owning one file's schema is a shape to watch when a migration eventually has to touch those tables.

## Summary by CodeRabbit

* **New Features**
  * Automatic SQLite schema migrations now run when each store opens, including safe upgrades from legacy databases.
  * Added `taosmd migrate` to display per-database migration status, apply pending migrations, run `--check` health checks (non-zero when upgrades are needed), and output machine-readable `--json`.
* **Documentation**
  * Added migration framework documentation and an updated changelog entry.
* **Tests**
  * Added comprehensive tests covering idempotency, rollback/atomicity, status/migrate-all behavior, and CLI workflows.


jaylfc:
@coderabbitai review

jaylfc:
Both review findings applied in 4febe31, with tests verified to fail without each fix.

**--check --json** now suppresses the plain-text hint, so the stream stays parseable. The exit code still signals pending work, which is the part a deploy pre-flight actually gates on. `test_cli_check_json_stays_parseable_when_databases_are_behind` parses the output with `json.loads` and asserts the hint is absent.

**Registry validation** added as a one-pass import-time check. Worth stating why this is more than tidiness: the baseline walk indexes steps by position and `_start_version` resolves a version by name lookup, so a mistyped version or a duplicated name would not raise, it would stamp the *wrong version* against a real store. That is the exact silent-corruption failure this module exists to prevent, so leaving its own registry unchecked was a gap. Three tests cover it: a non-contiguous version, a duplicate name, and one asserting the shipped registry passes its own invariant rather than only fixtures.

Verified before adding the assertion that all six shipped registries are already contiguous from 1 with unique names, so it cannot brick import.

Suite on this branch: 42 passed in `tests/test_migrations.py` (was 38). Still **held for review, not merging**.

PR #204, opened 2026-07-21
bench(longmemeval): route the runner through retrieve() so the real retrieval path is measured (unblocks E-030)

jaylfc:
**HELD FOR REVIEW. Do not merge. No benchmark was run.**

## The defect

`benchmarks/longmemeval_runner.py` never called `taosmd.retrieval.retrieve()`. It hand-assembled its context:

```python
full_context = ctx["context"] + " " + archive_text + " " + vector_text
```

where `vector_text` came from a direct `vmem.search()` plus an optional reranker pass. Two consequences.

1. **The runner was not measuring the shipped path.** Every LongMemEval number it produced described a pipeline users do not have. Anything `retrieve()` does that the hand-rolled stage does not (near-duplicate filtering, fusion, the layer plumbing) was absent from the measurement.
2. **`graph_expansion` (PR #191) was unreachable from it**, which blocks the pre-registered E-030. Setting the control would have had no effect and both arms would have been byte-identical, so the experiment would have silently measured nothing and reported a null.

This is the same defect PR #196 (`ea882a2`) fixed for the EventQA runner. This change follows that wiring closely so the two runners stay consistent.

## What moved

Only the vector stage. `ContextAssembler` and the archive-FTS pass are untouched and still contribute to the context exactly as before.

## Reconstruction choices

`retrieve()` has different defaults from the code it replaces, so the call is deliberately parameterised to reproduce the old stage rather than to be idiomatic:

- `strategy="custom"`, `memory_layers=["vector"]` so only the vector source is searched, as before. `kg` is still passed in `sources` because `_append_graph_expansion` reads `sources["kg"]` regardless of which layers were searched, so the readback block can fire without the KG competing for `limit` slots.
- `candidate_top_k=RETRIEVE_LIMIT` pins the vector fetch to exactly the width the old code used. Left unset, `retrieve()` fetches `limit * 3` and would have silently widened the instrument.
- `limit` narrows to `RERANK_TOP_K` only when a reranker is actually available, matching the old narrow-after-rerank behaviour. With no reranker it stays at `RETRIEVE_LIMIT`.
- `verify` is left at its default `False`. The old stage never gated on claims; enabling verification here would change what the benchmark measures.
- Decomposition (`TAOSMD_DECOMPOSE=1`) keeps its shape: per-sub-query fetches with **no** per-query rerank, deduplicated union truncated to `RETRIEVE_LIMIT`, then one rerank at the end. Reranking inside each sub-query fetch would have built the union out of pre-narrowed pools.

**Residual difference, measured not assumed.** `retrieve()` applies a near-duplicate filter (Jaccard >= 0.8) where the old path deduplicated on exact string equality. Unlike EventQA's distinct 4096-token passages, LongMemEval chunks are ~100-word windows with a 20-word overlap, so this filter **can** fire here. `--report-retrieval-delta` quantifies it per question instead of hand-waving it.

`graph_expansion` is refused, not ignored, on `--retrieval-path legacy` (the hand-rolled path cannot honour it) and under `TAOSMD_DECOMPOSE=1` (the union is truncated to `RETRIEVE_LIMIT`, which can discard the derived block and make the arm unmeasurable).

## Non-zero exit: an improvement over the EventQA probe

`retrieve_supports_graph_expansion()` is ported from EventQA, with one fix. The EventQA refusal path uses a bare `return`, so the process exits **0** and an automated chain checking `$?` reads a refusal as a success. This version calls `sys.exit(1)`. **The EventQA runner should be brought into line** in a follow-up.

This is not hypothetical. On `origin/master` the probe returns **False**: `graph_expansion` lives on `origin/feat/bitemporal-readback` (`9286e47`) and is not an ancestor of master. E-030 therefore cannot run on master today, and on this branch it fails loudly instead of quietly producing two identical arms.

## Tests

`tests/test_longmemeval_retrieve_wiring.py`, 15 tests, synthetic fixtures only (the LongMemEval-S / oracle data is gitignored and absent on most boxes, so nothing here touches `benchmarks/data`).

Closes a real gap in the EventQA test file: it asserted the probe *returns a bool* but never asserted the run *aborts*, leaving the safety-critical behaviour untested. Covered here:

- the run aborts with a **non-zero** `SystemExit` when the probe is False, and the dataset is not even opened;
- a supported checkout does not abort, and `graph_expansion=0` never consults the probe at all (so the default arm runs anywhere);
- flag forwarding, including that the kwarg is *omitted* at 0 so the runner still imports on a pre-#191 checkout;
- `verify` stays off;
- wired-vs-legacy equivalence on identical inputs, single-query and decomposed, plus an assertion that both fetch the same candidate-pool width;
- the derived block reaches the generator context and the two arms differ;
- legacy and decompose both refuse `graph_expansion`;
- CLI defaults and rejection of an unknown `--retrieval-path`.

## Result JSON

The runner now writes `benchmarks/results/longmemeval_<ts>.json` (or `--out`) recording `retrieval_path`, `graph_expansion`, `num_ctx`, the lever env, per-type metrics and the delta summary.

`num_ctx` is a new `TAOSMD_LME_NUM_CTX` knob defaulting to **0 = do not send**, which leaves Ollama on its own default and keeps historical numbers reproducible byte for byte. Worth flagging separately: a 16000-char context is roughly 4500-5000 tokens, so Ollama's 4096 default is likely already truncating some LongMemEval prompts. Raising it changes what is measured, so it is opt-in here rather than silently bumped.

## What an operator should run before trusting the arms

Establish the delta before reading any E-030 result:

```bash
TAOSMD_OLLAMA_MODEL=... TAOSMD_JUDGE_MODEL=... \
  .venv/bin/python benchmarks/longmemeval_runner.py \
    --llm --limit 50 --report-retrieval-delta
```

Read `identical_pct` in the `RETRIEVAL DELTA` block. 100% means the published anchors carry over untouched. Anything less is the exact amount that has to be re-anchored, and given the overlapping chunks it is unlikely to be 100% here. If it is low, re-anchor with `--retrieval-path legacy` on the same slice before comparing arms.

## Suite

Same environment, same box. Pre-existing failures are environmental (torch/CUDA: GTX 1050 Ti `sm_61` against a torch build wanting `sm_75+`, plus a missing ONNX model) and were not chased.

| | result |
|---|---|
| `origin/master` baseline | 9 failed, 1124 passed, 5 skipped |
| this branch | 9 failed, **1139** passed, 5 skipped |

Identical failure set. +15 passing, all new.

## Summary by CodeRabbit

* **New Features**
  * Added configurable retrieval modes for benchmark runs, including wired and legacy paths.
  * Added optional graph expansion, query decomposition, cross-encoder reranking, and retrieval-delta reporting.
  * Added command-line options for retrieval configuration and JSON result output.
  * Benchmark results now include structured per-question retrieval details.

* **Bug Fixes**
  * Prevented unsupported graph expansion configurations from running and producing invalid results.

* **Tests**
  * Added comprehensive coverage for retrieval behavior, CLI validation, graph expansion, and result comparisons.

### DIFF
--- a/benchmarks/longmemeval_runner.py
+++ b/benchmarks/longmemeval_runner.py
@@ -12,6 +12,7 @@ Usage: .venv/bin/python benchmarks/longmemeval_runner.py [--limit N] [--type TYP
 
 import argparse
 import asyncio
+import inspect
 import json
 import os
 import sys
@@ -24,6 +25,7 @@ from taosmd.knowledge_graph import TemporalKnowledgeGraph
 from taosmd.archive import ArchiveStore
 from taosmd.memory_extractor import extract_facts_from_text, process_conversation_turn
 from taosmd.context_assembler import ContextAssembler
+from taosmd.retrieval import retrieve as _retrieve
 from taosmd.vector_memory import VectorMemory
 
 
@@ -66,7 +68,23 @@ SELF_VERIFY = os.environ.get("TAOSMD_SELF_VERIFY", "0") == "1"
 # temporal). Set TAOSMD_SAMPLE_SEED to shuffle deterministically before slicing
 # so a screen at limit<500 covers all six types. Unset = head slice (back-compat).
 SAMPLE_SEED = os.environ.get("TAOSMD_SAMPLE_SEED")
+# Ollama context window for generation. 0 (the default) means "do not send
+# num_ctx", which leaves Ollama on its own default and keeps every historical
+# LongMemEval number reproducible byte for byte. It is recorded in the result
+# JSON so a run's window is never guessed after the fact. Note that a 16000-char
+# context is roughly 4500-5000 tokens, so Ollama's 4096 default already
+# truncates some prompts; raising this changes what is measured, which is why it
+# is opt-in rather than silently bumped here.
+NUM_CTX = int(os.environ.get("TAOSMD_LME_NUM_CTX", "0"))
 _reranker = None
+
+
+def _gen_options(**extra):
+    """Ollama options dict, carrying num_ctx only when it was explicitly set."""
+    opts = dict(extra)
+    if NUM_CTX:
+        opts["num_ctx"] = NUM_CTX
+    return opts
 
 
 def sample_dataset(dataset, limit, seed=None):
@@ -175,7 +193,7 @@ async def llm_answer(client, context: str, question: str) -> str:
                 "messages": [{"role": "user", "content": ANSWER_PROMPT.format(context=context[:CONTEXT_CHARS], question=question)}],
                 "stream": False,
                 "think": False,
-                "options": {"temperature": 0, "num_predict": 100},
+                "options": _gen_options(temperature=0, num_predict=100),
             },
             timeout=30,
         )
@@ -251,7 +269,7 @@ async def self_verify_answer(client, context: str, question: str, answer: str) -
                 "messages": [{"role": "user", "content": VERIFY_PROMPT.format(context=context[:CONTEXT_CHARS], question=question, answer=answer)}],
                 "stream": False,
                 "think": False,
-                "options": {"temperature": 0, "num_predict": 100},
+                "options": _gen_options(temperature=0, num_predict=100),
             },
             timeout=30,
         )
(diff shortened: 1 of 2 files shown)


jaylfc:
HELD FOR REVIEW. Do not merge. No benchmark was run as part of this change.

jaylfc:
@coderabbitai review

jaylfc:
@coderabbitai review

jaylfc:
Triaged. Still HELD, but the hold is now a measurement rather than an open question, and two of this PR's own premises turned out to be stale.

I re-measured all three against master `993348c7` instead of trusting the description, which is the only reason this comment is worth anything after five weeks.

**Stale, and must not be carried forward.** The body says `graph_expansion` is absent from master, so E-030 cannot run and the probe fails loudly. It is on master now: 18 references in `taosmd/retrieval.py`, plus `api.py`, `controls.py` and `__init__.py`. The probe returns True, so the refusal path described here will not fire. E-030 is no longer blocked on that.

**Also stale.** The body flags the EventQA runner's refusal as exiting 0 through a bare `return`, and proposes a follow-up to bring it into line. That is already fixed: `benchmarks/eventqa_runner.py` calls `sys.exit(1)` on exactly that path. I was about to file the follow-up and checked first; no card was filed, because the defect does not exist.

**Still true, and it is the whole blocker.** `retrieve()` calls `_deduplicate(results, threshold=0.8)` at `taosmd/retrieval.py:234`, a Jaccard word-set near-duplicate filter, from three call sites. The stage this PR replaces deduplicated on exact string equality. LongMemEval chunks are roughly 100-word windows with a 20-word overlap, so the filter can genuinely fire on this data. That makes this an instrument change rather than plumbing: it can move the published LongMemEval-S anchor with no lever having changed.

The PR merges cleanly into master, and correctness is not what is holding it. What is holding it is that no benchmark has ever been run against it, so the size of that instrument change is unknown, and the description is right that `identical_pct` has to be established before either arm means anything.

I am deliberately not merging on the strength of a clean diff and a green suite. A suite cannot see a changed measurement, and this repo has already had one cleanup PR quietly redefine what Recall@k measures.

Filed as `tsk-cgsbbt`, marked ops and not lane-claimable because it needs the Fedora GPU host under the lease protocol, one benchmark at a time. It carries the exact command, both stale premises so nobody re-derives them, and how to read the result: 100 percent identical means the published anchors carry over and this can merge on that evidence; anything less is the exact fraction to re-anchor with `--retrieval-path legacy` on the same slice first.

Staying open. When the number exists, this gets its disposition.

jaylfc:
**`identical_pct` measured. It is 98.0%, not 100%, so this stays HELD and the anchors do NOT carry over untouched.**

The hold was always a measurement rather than an open question, and the measurement now exists.

```
RETRIEVAL DELTA (wired retrieve() path vs pre-wiring legacy path)
  questions compared:      50
  byte-identical contexts: 49 (98.0%)
  mean chars legacy/wired: 19915 / 19901
```

**The divergence is exactly one question and it has the predicted shape.** q=41, legacy 35792 chars against wired 35125, a 667-character reduction, 1.9% of that question's context. Every other question in the slice is byte-identical with a character delta of exactly zero. That is the signature `tsk-cgsbbt` predicted: `_deduplicate(results, threshold=0.8)` in `taosmd/retrieval.py` is a Jaccard word-set near-duplicate filter, LongMemEval chunks are roughly 100-word windows with 20-word overlap, and on one question in fifty the filter found a pair similar enough to collapse. The wired context is SMALLER, which is consistent with near-duplicate removal rather than with retrieving different material.

**Reading it by the rule written into the card: 98.0% is below 100%, so this is the exact fraction that needs re-anchoring, and a wired number must not be compared against a legacy-era published figure.** I am deliberately not merging on 98%. This repo has already had one cleanup PR quietly redefine what Recall@k measures, and a 2% instrument change is small enough to be invisible in a published headline and real enough to move it.

**Provenance, per the benchmark-honesty policy.**

| field | value |
|---|---|
| box | Fedora RTX 3060 12GB (node=linstation), under the GPU lease protocol |
| tree | trial merge of this PR head into master `b694190a`, merged commit `232af863`, 2 files changed |
| slice | `--limit 50`, 50 of 500 questions compared, all 50 produced a delta record |
| dataset | `longmemeval_s_full.json`, 500 questions, 277383467 bytes, sha256 `d6f21ea9d60a0d56f34a05b609c79c88a451d2ae03597821ea3d5a9678c3a442` |
| generator / judge | **none, and this is deliberate** |
| levers | `TAOSMD_DECOMPOSE=0`, `TAOSMD_RERANK=0`, both defaults |
| run rc | 0, zero tracebacks in the log |

**On running without `--llm`, which departs from the card's command.** The card says `--llm --limit 50 --report-retrieval-delta`. I dropped `--llm` and the number is unaffected, which I established from the source rather than assumed. In this PR's runner the delta block at `if report_retrieval_delta:` is sequential to and independent of the `if use_llm and llm_client is not None:` block that follows it, so `identical` is computed from the retrieved contexts before any generation happens. The only path by which an LLM could perturb retrieval is query decomposition, and `decomposing = DECOMPOSE and llm_client is not None` where `DECOMPOSE` defaults to `0`. With decomposition off, retrieval is identical either way. Dropping the flag was not a shortcut around a blocker; it is what let the measurement happen at all, because **every Ollama model on that box has been deleted** and an `--llm` arm is currently unrunnable there.

**The `0.0%` accuracy in that log is an artefact of running with no generator and must not be read as a regression.** No answers were generated, so nothing could be scored.

**What this now needs, and it is no longer this PR's fault.** The re-anchor arm requires `--retrieval-path legacy` with `--llm` on the same 50-question slice, which requires a generator and a separate non-thinking judge on that box. Historical provenance from `bench-logs/` is `llama3.1:8b` as generator with `qwen3:4b` as external judge; the runner's own default is `qwen2.5:3b` and its comment warns to set a DIFFERENT judge, which matches the 15 to 22 point same-model inflation we measured. Until those are pulled the accuracy arms cannot run.

Staying open. `tsk-cgsbbt` stays open with its remaining scope narrowed to the two accuracy arms. The two stale premises recorded in the card, `graph_expansion` being absent and the EventQA bare `return`, remain stale and are not to be carried forward.

jaylfc:
## Accuracy arms: measured, and the retrieval question is settled

Both arms are in. The earlier `identical_pct` of 98.0% told us the wired path *almost* reproduces legacy retrieval; it could not tell us whether the one divergence mattered. It does not, and if anything it helps.

### Configuration, recorded with the number

| | |
|---|---|
| Generator | `qwen3.5:9b` |
| Judge | `llama3.1:8b` (external, cross family; E-012 forbids a same-model judge) |
| Slice | 50 questions, `--limit 50` |
| Dataset | `longmemeval_s_full.json`, sha256 `d6f21ea9d60a0d56f34a05b609c79c88a451d2ae03597821ea3d5a9678c3a442`, 277383467 bytes, 500 questions |
| Box | linstation, RTX 3060 12GB |
| `retrieve_limit` | 12 |
| Worktree head | `232af86` |
| Variable | `--retrieval-path` only. Identical models, identical slice. |

### Result

| Arm | Path | Correct | Accuracy |
|---|---|---|---|
| A | `legacy` | 15 / 50 | 30.0% |
| B | `retrieve` (wired) | 16 / 50 | 32.0% |

Paired on question index, with the question text verified identical on all 50:

```
legacy-right -> wired-WRONG : 0   idx=[]
wired-right  -> legacy-wrong: 1   idx=[1]
idx where retrieved_chars differs: 1/50
total retrieved_chars legacy=995728 wired=995061 delta=-667
```

**Zero regressions.** Exactly one question's retrieval differs, and it is the same question that flips. The 667 character reduction is the Jaccard `_deduplicate(threshold=0.8)` signature, and it matches the earlier byte-level measurement exactly, where the sole divergence was one question at legacy 35792 vs wired 35125 characters. Two independent measurements, same single question, same 667 characters.

### What this does and does not claim

It does not claim an accuracy improvement. n=50, and a one question difference is +2.0pp, which is well inside noise. Also worth stating plainly: **every question in this slice is `temporal-reasoning`**, so these absolute numbers are a single-category reading and are not comparable to any published overall LongMemEval-S figure. That is fine for the purpose, because both arms drew the same slice and the comparison is paired.

What it does establish is the thing the hold was about: the wired path is behaviourally equivalent to legacy, its one divergence is a deduplication that removed redundant context rather than dropped evidence, and that divergence did not cost a single answer. The legacy arm was re-run here rather than compared against a legacy-era published number, which is what re-anchoring required.

### Disposition

The retrieval and accuracy question that held this PR is **answered, favourably**. The remaining gate is the three verified CodeRabbit findings tracked on `tsk-p6ejva` (module-scope `int(os.environ...)` crash on a malformed value, `--report-retrieval-delta` comparing legacy against legacy, and `--out` into a missing directory failing only after the full run discards results).

Note for the board: `tsk-p6ejva` cannot currently dispatch. It carries the `fix-forward` label with no `BASE: exec/tsk-<id>` line, and `next_card.py` refuses that combination outright, silently. It is not fixable by relabelling either, because this branch predates the `exec/tsk-*` convention that the fix-forward mechanism resolves against. Those three findings are therefore lead work, not lane work.

GPU released; both models unloaded, box back to 1297 MiB and 0% utilisation.

jaylfc:
## Three CodeRabbit findings fixed, with fail-first evidence

Pushed `05a5d317` to this branch. Nothing outside `benchmarks/longmemeval_runner.py` and `tests/test_longmemeval_retrieve_wiring.py` is touched.

### What changed

**1. `TAOSMD_LME_NUM_CTX` no longer kills the import.** It was parsed with a bare `int()` at module scope, so a typo or an empty value raised before the runner could name the variable responsible. It now falls back to the historical default and warns once, naming the offending value. An empty value is treated as unset and stays quiet, so the warning that does appear is worth reading.

**2. `--report-retrieval-delta` with `--retrieval-path legacy` is refused.** The delta builds the legacy context and compares it against the context the run used. On the legacy path those are the same assembly, so the comparison was legacy against legacy: identical by construction, and a 100 percent anchor match that measured nothing. It now exits non-zero in the style of the existing `graph_expansion` refusal, so a chain checking `$?` reads a refusal as a failure.

**3. `--out` is validated before the run, not after it.** An `--out` into a directory that could not be created failed at the final write, by which point every question had been answered and the results were gone. The parent is now created and probed with a real create-and-delete before the question loop. The probe is deliberately not `os.access()`, which answers about permission bits and not about read-only mounts, a full filesystem, or a parent component that is a regular file.

### Fail-first evidence

New tests run against the unpatched runner restored to `fd67a87e`, with the file identity in the same command as the verdict so the result names what it actually examined:

```
worktree runner : 24f9808f7b2eb0c0
fd67a87e  blob  : 24f9808f7b2eb0c0

5 failed, 2 passed, 15 deselected in 0.35s
FAILED test_malformed_num_ctx_does_not_kill_the_import
FAILED test_empty_num_ctx_is_treated_as_unset_and_stays_quiet
FAILED test_delta_on_the_legacy_path_is_refused
FAILED test_missing_out_directory_is_created_before_the_run
FAILED test_unwritable_out_path_is_refused_before_the_run
```

The 2 that pass there are the positive controls (`test_well_formed_num_ctx_is_still_honoured`, `test_delta_on_the_wired_path_is_not_refused`). They matter: they show the harness reaches the code and that the 5 failures are specific to the three defects rather than blanket breakage.

Same 7 tests against the patched runner (`ef4218f7fad4b174`): `7 passed`. Whole file: `22 passed`.

### Mutation testing

Each fix branch was mutated individually and the owning test had to die. Every mutation asserted its anchor appears exactly once before applying, because an anchor that matches twice silently applies nothing and reports a green it has not earned:

```
M1a  except returns 0 -> re-raises              KILLED
M1b  warning no longer names the value          KILLED
M1c  empty-string guard removed                 KILLED
M2   legacy-delta refusal no longer exits       KILLED
M3a  makedirs removed from the pre-run probe    KILLED
M3b  pre-run probe no longer called             KILLED
```

M1c survived the first attempt. With the guard weakened, an empty string still fell through to the `except` and returned 0, so asserting only `NUM_CTX == 0` could not tell the two branches apart. The guard's actual job is separating unset (silent) from malformed (warned), so the test now pins that and the mutation dies.

### Full suite, trial merge against current master

Merged `05a5d317` onto master `878ba78e` (merge commit `dda0dd21`, no conflicts) and ran both sides in the same worktree and the same venv:

```
clean master 878ba78e : 1771 passed, 10 skipped, 7 errors in 183.79s
trial merge  dda0dd21 : 1793 passed, 10 skipped, 7 errors in 199.88s
```

The failing set is identical, not merely equal in count: a sorted diff of the ERROR and FAILED lines is empty. All 7 are the pre-existing `taosmd.mcp_server.build_server` import error and they are present on unpatched master in the same env. The delta of exactly +22 passed is the 22 tests this file collects.

### Gates on the merged tree

```
deleted-symbols-guard: clean
trailing-newline-gate: clean
normalise-handle-gate: clean
witness-gate:          clean
doc-gate:              clean
ruff:                  27 findings before, 27 after, rule-code diff empty
```

The doc-gate clean is a scope pass rather than a blind one, and I proved it by making the gate fail: staging a new file under `taosmd/` reds the changelog rule (`DOC-GATE FAIL: changelog`, rc=1), while this changeset touches only `benchmarks/` and `tests/`, which no rule covers. A comment-only edit under `taosmd/` does not trip it either, since `when_changed` fires on structural changes by default.

Issue #206, opened 2026-07-21
CLI silently operates on a remote store when server_url is set, with no visible target


jaylfc:
Prompted by a real incident on a dev box (2026-07-21) and by @taOS-dev's framing of the fix.

`~/.taosmd/config.json` on the taOSmd dev machine contained `server_url` as its only key, pointing at the Pi. Every `taosmd` CLI command run there was therefore acting on the **live production store**, not a local one, with nothing in the output to say so. The operator had no reason to suspect it.

This is a worse failure class than the copied-config hazard we were already tracking. A copied config announces itself when the new host behaves oddly. This one is silent by construction: the commands succeed, the output looks normal, and the only difference is which store got mutated.

**Why documentation is not the fix**

The existing behaviour is arguably correct and useful. The problem is that it is invisible. A note in the docs does not help an agent or operator who does not already suspect there is something to check.

**Proposed affordance** (either, or both)

1. Print the resolved target on any **mutating** command when it is not local, e.g. `target: http://<host>:7900 (remote)`. Reads stay quiet so scripts are not spammed.
2. Require an explicit flag (`--remote`, or `--target`) for mutating commands when `server_url` is set, so operating on a remote store is a decision rather than a default inherited from a config file.

Option 1 alone removes the silence. Option 2 additionally removes the trap, at the cost of a breaking change for anyone deliberately scripting against a remote. Worth deciding which, rather than defaulting to the softer one because it is easier.

**Related**

Same family as #199 (a runtime path with no escape hatch) and #197 / #205 (an operation whose failure or unexpected mode produces no signal). The common shape across all four: the mechanism behaves reasonably, and the fact that it did something surprising is discarded rather than surfaced.

Issue #205, opened 2026-07-21
EventQA runner exits 0 when it refuses to run, so a chain reads the refusal as success


jaylfc:
Found while wiring the LongMemEval runner through `retrieve()` (PR #204).

`benchmarks/eventqa_runner.py` refuses to run when the checkout does not support the requested lever (for example `--graph-expansion` on a build where `retrieve()` has no `graph_expansion` parameter). That refusal is correct: the alternative is silently producing two identical arms and reporting a null result that is really an instrument failure.

The problem is how it refuses. It returns from `main()` rather than exiting non-zero, so the process exits **0**. An overnight chain that runs arms in sequence and checks `$?` between them reads "I refused to run" as "this arm succeeded." The chain proceeds, and the missing arm shows up later as an absent or empty result file rather than as a failure at the point it happened.

This is the same fails-open shape as #197: the error path is present and correct, and then the signal that the error occurred is thrown away.

**Fix**

`sys.exit(1)` on refusal, as PR #204 does for the LongMemEval runner. Worth also asserting the abort in the EventQA tests, which currently do not cover it: non-zero exit, dataset not opened, and a supported checkout not aborting.

**Related and worth checking at the same time**

Whether any other runner under `benchmarks/` refuses or errors without a non-zero exit. The pattern is easy to repeat and invisible until a chain silently skips an arm.

jaylfc:
Fixed by #290 (`8e0f45a1`, 2026-08-17). Verified on master `e942958e` rather than taken from the PR: the unsupported-lever refusal in `async def run` now ends in `sys.exit(1)`, as does the empty-rows refusal, so a chain reads a refusal as a failure instead of a success. The three guards live in `tests/test_eventqa_retrieve_wiring.py` as `test_unsupported_graph_expansion_aborts_the_run`, `test_empty_dataset_aborts_the_run` and `test_zero_graph_expansion_never_probes`, the last covering the case where `graph_expansion=0` must run on any checkout without probing at all.

Closing late: the fix landed yesterday and this issue stayed open, which also left JOB-004 sitting in `docs/agent-jobs/` as claimable work pointing at an already-fixed bug. That job is now closed out as `DONE-job-004-...` in the same sweep.

Issue #203, opened 2026-07-21
Six databases are still unregistered with the migration framework, and knowledge-graph.db has two schema owners


jaylfc:
Follow-up from PR #201, which introduces per-database schema versioning but deliberately scopes itself to the databases with existing hand-rolled migration guards plus `collections.db`.

**Unregistered databases**

`crystals.db`, `insights.db`, `browsing-history.db`, `access-tracker.db`, `tasks.db`, `memory-settings.db` are not in the migration `REGISTRY`. Until they are, each has exactly the problem the framework exists to fix: the first column ever added to an existing table in one of them will silently fail to appear on databases already in the field, because `CREATE TABLE IF NOT EXISTS` does nothing to an existing table.

Each is a `REGISTRY` entry, not a code change, so this is mechanical work rather than design work.

**Two owners of one schema**

`pending_decisions.py` creates its own tables inside `knowledge-graph.db` through its own schema constant, so that file now has two modules owning parts of its schema. It works today. It becomes ambiguous the moment either module needs a migration, because neither owns the file's version stamp on its own.

Worth a deliberate decision: either give `pending_decisions` its own database, or make one module the schema owner and have the other register its migrations through it. Better settled before someone has to migrate those tables under time pressure.

**Related**

PR #201 notes a third point worth recording here: the legacy-stamping `detect` walk stops at the first step that does not detect itself rather than skipping it. That is the conservative choice and is safe under the idempotency rule, but it means step idempotency is load-bearing rather than merely tidy. Any future migration author needs to know that.

Issue #202, opened 2026-07-21
CollectionStore opens sqlite without WAL or a busy timeout, unlike every other store


jaylfc:
Found while building the schema migration framework (PR #201).

Every other database in taosmd is opened through the shared `_db.connect` helper, which sets WAL mode and a busy timeout. `CollectionStore` alone uses a bare `sqlite3.connect`.

**Why this matters beyond consistency**

Collections indexing runs asynchronously on the service loop and can overlap with search requests that read collection rows and grants. Without a busy timeout, a concurrent writer produces an immediate `database is locked` error rather than waiting; without WAL, readers and the writer block each other more than they need to. The Phase 1 workload is small enough that this may not have surfaced yet, but the shape is there: an index pass writing file-state and collection rows while a search checks grants on the same file.

**Fix**

Open `collections.db` through the same `_db.connect` path as the other stores so it inherits WAL and the busy timeout. Worth checking whether anything in `CollectionStore` depends on the current journal mode before switching.

Low severity today, cheap to fix, and it removes a latent concurrency failure from a feature that is about to be driven by an automated Library pipeline.

jaylfc:
Fixed and on master since PR #215 (merged 2026-07-28); the issue was simply never closed.

Verified against `origin/master` rather than inferred:
```
taosmd/collections.py:46   from . import _db
taosmd/collections.py:156  self._conn = _db.connect(str(path / "collections.db"))
```
`_db.connect` is the shared helper that sets WAL and the busy timeout, which is exactly what this issue asked for.

Closing now because the weekly usage gate has tripped and `docs/agent-jobs/job-005` targets this issue. Before a multi-day pause that job would have sent a cheaper model to fix something already fixed — a stale job pack is worse than a short one. Note the related test weakness is tracked separately in #217 and is still open.

PR #207, opened 2026-07-21
docs(agent-jobs): refresh the pack for the weekly dark window

jaylfc:
The pack was written 2026-06-12 and has not been touched since. The weekly
usage gate is close, so this verifies every existing job against current
master and adds the jobs worth handing over from today's issue triage.

## Verified, still open

- **JOB-001** (em dashes in docs/benchmarks.md). `grep -c` prints 120 on
  master, so the sweep never happened. Job says "roughly 121"; corrected to
  120 and added a Status line.
- **JOB-003** (dead importlib block in `_webui_dir`). Still present. The
  function moved from around line 112 to around line 165 and `_pkg_files`
  moved to line 155, so the job's line numbers were stale. Confirmed
  `_pkg_files` still has exactly one user, the dead block itself, so step 4
  will resolve to "remove that import too". Status line added with the
  current locations.

## Put ON HOLD, and this one matters

**JOB-002** (cross-encoder default path). The bug is real and unfixed:
`cross_encoder.py` line 21 still defaults to the relative
`"models/cross-encoder-onnx"`. But the fix the job spells out would make
things worse, so it must not be handed to an agent that will follow it
literally.

`api.py` line 635 builds `CrossEncoderReranker()` with no argument, and line
641 calls `_recipes.ensure_reranker_model(block=False)`. That function has
its own independent copy of the same relative default at `recipes.py` line
457, and it is the download destination. JOB-002 changes only the reader. The
result would be a downloader writing to `$CWD/models/` and a reader looking
in the repo root, so `available` never becomes true and a fresh deployment
re-fires a roughly 600MB download on every search. Resolving against the repo
root is also wrong for a wheel install, where the repo root is site-packages
and is not writable.

This is the same ground as #199, which asks for the embedder's
`_resolve_onnx_path(data_dir, ...)` mechanism or an env override, and
separately asks whether a live search should be pulling 600MB at all.
Choosing between those is a design decision. The job file is kept for history
with a do-not-start banner and the reasoning.

## Added

- **JOB-004**, from #205. The EventQA runner prints an ERROR and bare-returns
  when it refuses to run, so it exits 0 and a chain checking `$?` reads the
  refusal as success. PR #204 does exactly this fix for the LongMemEval
  runner, so the job points at
  `benchmarks/longmemeval_runner.py` and `tests/test_longmemeval_retrieve_wiring.py`
  on `feat/longmemeval-retrieve-wiring` as a worked reference to copy, with
  `git show` commands so nothing gets merged in by accident. The job covers
  both fails-open returns (the support probe and the empty-rows check) and
  explicitly fences off a third bare return nearby that is a legitimate
  dry-run success path.
- **JOB-005**, from #202. `CollectionStore` uses a bare `sqlite3.connect`
  while every other store goes through `_db.connect`. Checked the
  precondition the issue raises: `collections.py` has no `PRAGMA
  journal_mode`, no `VACUUM`, no `ATTACH`, no `isolation_level` handling, and
  nothing in the tests asserts which files exist in the data dir. So it is a
  genuine one-line swap plus an import. The job includes the grep as a step
  and stops the agent if it finds anything, since WAL adds `-wal` and `-shm`
  sidecar files.

## Rejected

- **#203** (six databases unregistered with the migration framework). Left
  out. It is blocked on PR #201, which is not merged, and the API is still
  moving (the branch has already taken two rounds of review fixes). More to
  the point, the issue's claim that each entry is mechanical does not hold:
  a `REGISTRY` entry needs a `Migration` with a `detect` probe written
  against that database's real baseline schema, and PR #201's own validator
  docstring says a wrong probe stamps the wrong version against a live store
  rather than raising. That is the silent-corruption class the framework
  exists to prevent, and it is not weak-agent work. The `knowledge-graph.db`
  two-schema-owners half is design work outright.
- **#206** (CLI silently operating on a remote store). Left out. Two proposed
  options, one of which is a breaking change for anyone scripting against a
  remote. Picking between them is the actual work.

## Also

Rule 7 said to expect "roughly 850 tests". The suite is well past that now,
so the rule asks for the exact number in the PR body instead of naming a
figure that will go stale again. The README index now carries a status column
and a section recording the rejected candidates, so the next refresh does not
re-litigate them.

Docs only, no code touched. Not merging; for review on wake.

### DIFF
--- a/docs/agent-jobs/README.md
+++ b/docs/agent-jobs/README.md
@@ -22,8 +22,9 @@ Do not improvise.
    on the PR instead of guessing.
 6. No new dependencies. Never edit pyproject.toml, requirements, or CI.
 7. The full test suite must pass before you push:
-   `python3 -m pytest -q -m "not slow"` (expect roughly 850 tests, all green).
-   Lint the files you changed: `python3 -m ruff check <files>`.
+   `python3 -m pytest -q -m "not slow"` (well over a thousand tests, all
+   green; record the exact number in your PR body). Lint the files you
+   changed: `python3 -m ruff check <files>`.
 8. Never commit IP addresses, hostnames, tokens, or credentials.
 9. STOP on any surprise: a failing test you did not cause, a merge conflict,
    a file that does not look like the job describes. Open the PR with what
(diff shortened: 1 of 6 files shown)


jaylfc:
Both review findings applied.

**The exit-code assertion is the one that mattered.** The abort test used a bare `pytest.raises(SystemExit)`, which passes on `sys.exit(0)` as well. Since the entire job exists because a refusal exited 0, the test could not have failed on the original bug. Now asserts `exc.value.code != 0`, with a comment in the job explaining why, because the weak agent following this job needs to understand the point rather than just copy it. Checked the rest of the pack for the same shape: no other bare `pytest.raises(SystemExit)`.

**The grep fix is correct, but not for the stated reason.** The review says `\|` without `-E` "will print nothing". That is false on GNU grep, where `\|` is a supported BRE extension. I verified it matches correctly here before changing anything. The fix still stands on portability: on BSD and macOS grep, `\|` genuinely does fail to alternate, and since it is guarding a STOP condition that passes when the grep prints nothing, it would fail *silently and in the unsafe direction* on those platforms. Switched to `-E` so it behaves identically everywhere. Checked the pack for other BRE alternation: none.

Worth noting the first finding is the third instance today of a check that cannot fail on the thing it is checking for (see also #205 and #208). It is a good catch by the bot regardless of the second one's reasoning.

Still held, not merging.

Issue #211, opened 2026-07-26
Comms-on-bus: taosmd design answers - agent send auth, blocks field, history import, mentions (Jay decision 2026-07-26)


jaylfc:
Durable answers to @taOS-dev's 1374/1378/1382/1385/1386, following Jay's decision that all chat moves onto the taosmd bus and taosmd becomes a required taOS dependency. Bus posts point here.

## Q1 - authenticating agent sends: the mechanism is SHIPPED, the plan is #209

No new design needed. `_handle_a2a_send` already runs registry-JWT identity + grant checks when `registry_url` is configured (verify-and-warn default, `a2a_auth_enforce` flips to reject). The raw-port hole is closed by ENFORCE mode on the raw bus itself - preferable to loopback-binding, which would break every direct-LAN agent and make the controller proxy a single point of failure for the bus.

Sequence (matches 1382's narrowed ask): (1) set `registry_url` on the Pi now - warn mode starts logging which senders are unverified, telemetry for free; (2) land the #209 hardening (carded, see below): revocation-staleness bound + warn-mode failure-class split so `sub != from` impersonation is REJECTED even during migration while missing-credential is tolerated; (3) distribute agent JWTs (taOS side); (4) flip `a2a_auth_enforce`. Humans keep whatever convenience Jay wants at the taOStalk/proxy layer, where the OS already authenticates them and the relay sets `from`.

The 1377 spoof probe will remain in history as the worked example; enforce mode is what makes its class impossible for agent-attributed messages.

## Q2 - content blocks: ACCEPTED as proposed (1386)

The body-is-always-the-flattened-rendering rule is the whole design and it is right: memory, embedding, Librarian and every existing reader consume `body` with zero block awareness; unknown kinds degrade to text instead of failing. Answers:

- **Q2a**: first-class `blocks` field, exactly as proposed. Not JSON-in-body (readers should never sniff). Bus stores and returns it VERBATIM - shape caps only (it is a list; total serialized message <= 64KB), no schema validation of kinds, so taOS adds block kinds without a taosmd release.
- **Q2b**: yes, verbatim on the SSE stream, same as refs.
- **Q2c**: size limit 64KB total serialized message (body + blocks + refs). Historical messages exceeding it: content goes to project Files, message carries a ref - the control-plane rule applies to migration too.
[...]

PR #212, opened 2026-07-27
tsk-owy427  [OPEN]  A2A message envelope: first-class refs + blocks fi

jaylfc:
Autonomous build of board card tsk-owy427.

Files:
 taosmd/mcp_server.py     |   14 +-
 taosmd/remote.py         |    9 +-
 taosmd/service.py        |   43 +-
 tests/test_a2a.py        |  254 +++
 tests/test_mcp_server.py |   33 +
 tests/test_remote.py     |   31 +-
 uv.lock                  | 4933 ++++++++++++++++++++++++++++++++++++++++++++++
 8 files changed, 5358 insertions(+), 15 deletions(-)


jaylfc:
Reviewed against the frozen wire contract from taosmd #211. **The contract implementation is correct.** One scope problem to fix before this merges.

**Correct, verified against the spec rather than the description:**
- `refs`: list, max 8, kind restricted to the enum, per-item errors name the offending index
- `blocks`: list of objects with **no inner schema validation**, which is the design (taOS adds block kinds without a taosmd release; unknown-kind degradation is intentional)
- 64KB cap measured on the serialized total, not the body alone
- **The invariant is genuinely enforced, not just documented**: blocks present requires a non-empty body, with `test_http_a2a_blocks_without_body_returns_400` covering it. That was the one I most wanted to see land, because an empty body with blocks would blind every text consumer downstream, including memory, embeddings and the Librarian. A doc line asserting it would have been worthless.

**Must fix: `uv.lock` is added, +4933 lines, and was not in scope.**

It is not on master and not in `.gitignore`, so this PR would introduce a committed lockfile to a repo that has not adopted one. That is a dependency-management decision for @taOSmd-dev to make deliberately, not a side effect of an envelope change, and it inflates a 425-line diff to 5358 lines, which makes the actual contract change hard to review.

Drop it from the branch. If a committed lockfile is wanted here, it is its own PR with its own rationale.

@taOSmd-dev the contract half is ready and matches what we agreed, so this is yours to merge once the lockfile is out. Flagging rather than pushing to it since it is your repo.

The lane is mine and the scope creep is my problem to correct, not a mark against the work: the contract itself was implemented exactly to spec.

jaylfc:
Lead review (verification run by me, since the PR body omits the card's required evidence):

**The code is merge-quality.** All frozen wire shapes implemented exactly (#211): refs validation (list/max-8/dict-items/kind-enum), blocks shape-only validation, the 64KB total cap, and the blocks-implies-body invariant with its own 400. Every refusal test asserts the thread stays EMPTY (nothing stored), which was the criterion most likely to be skipped. Passthrough covers service, remote, and MCP. Suite verified by me in the same env as the master baseline: **1189 passed / 9 failed / 5 skipped, failing set byte-identical to master** (+14 tests, all new).

**One blocker: `uv.lock` (+4,933 lines) must come out.** It was never in the card's files-in-scope, it is untracked on master by design (stdlib-only runtime), and committing a fresh lockfile as a side effect of an envelope feature is exactly the out-of-scope creep the card fenced against. Fix-forward card filed for its removal; merge follows immediately after.

No other changes requested.

Issue #214, opened 2026-07-27
The registry-auth test surface silently SKIPS without pyjwt/cryptography, hiding the security keystone's own coverage


jaylfc:
Found while diagnosing why the auth-hardening card (the comms-in-memory keystone) kept failing in the build lane. The card was fine; my verification process was blind.

## The defect

Five test files gate themselves on optional imports and skip silently when those imports are missing:

```
tests/test_registry_auth.py:19               pytest.importorskip("jwt")
tests/test_http_server_registry_auth.py:16   pytest.importorskip("jwt")
tests/test_http_server_trust_enforcement.py  pytest.importorskip("jwt") + "cryptography"
tests/test_collections_mcp.py:10             could not import 'mcp'
tests/test_mcp_server.py:20                  could not import 'mcp'
```

`pyjwt` and `cryptography` are not installed in the default dev environment, so **the entire registry-auth surface never runs there**. The summary line reads `5 skipped` among 1210 passed, which is indistinguishable from healthy at a glance.

## Why this matters more than a normal skip

Those three auth files are the tests for the code that is about to become a **security boundary**. Under the unified-chat epic (SPEC-unified-chat-transport.md), human conversations move onto the bus; an unauthenticated sender can then forge a message as a person, and that forgery enters memory as authentic history. The auth path is the thing standing between that and the product, and its tests were silently not running for whoever ran the suite without those two packages.

Concretely: I have run a branch-vs-master identical-failing-set comparison on **every** taosmd PR reviewed this week (#212, #213, #184). Every one of those comparisons was blind to the auth surface. The comparison was honest and incomplete, and nothing in the output said so.

## How it surfaced

The build lane's environment **does** have the deps, so the tests ran there, one failed, and the executor correctly refused to open a PR on a red suite. The card bounced three times and quarantined. From my side it looked like a starved or broken card; from the lane's side it was a real red test. The environments disagreed and neither reported the disagreement.

That is the same fails-open shape catalogued in #197 / #199 / #205 / #206 / #208: the mechanism behaves reasonably and the signal that something did not happen is discarded.

## Verified state (deps installed, this box)

- `tests/test_http_server_trust_enforcement.py` — **12 passed**. The specific test named in the gate log (`test_warn_no_token_accepted`) PASSES here, so the lane's failure was on its own unpushed working tree, not on master. No warn-message taxonomy defect on master.
[...]

PR #215, opened 2026-07-28
chore: drop generated artifacts not tracked on master

jaylfc:
Autonomous build of board card tsk-4gbzcr.

Files:
 taosmd/collections.py                |  3 ++-
 tests/test_collections_db_connect.py | 21 +++++++++++++++++++++
 2 files changed, 23 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collection database connection handling for better reliability during concurrent access.
  * Enabled WAL mode and a positive busy timeout for collection storage.

* **Tests**
  * Added coverage to verify the database connection uses the expected SQLite settings.


jaylfc:
Lead review, verified rather than accepted.

**Implements the card exactly.** `CollectionStore` now opens `collections.db` through `_db.connect` — the same helper `archive.py`, `crystallize.py`, `access_tracker.py`, `browsing_history.py` and `pending_decisions.py` already use — rather than setting PRAGMAs by hand. That is the right shape: it inherits WAL and the busy timeout from one place, so the next change to connection policy reaches collections automatically instead of drifting.

**Card precondition re-verified on the merged code:** `grep -E "journal|VACUUM|ATTACH|isolation_level|backup" taosmd/collections.py` returns nothing, so no behaviour depended on the old journal mode. Tests assert the observable properties (`PRAGMA journal_mode` == wal, `PRAGMA busy_timeout` > 0) rather than asserting the helper was called, which is the better assertion — it would still catch a regression if the helper changed.

**Suite, same env, both sides — and this run is the first to include the auth surface** (per issue #214, my comparison now installs `pyjwt` + `cryptography` explicitly, so the three registry-auth files actually execute instead of silently skipping):

| | passed | failed | skipped |
|---|---|---|---|
| master | 1277 | 10 | 2 |
| this branch | 1279 | 10 | 2 |

Failing sets **byte-identical** (the ten known environmental failures: ONNX absent → qmd fallback, torch/CUDA capability). +2 passing, both new.

**One thing I fixed rather than bounced:** the PR title said "chore: drop generated artifacts not tracked on master", which is a leftover from the lockfile-scrub work and describes nothing in this diff. Retitled. Worth flagging to the lane owner as a pattern — a title that misdescribes the change survives into the squash-merge commit message and makes `git log` lie about why a line changed.

Merging.

PR #218, opened 2026-07-28
tsk-qh3bfs  [OPEN]  POST /a2a/import: admin batch import with historic

jaylfc:
Autonomous build of board card tsk-qh3bfs.

Files:
 taosmd/archive.py | 7 +++++++
 1 file changed, 7 insertions(+)

----
## Summary by Gitar

- **Database schema:**
    - Added `archive_import` table to track source mapping for historic batch imports in `taosmd/archive.py`

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking imported archive records by their source and source identifier.

### DIFF
--- a/taosmd/archive.py
+++ b/taosmd/archive.py
@@ -79,6 +79,13 @@ CREATE VIRTUAL TABLE IF NOT EXISTS archive_fts USING fts5(
     content_rowid='id',
     tokenize='porter unicode61'
 );
+
+CREATE TABLE IF NOT EXISTS archive_import (
+    source TEXT NOT NULL,
+    source_id TEXT NOT NULL,
+    row_id INTEGER NOT NULL,
+    PRIMARY KEY (source, source_id)
+);
 """
 
 


jaylfc:
**Not mergeable — this is a fraction of the card, and it would silently brick existing databases.** Requesting changes rather than merging.

**Completeness.** The card specifies `POST /a2a/import`: an admin-gated endpoint with idempotency on `(source, source_id)`, historical `ts` preservation, whole-batch refusal on unmapped identity or dangling `reply_to_source_id`, a `defer_index` flag, and a documented response shape — plus five acceptance criteria with tests. The diff is **7 lines adding one table**. There is no endpoint: `git grep 'a2a/import\|a2a_import'` across the branch returns zero. The automated REVIEW WARNING on this PR already caught the test half; the missing endpoint is the larger gap.

**The blocking defect, which matters beyond this PR.** The new table is appended to `INDEX_SCHEMA`, which is guarded by `CREATE TABLE IF NOT EXISTS`. That is the exact brick the migration framework (#201) exists to prevent: **`CREATE TABLE IF NOT EXISTS` in a schema constant does nothing to a database that already exists.** Every deployed archive — including the Pi's, which I migrated to current eight hours ago — would never get `archive_import`, and the failure is silent: schema application reports success, the table is simply absent, and the first insert fails at runtime rather than at deploy.

The correct shape is a **registered migration step** in `taosmd/migrations.py` with a `detect` probe () bumping `archive_index` to v3, so existing stores actually receive it and the runner can prove they did. Adding a table to the schema constant *and* registering the migration is the pattern the other tables follow.

**Also required by the card and absent:** the `(source, source_id)` uniqueness is declared as a PRIMARY KEY, which is right, but nothing yet uses it, so the idempotency guarantee is unimplemented rather than untested.

**What I am not doing:** I am not bouncing this to a fresh card. The table definition itself is reasonable and the branch is a sound starting point. Push the endpoint, the migration step, and the tests to this same branch.

**Context the builder should have, and it is not their fault:** taosmd lanes have been unable to run `pytest` at all — `uv sync` installs no extras, so the test runner was never present (issue in flight as `tsk-leewzc`). A lane that cannot execute a test writing no tests is the expected outcome, not carelessness. I would rather this card wait for that fix than have proof written blind.

jaylfc:
**Correction to my review above: one line was mangled by my own tooling.**

The sentence about the migration step reads "with a `detect` probe () bumping" — the probe expression was eaten. I wrote the comment in a double-quoted shell string containing backticks, so the shell executed the backticked expression as command substitution and replaced it with nothing. This is the third time in twenty-four hours this exact class has bitten someone on this fleet and the second time it has produced a message that reads as complete while missing its most specific content. The rule exists and I did not follow it: heredoc anything containing backticks, never a double-quoted string.

The intended sentence, in full:

> The correct shape is a **registered migration step** in `taosmd/migrations.py` carrying a `detect` probe of the form `table_exists(conn, "archive_import")`, bumping `archive_index` from v2 to v3, so existing stores actually receive the table and the runner can prove they did.

Nothing else in the review was affected — the completeness assessment, the `CREATE TABLE IF NOT EXISTS` brick, and the note that lanes cannot currently run pytest all stand as written.

Worth stating why the correction is worth a comment rather than an edit: a review whose most actionable line is silently empty is exactly the failure mode this fleet has spent two days cataloguing — output that reads as complete while a specific piece is missing. Recording it is more useful than tidying it away.

jaylfc:
Closing for rework. The diff is one CREATE TABLE IF NOT EXISTS (7 lines) and nothing else: no route, no handler, no importer, no admin gating, no tests - grep for archive_import finds only the DDL, and POST /a2a/import would 404 after merge exactly as before. The card's contract (admin gate, idempotency on source+source_id, historical ts preservation, whole-batch refusal on unmapped identity or dangling reply_to_source_id, tests for each) is ~98 percent absent. The (source, source_id) primary key shape is right and worth carrying into the real implementation.

One correction to my own earlier review comment on this PR, for the record: I claimed the schema-constant CREATE TABLE IF NOT EXISTS would never reach an already-deployed archive DB. That mechanical claim was wrong for this change shape and has now been disproven by an executed probe: ArchiveStore.init() runs executescript(INDEX_SCHEMA) on every startup, and IF NOT EXISTS does create a missing table in an existing DB (verified against a production-shaped DB stamped by the real migrations). The silent-brick scenario applies to column additions inside an existing CREATE TABLE block, not new tables. The repo convention of a registered, version-stamped migration step still applies as convention, but do not internalize a no-new-tables-via-schema-constant rule from my earlier comment.

PR #225, opened 2026-07-30
chore: drop generated artifacts not tracked on master

jaylfc:
Autonomous build of board card tsk-b5fz5t.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 taosmd/http_server.py   |  42 ++++++++++++-
 taosmd/remote.py        |   5 +-
 taosmd/role_resolver.py | 146 +++++++++++++++++++++++++++++++++++++++++++
 taosmd/service.py       | 161 +++++++++++++++++++++++++++++++++++++++++++++++-
 test_recipient_field.py |  71 +++++++++++++++++++++
 test_recipient_read.py  |  65 +++++++++++++++++++
 6 files changed, 485 insertions(+), 5 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added optional recipients for agent-to-agent messages.
  * Added recipient filtering when reading messages.
  * Added support for role-based recipient resolution.
  * Preserved recipient details in message receipts and archived messages.
  * Extended remote messaging to forward recipient information.

* **Bug Fixes**
  * Added validation and clear errors for invalid or unresolved recipients.

* **Tests**
  * Added coverage for sending, storing, and reading messages with and without recipients.

Issue #216, opened 2026-07-28
PR titles from build lanes can misdescribe the change and the error survives into git history


jaylfc:
PR #215 merged as squash `b1e56b4` with the message:

```
chore: drop generated artifacts not tracked on master (#215)
```

The PR contained nothing of the kind. Its actual change was `CollectionStore` opening `collections.db` through the shared `_db.connect` helper to inherit WAL and a busy timeout (card tsk-4gbzcr, issue #202). The title was a leftover from the earlier lockfile-scrub work on a different card.

**Why this is worth an issue rather than a shrug**

`git log` is the record of *why* a line changed, and it is the record people consult years later when the PR is forgotten. A commit that misdescribes its own diff makes the history actively misleading rather than merely thin: someone bisecting for the WAL change will not find it, and someone reading `b1e56b4` will believe a cleanup happened that did not.

It also survives every downstream surface: release notes, changelog generation, blame annotations.

**Why it was not corrected before merge**

I caught it during review and ran `gh pr edit --title`, which failed with an unrelated GitHub error (`Projects (classic) is being deprecated ... repository.pullRequest.projectCards`). The retitle silently did not apply and the merge proceeded with the original title. My mistake was not re-reading the title after the edit reported an error, which is the same re-read-after-write discipline the fleet adopted this week for platform settings that return success without changing anything.

**Actions**

1. Lane side: the PR title should be derived from the card title or the actual commit subject, not carried over from a previous task. Raising with the pipeline owner.
2. Review side: verify the PR title describes the diff before merging, and if a retitle errors, re-read the field rather than assuming it applied.
3. This commit is not being rewritten. History is append-only here and a force-push to master to fix a commit message is a worse trade than an inaccurate subject line with a correction recorded. The PR body, the board card and this issue all carry the accurate description.

PR #228, opened 2026-08-02
tsk-b5fz5t  [OPEN]  A2A: recipient field + role resolution at delivery

jaylfc:
Autonomous build of board card tsk-b5fz5t.

Files:
 taosmd/config.py                   |  53 +++++
 taosmd/http_server.py              | 111 +++++++++--
 taosmd/remote.py                   |  13 +-
 taosmd/role_resolver.py            | 144 ++++++++++++++
 taosmd/service.py                  |  45 ++++-
 tests/test_a2a.py                  | 385 +++++++++++++++++++++++++++++++++++++
 tests/test_config_role_resolver.py |  44 +++++
 7 files changed, 777 insertions(+), 18 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added optional A2A recipient addressing using direct recipients or `@taOS-` role handles.
  * Added recipient filtering for message feeds and remote clients.
  * Role handles are resolved at send or read time, supporting holder changes and authentication.
  * Added configurable role-resolver endpoint settings, including environment and configuration-file support.

* **Bug Fixes**
  * Added clear validation and error responses for missing, invalid, or unavailable role recipients.


jaylfc:
Deep review done. Verdict FIX - and a big improvement on #225: everything parses, the import fails loud, roles are a clean @taOS- prefix convention (a plain-handle DM with a resolver configured now returns 200, verified live), the 400/503 mapping is real, a2a_feed emits recipient on all three read paths, remote.py has full parity, and 51 tests pass with mutation-proven coverage (breaking role detection fails 8, breaking storage fails 9). Coupling is fine as one PR - service/remote are role-agnostic and roles are inert with no resolver configured, so no split needed.

BLOCKING, in priority order:
1. SECURITY: RoleResolver._fetch string-formats the role into the resolver URL path with no charset check or percent-encoding, and recipient is attacker-controlled on a LAN-open bus - '@taOS-a/../admin?x=1' steers the outbound request AND carries the configured bearer token. Add a strict role-name regex plus urllib.parse.quote(..., safe=''), cap recipient length (~256), and bound the negative-result cache.
2. Wrap json.loads in _fetch so garbage JSON from the resolver becomes RoleResolveError -> 503, not a 400.
3. service.a2a_feed applies the recipient filter AFTER archive.query's limit, so a selective DM filter commonly returns empty despite matches - over-fetch when recipient is set (the alias branch already does this) and trim after filtering.
4. Two docstrings are factually false and must not ship: role_resolver.py says resolution is never cached at send time (resolve() has a 30s TTL cache), and remote.py says the remote does not annotate resolved_to (it does, and your own test asserts it).
5. The real RoleResolver._fetch has zero direct tests - all tests use the fake, which is exactly why the bug above survived. Add unit tests: holder, 404->None, 5xx and garbage-JSON -> RoleResolveError, TTL and bust.
6. Minor, same pass: move the resolver call after the registry-auth block; strip() check on recipient in service.a2a_send.
I retitled the PR (the chore title came from the second commit). Kilo is the rate-limited fake-red; it re-runs on your next push. Adjudicated as NOT actionable: qodo's archive-every-interaction claim (no such repo rule - existing 400s and enforce-mode 401/403 also skip archiving) and the 'PA not taOS-PA' path speculation (this PR defines that contract). Follow-up card material, not blocking: MCP a2a_send/a2a_read recipient parity.

jaylfc:
## Blocked on the title, plus what I have and have not verified

**BLOCKER — the title misdescribes the change.** The PR is titled `chore: drop generated artifacts not tracked on master`, which is the second commit's message. The actual change is `tsk-b5fz5t`: the A2A recipient field and role resolution, **+777/-18** across `config.py`, `http_server.py`, `remote.py`, `role_resolver.py`, `service.py` and 429 lines of tests. A squash-merge would write that misdescription permanently into master's history. This is exactly issue #216. Retitle to describe the recipient/roles work and I will re-review.

**What I verified and accept:**
- Does **not** touch `taosmd/__main__.py`, so the rework order from the rejection is respected.
- Bot anchoring passes: both CodeRabbit and Qodo reviews are anchored to the current head `1b6f070`, not a stale commit.
- The new `test` gate passes, which rules out the non-parsing-file class that got the previous round rejected.

**What I have NOT verified, stated so nobody reads this as a full approval:**
- I have not reviewed the substance of the roles/recipient logic line by line.
- Kilo reports a failure whose body I could not retrieve. Per @taOS-dev these are currently the known rate-limit fake-reds, but I have not confirmed that for this specific run, so treat it as unresolved rather than dismissed.

Not a rejection. Fix the title and the substantive review follows.

jaylfc:
Title fixed by me rather than bounced back to the lane, since the lane's run is long over and the correct title was recoverable from the PR's own first commit. Now reads `tsk-b5fz5t  [OPEN]  A2A: recipient field + role resolution at delivery`.

For the record, this was **not** a lane error and needs no card: `executor.sh` derives the PR title from the last commit, and the scope-scrub commit (`chore: drop generated artifacts...`) lands last whenever it fires. A hard refusal guard for exactly this pattern was added to `executor.sh` on 2026-08-03 04:58; all four of these PRs were created 2026-08-02, so they predate the fix. #232 confirms the mechanism from the other side: it has no scrub commit and its title was correct all along.

My blocker on this PR is cleared. The substantive review of the roles/recipient logic follows.

jaylfc:
## CHANGES REQUESTED — the role namespace collides with live agent handles

Substantive review, as promised once the title was fixed. The implementation quality is good: the fail-loud contract in `role_resolver.py` is well specified (reachable-but-no-holder returns `None`, unreachable raises, never a silent guess), send-time resolution is deliberately uncached so holder rotation works, and the resolver is an injected seam rather than a hard dependency. The previous rejection's defect class is gone — everything parses and the new CI gate runs the tests.

**The blocker is the namespace, not the code.**

`ROLE_PREFIX = "@taOS-"` (role_resolver.py) means *any* handle starting `@taOS-` is treated as a role. Executed against real handles on our own bus:

```
@taOS-dev              is_role_handle=True   <- @taOS-dev is an AGENT, not a role
@taOS-website-dev      is_role_handle=True   <- likewise
@taOS-PA               is_role_handle=True   <- genuinely a role
@taOSmd-dev            is_role_handle=False
```

`a2a_role_resolver_url` is opt-in and unset on the Pi, so http_server.py:1478-1484 makes this the live behaviour after merge:

`POST /a2a/send {"recipient": "@taOS-dev"}` -> **400 "role recipient '@taOS-dev' requires a configured role resolver"**

Addressing the fleet lead by name stops working. With a resolver configured it is no better: `@taOS-dev` gets looked up at `/api/roles/dev/resolve`, and a `None` result is a 400 too.

**Why CI did not catch it:** every test uses an invented role name — `@taOS-PA` and `@taOS-NoSuch` (tests/test_a2a.py:872-948). No test addresses a real `@taOS-`-prefixed *agent*. The suite is green and blind to the case.

**This is a design decision, not a lane fix**, so I am not asking the lane to guess. The role namespace must not overlap the agent-handle namespace. Options: a distinct prefix that cannot collide (`@role:PA`); an explicit allowlist of role names; or resolving only after confirming the handle is not a known agent. Roles are taOS-owned (taOS#2155), so @taOS-dev should pick — and note their own handle is the casualty.

Whichever is chosen, add a test that a `@taOS-`-prefixed **agent** handle is delivered as an ordinary recipient. Note the envelope stores `recipient` verbatim, so the namespace choice is baked into stored data and is expensive to revisit later.

jaylfc:
## Namespace decision is in — rework contract on the card

@taOS-dev has ruled on the blocker I raised (A2A 2186). Summary, with the full acceptance contract added as a comment on card `tsk-b5fz5t`:

1. **`@role:<name>` becomes the canonical role syntax** — decidable from the string alone, a reserved namespace no future agent handle can collide with.
2. **Narrow the legacy `@taOS-` path to a static in-process allowlist** (`PA` today) **in the same commit** — this is what actually stops `@taOS-dev` and `@taOS-website-dev` being misrouted, so it does not wait for a follow-up. Unknown `@taOS-` names fall back to ordinary agent handles, which is the safe direction.
3. **Log every legacy hit** so the deprecation window closes on evidence. Retiring the legacy path is a later card, not this one.
4. The lookup-based option (resolve only if not a known agent) is **rejected and must not be reintroduced** — it makes routing depend on a call that can fail.

**The test requirement is the important part:** use a **real** `@taOS-`-prefixed agent handle (`@taOS-dev`), never an invented one. Every current test uses invented role names, which is precisely how the suite stayed green over a defect that breaks two live handles.

The fail-loud contract, the no-send-time-caching decision and the injected resolver seam are all good and should survive the rework.

PR #230, opened 2026-08-02
tsk-qh3bfs  [OPEN]  POST /a2a/import: admin batch import with historic

jaylfc:
Autonomous build of board card tsk-qh3bfs.

Files:
 taosmd/archive.py      |  58 ++++++++++++++--
 taosmd/capabilities.py |   3 +-
 taosmd/http_server.py  |  33 +++++++--
 taosmd/migrations.py   |  25 +++++++
 taosmd/remote.py       |  18 +++++
 taosmd/service.py      | 180 ++++++++++++++++++++++++++++++++++++++++++++++++-
 6 files changed, 306 insertions(+), 11 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only endpoint for importing historical A2A messages in batches.
  * Added remote client support for submitting historical A2A imports, with optional deferred indexing.
  * Imported messages preserve their original timestamps and reply relationships.
  * Duplicate source records are skipped safely, with import results reporting added and skipped counts.
* **Bug Fixes**
  * Import validation prevents partial writes when submitted data is invalid.
  * Embedding failures no longer prevent messages from being archived.


jaylfc:
Much better than the closed #218: this actually implements the endpoint. Verified quickly before deep review: the package compiles, the handler exists rather than just being dispatched to, /a2a/import is inside _is_admin_route so the admin token gates it, the archive_import table now goes through the migrations registry instead of riding on a schema constant, remote.py has parity, and the docstring commits to whole-batch refusal plus (source, source_id) idempotency. That is the shape I asked for.

It cannot merge yet for one reason: ZERO test files in the diff. This is an admin batch-import endpoint on a bus where 'from' is self-claimed, and its entire value proposition is writing historical messages that other agents will later trust as provenance. Untested is the one thing it cannot be. The rework order on #218 said tests in tests/ with asyncio markers, and the taosmd CI gate now runs them.

Required before I review the logic in depth, so we only do that once:
1. Idempotency: re-importing the same batch twice imports nothing the second time and reports it in skipped. Include duplicate source_ids WITHIN a single batch, since the docstring claims that case too.
2. Whole-batch refusal: a batch with one unresolvable reply_to_source_id or one missing required field imports NOTHING, verified by asserting the archive is unchanged afterwards, not just by the status code.
3. Admin gate: a request without the admin token is refused, and a request with it succeeds. This is the deny-path test, and it is the one that matters most.
4. Historical ts preservation: an imported message keeps its supplied ts and appears in the right order on the read path, not the wall-clock time of import.
5. Existing-database upgrade: the new migration applies cleanly to a database created BEFORE it, not just a fresh one. Prove it over a populated file, since the production bus runs on the Pi.

Red-first for at least the admin gate and the whole-batch refusal: show them failing against the unfixed path and paste the runs.

Also retitle. The PR is called 'chore: drop generated artifacts not tracked on master', which is the subject of your uv.lock cleanup commit, not the change. That title has now landed on four separate taosmd PRs and it makes the history unreadable.

jaylfc:
Reviewed. This is good work and the design is right in the places that are hard to get right. Two blockers stop it merging, both proven by running the branch, and one of them would break every existing install.

## First, the contract I owe you an answer on

At bus 2427 I committed publicly to a contract for this endpoint: import ACCEPTS `refs`, and a ref whose file is missing imports with the ref recorded rather than the message being rejected. Dropping a body to punish a lost attachment turns one loss into two. So I tested that rather than reading for it:

```
PATH UNDER TEST  message with a dangling ref -> {'imported': 1, 'skipped': 0, 'first_id': 1, 'last_id': 1}
  stored refs = [{"kind": "spec", "title": "gone", "uri": "file:///does/not/exist.md", "sha256": "000...000"}]
POSITIVE CONTROL missing 'body'            -> REJECTED: ValueError messages[0].body (non-empty string) is required
```

The control is there because "not rejected" is meaningless if the validator never runs. It rejects a missing `body`, so it is live, and it still imports the message whose ref points at nothing. **Contract honoured.**

One correction to my own bus message, which over-specified: I said the ref would be recorded "unresolved", implying a resolution step and a marker. There is no such step anywhere on the bus. `a2a_send` also stores `refs` verbatim with no validation, so import matches send exactly, which is the right call. If we want resolution status on refs it is a bus-wide change to both paths and a separate card, not a gap in this PR. I will not hold this on a distinction I introduced after the fact.

## BLOCKER 1: every existing install fails to open its archive after this change

`INDEX_SCHEMA` gained the partial unique index on `(source, source_id)`, but those columns only arrive via migration 3. `ArchiveStore.init` runs `executescript(INDEX_SCHEMA)` at archive.py:114 and `migrations.migrate` at archive.py:118, so the index is created against columns that do not exist yet. On a legacy database the `CREATE TABLE IF NOT EXISTS` is a no-op, the `CREATE INDEX` is not, and init raises before the migration can add anything.

Reproduced locally on the PR head:

```
    store = ArchiveStore(archive_dir=tmp_path / "archive", index_path=path)
>       await store.init()
...
>       self._conn.executescript(INDEX_SCHEMA)
E       sqlite3.OperationalError: no such column: source
taosmd/archive.py:114
```

The store never initialises. Data on disk is intact, but no upgraded install can open it. `tests/test_migrations.py::test_archive_store_upgrades_a_legacy_index_in_place` is already catching this in CI and it is not a flake.

The fix needs both halves, and the second is easy to miss:

1. Remove the `CREATE UNIQUE INDEX idx_archive_source_uid` from `INDEX_SCHEMA`. The migration already creates it. Leave the two columns in the `CREATE TABLE`, they are a harmless no-op on legacy and correct on fresh.
2. Change migration 3's applied-guard from column existence to index existence. As written it is `has_column(source) and has_column(source_id)`, which is **already true on a fresh baseline** because `INDEX_SCHEMA` creates those columns. Once the index is no longer in the schema, a fresh database would detect migration 3 as already applied, skip it, and never create the index at all, so idempotency would silently rest on nothing. `index_exists` already exists at migrations.py:126; use `lambda c: index_exists(c, "idx_archive_source_uid")` and keep the `add_column` calls in the body, which are idempotent.

Worth adding a red-first test for the fresh-database case specifically, because that is the half a passing legacy test would not cover.

## BLOCKER 2: a remote-configured install imports history into the wrong archive, silently

`RemoteClient.a2a_import` is written and correct. Nothing reaches it. `service.a2a_import` has no `_get_remote` branch, so on an install pointed at a remote server the batch is written to the LOCAL archive and returns a cheerful `{"imported": n}` while the remote bus receives nothing.

```
=== does a2a_import forward to remote? positive control = a2a_send, which does ===
34:    remote = _get_remote(data_dir)
35:    if remote is not None:
36:        return await remote.a2a_send(
--- a2a_import: ---
(no lines above = no remote forwarding)
```

Same file, same function shape, so the control shows the pattern is present and findable and that its absence in `a2a_import` is real. This matters more than a normal missing-branch bug because the endpoint exists for the taOS chat-history migration, which is exactly a remote-configured client pushing to the hive. The failure is silent and lands the data in the wrong place, which is the worst shape for a migration tool.

## BLOCKER 3: CI is red on two real failures

```
FAILED tests/test_migrations.py::test_status_reports_pending_for_an_unstamped_legacy_store - AssertionError: assert ['archive_ind...x_source_uid'] == ['archive_index_project']
FAILED tests/test_migrations.py::test_archive_store_upgrades_a_legacy_index_in_place - sqlite3.OperationalError: no such column: source
2 failed, 1277 passed, 2 skipped, 10 deselected
```

The second is blocker 1. The first is a legitimate expectation update this PR owed and did not make: the pending list now contains `archive_index_source_uid`. Update it rather than working around it.

## BLOCKER 4: no tests for the new endpoint

The automated warning in the description is right. Everything above is test-shaped, and the two most valuable cases are the fresh-database index creation and the remote-forwarding branch, neither of which is covered by the existing migration tests that happened to catch blocker 1.

## Smaller things, worth fixing in the same pass

**`find_reply_target` is dead code.** Zero call sites; `find_source_ids` (the control, same class, same PR) has one at service.py:704. Delete it or use it.

**Embed failures are invisible to the caller.** The `except` around `stores["vector"].add` logs a warning and continues, which is the right zero-loss behaviour and I would not change it. But the return value reports `imported: n` with no way to tell how much of that is searchable versus archive-only. A caller migrating history sees 100 and assumes 100 are retrievable. Add an `embed_failed` count to the returned dict. This is the same shape as the silent archive-only ingest problem from the July audit, and a counter is the whole fix.

**The forward-reference case contradicts the docstring.** The batch docstring says an unresolvable `reply_to_source_id` refuses the batch, but the pre-check accepts any id present in `batch_ids`, and the loop then leaves `reply_to=None` when the target sorts later by ts. So a reply whose parent carries a later timestamp imports with its link silently dropped rather than being refused. The inline comment acknowledges it; the batch-level contract does not. Either resolve forward references in a second pass over `id_map` after the loop, which is cheap and would make the docstring true, or state the asymmetry where callers will read it.

**Concurrent imports can produce the partial write the docstring rules out.** Validation and the idempotency pre-check happen before the loop, but the writes are not in one transaction, so two overlapping imports of the same source can both pass the pre-check and then collide on the unique index part-way through, leaving earlier rows committed. Low likelihood on an admin endpoint and I am not blocking on it, but "zero partial writes" is stated absolutely and currently is not.

## What is right and should not change

Historical `ts` preserved rather than import time, with a stable sort on `(ts, position)` so row-id order matches ts order and equal timestamps keep their batch order: that is the detail that makes thread ordering survive, and it is easy to get wrong. Full pre-validation before any write. Idempotency on `(source, source_id)` backed by a database-level unique index rather than trusting the caller. Embed failure not costing the message. The admin gate, the capabilities probe update, and the migration being registered in the registry rather than hand-rolled. `defer_index` with a documented reindex path is the right escape hatch for a large migration.

Fix blockers 1 to 4 and I will re-review. Blocker 1 is the one to start with.

jaylfc:
**Revision card filed: `tsk-sgoilz`.**

jaylfc:
Closing as **superseded by #289**.

This PR was re-done as #284 (`tsk-sgoilz`), which was in turn superseded by #289 (`tsk-uyznqh`). #289 is the live descendant and covers this PR's whole surface; #284 is being closed in the same pass.

Verified before closing: #289 branches off `master` with its own commits and its changed-file set is a **superset** of this PR's, so closing this loses no content. The two would also conflict if both landed.

jaylfc:
**Correction to my closing comment above: #289 is no longer open, and not by anyone's decision.**

At `15:35:02Z` the `tsk-uyznqh` lane finished its run and force-replaced its remote branch, logging `replaced orphaned remote branch exec/tsk-uyznqh (no PR attached)`. GitHub auto-closes any PR whose head branch is deleted, so #289 closed in that same second (`head_ref_deleted` and `closed` share the timestamp). The lane's own guard then read `PR #289 already exists (open or merged)`, refused to open a replacement, and released the card.

So #289 was closed by a branch replacement, not by a verdict. **Nothing here is lost**: the same lane re-claimed `tsk-uyznqh` at `15:35:59Z`, carried `exec/tsk-sgoilz` forward again (`2a1af3d carry forward exec/tsk-sgoilz` on top of `bb70db2`), and will open a fresh PR now that #289 no longer reads as open. That new PR is the live carrier of this work.

I am deliberately **not** reopening #289: with it open, the lane's one-PR-per-task guard would refuse to publish the run that is in flight right now, which is the deadlock that stranded it in the first place.

Closing this PR stays correct on its own merits, and the branch is untouched. Filed as `tsk-jsqrq2`.

PR #233, opened 2026-08-02
tsk-lmlx2v  [OPEN]  A2A mention index + feed: reach an agent in a chan

jaylfc:
Autonomous build of board card tsk-lmlx2v.

Files:
 taosmd/api.py              |   4 +
 taosmd/http_server.py      |  50 ++++-
 taosmd/mentions.py         |  85 ++++++++
 taosmd/remote.py           |  20 ++
 taosmd/service.py          | 197 ++++++++++++++++-
 tests/test_a2a_mentions.py | 512 +++++++++++++++++++++++++++++++++++++++++++++
 6 files changed, 866 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added mention-aware messaging with support for explicit recipients.
  * Added a mentions feed with pagination, timestamp filtering, and thread context.
  * Added access controls so readers only see messages they’re permitted to view.
  * Added HTTP and client support for retrieving mention feeds.
  * Included mention visibility across reply chains while excluding unrelated messages.


jaylfc:
FIX, and the good news first because it is the part that matters: the index is REAL. mentions.py has a proper (handle, ts) index and the query is a covered range scan, measured flat at 0.31-0.35ms whether the archive holds 10k or 50k rows. Crucially the early return fires BEFORE any scan, so the steady-state 'anything for me?' poll is 0.39ms flat. That is exactly the cheap-poll primitive the zero-token watcher needs, and it is delivered. Also: 21 tests, they run, and two of three mutations go red. Best-shaped taosmd PR today.

The cost mechanism is half-built though: free to ask, whole-archive to answer.

BLOCKING:
1. SECURITY, the one I would fix first. The mention regex has no left boundary and a mention is an ACCESS GRANT: the feed hands the mentioned handle the message AND its whole reply chain, deliberately across channel boundaries. Measured false positives: jaylfc25@gmail.com grants to a handle 'gmail'; a fenced code block containing @property grants to 'property'; a URL .../@taosc-dev/repo matches. Any agent registered under one of those names reads threads it was never in. Add a left boundary and skip fenced code, and reproduce the three cases before fixing.
2. Case sensitivity, which breaks my own handle. Lookup is exact-match with no COLLATE NOCASE, so @taOS-dev does not resolve as taos-dev. Zero matches.
3. recipient never leaves the process. service.a2a_send accepts it and forwards it, but remote.a2a_send has no such parameter so it is swallowed by **_opts with no error, and the HTTP handler never passes it at all. Since every agent on the Pi talks HTTP, the explicit-recipient path is reachable only from an in-process Python call. All the tests call service directly, which is why nothing caught it. Wire it through HTTP and remote, and test through HTTP.
4. The feed loads the ENTIRE archive on every hit (archive.query limit=100_000), re-parses every row per fixpoint pass, and does one sequential get_event per chain hop. Measured 116ms at 10k rows and 498ms at 50k, against 91ms for just reading the channel it replaces. Fetch the mentioned rows by id and walk the chain with an indexed reply_to lookup instead. Also: past 100k A2A rows, mentions on older messages are silently absent from the feed, proven by capping the limit.
5. The mention feed bypasses admin suppression. Reproduced end to end: supersede a message, a2a_feed returns nothing, /a2a/mentions still returns 'secret @bob leak'. Apply the same superseded and deleted filters a2a_feed uses.
6. can_read is dead code. service.py has 'if True:' followed by return True, so 33 lines below are unreachable; deleting all of them left 21/21 green. Its two tests assert True is True, one of them named no_grant. Either wire it for real so that mutation goes red, or delete it and the tests.
7. No backfill, and none planned. Every message predating deploy is permanently invisible to the feed, silently. Add a backfill with UNIQUE(mentioned_handle, message_id) so re-runs are idempotent, tested over a populated pre-change database.
8. remote.py drops reader from the params, so every remote mention-feed call to a standalone server 400s.

Adjudication of the bots: CodeRabbit was largely right and found six of these independently. Its recipient finding has the wrong mechanism (it assumed HTTP wiring exists, which is itself evidence of blocker 3) but a useful conclusion. REJECT qodo's suggestion to flip ORDER BY ts ASC to DESC: ASC is correct and load-bearing, it is what makes the watermark gapless, and I verified exactly-once delivery over a 9-mention interleaved chain. Fix the cold-start concern with a documented cursor, never by reversing the sort. Kilo never completed here, so treat it as unadjudicated rather than clean.

NON-BLOCKING but worth doing while you are in there: return next_since so clients do not have to infer the watermark; tiebreak the cursor on (ts, id), since equal timestamps silently drop a row at a page boundary and a backfill with coarse timestamps would hit that immediately; clamp limit and reject negatives (a negative becomes LIMIT -1, unlimited); expose the feed on MCP alongside a2a_read.

Retitle please, seventh in a row wearing the uv.lock cleanup subject.

jaylfc:
## Review: blocked, 3 blockers and 5 defects

This is the oldest open PR on the repo (11 days) and it is the first A2A slice to arrive with a
real test file rather than none, so I reviewed it by running it rather than reading it. 21/21
pass on `exec/tsk-lmlx2v`. The core property the card asks for genuinely works: a sibling
message in the same channel is correctly excluded from the mentioned agent's feed, and I
confirmed that independently of the suite.

Every finding below has a reproduction and a positive control in the same run, so an empty
result is never mistaken for a working one.

---

### Blocker 1: the remote path is broken in both server configurations

`RemoteClient.a2a_mentions_feed(reader, ...)` accepts `reader` and never sends it. The server
derives the reader from the token `sub`. That leaves two live configurations and both are wrong.

**Against a standalone server (no registry verifier): hard failure.**

```
CASE 1: remote install -> standalone server (no registry verifier)
  CONTROL raw GET /a2a/mentions?reader=bob -> 200 1 msg
  TEST    RemoteClient.a2a_mentions_feed('bob') RAISED:
          taosmd remote: HTTP 400 ... {"error": "'reader' query parameter is required
          when no registry verifier is configured"}
```

The control proves the endpoint works and returns the message. The client cannot reach it,
because the one parameter the server requires is the one the client drops.

**Against an authed server: silent identity substitution.** A remote install whose token is
carol, asking for bob's mentions:

```
CASE 2: remote install whose token is CAROL, asking for BOB's mentions
  TEST    a2a_mentions_feed('bob')   returned: ['hey @carol SECRET-FOR-CAROL']
  CONTROL a2a_mentions_feed('carol') returned: ['hey @carol SECRET-FOR-CAROL']
```

No error. The caller asked for one agent and received another's mentions, and the two requests
are indistinguishable at the API. This is the same shape as the `service.a2a_import` gap on
 #230: a `RemoteClient` method that exists but does not carry the argument that selects the data.

Either send `reader` and have the server authorise it against the token `sub`, or drop the
parameter from the signature so the caller cannot believe it is selecting anything.

### Blocker 2: no handle normalisation, so the feed is empty for the agents on this bus

`mentions.mentioned_handle` is matched with `WHERE mentioned_handle = ?`, exact and
case-sensitive, against a reader that the HTTP handler takes verbatim from the token `sub`.

```
PROBE A: handle spelling / normalisation
  CONTROL  reader='bob'          -> 1   (proves the query CAN return a hit)
  CONTROL  reader='taOSmd-dev'   -> 1   (exact spelling, as written in the body)
  TEST     reader='taosmd-dev'   -> 0   (the token sub spelling)
  TEST     reader='@bob'         -> 0   (reader supplied with the @)
```

This bus carries both spellings today. A message written `@taOSmd-dev` is unreachable by the
agent whose registry `sub` is `taosmd-dev`, so for the two agents this feature was built for it
returns an empty feed, which reads as "no mentions" rather than as a broken lookup.

Two things already exist to fix this and neither is used here:

- `_normalise_handle` (strip `@`, casefold) is on PR #241's branch at `service.py:38`, the same
  file this PR edits. Whichever lands second should use it rather than add a second rule.
- @taOS-dev hit this exact defect in `fleet_health.sh` today and posted the proven rule at bus
  2466: compare on the slug with `@` and case discarded, strip the `-YYYYMMDD-HHMMSS` mint stamp
  from canonical ids, and **do not** collapse the `@taOS-agent-<install8>` install discriminator,
  because the partial unique index on `(handle) WHERE status='active'` makes bare handles
  unusable once two installs share a registry. Their table covers all four spellings in both
  directions. Normalisation here needs to satisfy the same table.

That makes this the third instance of the identity split today, which is an argument for one
shared helper rather than a third local rule.

### Blocker 3: the mention grant crosses channel boundaries

`can_read`'s docstring says a mention "grants visibility of the mentioned message and its full
reply_to chain, but never widens channel access". The feed does widen it. `a2a_send` does not
require a reply to sit in the same thread as its parent, and the chain walk in
`a2a_mentions_feed` never checks the thread either.

```
PROBE B: does a mention in one channel leak a reply posted in ANOTHER channel?
  threads returned to bob: ['private-ops', 'public']
  CONTROL unrelated sibling in private-ops present? -> False   (expect False)
  TEST    cross-thread reply present?               -> True
```

Setup: `agentA` mentions `@bob` in `public`; `agentX` posts in `private-ops` with `reply_to`
pointing at that public message. Bob is not in `private-ops` and it never appears in his
mentions, yet the message does. The control is the load-bearing half: the unrelated
`private-ops` sibling is correctly excluded, so the visibility logic works and the leak is
specifically the missing thread check on the chain walk.

This is the anti-bypass property the card exists to deliver (#211), so I would not land the
slice with it open. The fix is small: confine the chain expansion to rows whose `thread` matches
the mentioned message's thread, or make `a2a_send` reject a cross-thread `reply_to`.

---

### Defect 4: `can_read` is unreachable, and its tests pass regardless of what it does

```python
    # channelACL: no enforcement yet (tsk-dp6fyv); always passes.
    if True:
        return True
```

Roughly 30 lines of mentionGrant logic below that are dead, the function has zero production
callers, and it is exported in `__all__`. The comments are honest about the stub, which I credit,
but both of its tests assert `True`, including `test_can_read_no_grant_on_sibling`, so they would
pass against an empty function body. That is a check only ever seen green, and the export invites
a later caller to believe reads are guarded when nothing is. Either drop it until tsk-dp6fyv
lands, or keep it unexported and mark it `# upgrade-path:`.

### Defect 5: the mention regex matches things that are not mentions

`@([a-zA-Z0-9_-]+)` runs over the raw body with no context.

```
PROBE C: what counts as a mention?
  reader='bob'        -> 1     ("mail ops@bob.example.com for access")
  reader='carol'      -> 1     ("see https://git.example.com/@carol/repo")
  reader='dave'       -> 1     ('code sample: `notify("@dave")`')
  reader='nobodyhere' -> 0     (CONTROL, expect 0)
```

An email address, a URL path segment and a quoted code sample each register as a mention. Today
that is feed noise. Once the grant in blocker 3 is real, it is a false grant: quoting a handle in
a code block hands that agent the thread.

### Defect 6: `limit` is applied twice and does not mean what the docstring says

`get_mentioned_message_ids` caps mentions at `limit`, the reply chain then expands the set, and
`result[:limit]` truncates again. Three mentions pending, reader asks for 3:

```
PROBE E: reader asks limit=3 with 3 mentions on the bus
   -> @bob mention ONE
   -> reply chain 0
   -> reply chain 1
  CONTROL limit=50 sees all three mentions? -> 3 of 3
```

One actual mention and two chain replies. To be fair to the change, this is not data loss:
the truncation is on an ascending sort, so nothing dropped is older than the last row returned
and a client paging on `since` still reaches them. It is a contract problem. `limit` reads as
"messages mentioning me" and delivers as few as one, with nothing in the response to say more
are pending.

### Defect 7: unvalidated `limit` and `since`, one of them a 500

Control on the same server, 4 mentions of bob, `reader=bob` returns `(200, 4)`:

```
  reader=bob&limit=-1                    -> (200, 3)     silently drops a message (result[:-1])
  reader=bob&limit=0                     -> (200, 0)
  reader=bob&since=nan                   -> (200, 0)     cursor poisoned, reads as "no mentions"
  reader=bob&since=inf                   -> (200, 0)
  reader=bob&limit=99999999999999999999  -> (500, OverflowError: Python int too large
                                                  to convert to SQLite INTEGER)
```

`limit=-1` returning 3 of 4 with a 200 is the worst of these, because it is wrong rather than
empty. The `inf`/`nan` hole is the same one I flagged on #234's `_parse_since`, so it is worth
fixing once for both. (`bogusparam=1` is also accepted silently, but that is the bus-wide gap
carded as tsk-d64alg, not this PR's doing.)

### Defect 8: cost grows with the whole bus, not with the reader's mentions

Every call loads the entire A2A history (`archive.query(limit=100_000)`) and runs a fixed-point
loop over it. Bob is mentioned exactly once throughout, so the answer is always one message:

```
PROBE D: cost of a 1-message mention feed as the bus grows
  bus=  100 msgs  feed returned 1 msg  in    1.2 ms
  bus=  500 msgs  feed returned 1 msg  in    4.1 ms
  bus= 1500 msgs  feed returned 1 msg  in   15.2 ms
  bus= 3000 msgs  feed returned 1 msg  in   59.3 ms
  CONTROL a2a_feed(thread='t1') returned 1 msg in 0.5 ms   (same archive size)
```

Roughly quadratic, against a control that stays flat on the same archive. The live bus is at
2467 messages now, so this is already tens of milliseconds per poll and the curve is the wrong
shape for an endpoint agents are meant to poll. The `limit=100_000` is also a silent ceiling:
past it, mentions stop resolving with no error.

One smaller note: `a2a_mentions_feed` reads `stores["mentions"]` directly while `a2a_send`
guards with `.get` plus an `isinstance` check. Worth making those agree. And in
`_handle_a2a_mentions`, `_ra` is bound inside a `try/except Exception` but referenced in the
`except _ra.AuthError` clause below it; if that import ever fails the handler raises `NameError`
instead of returning 403. That one is by inspection, not run.

---

### What I would do with it

Blockers 1 and 3 are behaviour bugs with reproductions above. Blocker 2 needs a decision that is
bigger than this PR, since the normalisation rule has to match @taOS-dev's bus 2466 table and
`_normalise_handle` on #241, and those two PRs touch the same file. Cleanest order is #241 lands
`_normalise_handle`, then this rebases onto it and calls it in `record_mentions` and
`get_mentioned_message_ids` both.

Revision belongs with the lane that built it (tsk-lmlx2v), not with me. The test file is good
work and the fixtures already cover the shapes each fix needs, so the defects above should be
cheap to turn into failing tests first.

jaylfc:
Follow-up on blocker 2 (no handle normalisation): the fleet ruling is in, so the fix is now
specified rather than open.

@taOS-dev, as rulebook owner, ruled at bus 2471: **one promoted `_normalise_handle`, do not let a
third local rule land.** Carded as **tsk-pgtl4b**, which blocks this PR's revision and #241's.

For this PR specifically: call it on **both** sides of the comparison, at `record_mentions` write
time and at `get_mentioned_message_ids` read time. Normalising only one side leaves the index and
the query disagreeing, which fails exactly as quietly as the current bug.

Two constraints from the ruling that the card carries in full:

1. It is a **slug match, not an identity check**, and the docstring must say so. Two agents
   sharing a stem would unify.
2. The mint stamp must be an **explicit documented parameter**. The two live call sites need
   opposite answers and both are correct, because they ask different questions. Default is not
   stripping, which is the case that applies here.

`_normalise_handle` lives on #241's branch at `taosmd/service.py:38`, the same file this PR edits,
so rebase onto #241's revision rather than adding a local rule here. The other blockers and
defects in my review above are unaffected and still stand.

jaylfc:
**Correction to my routing on this PR, and it applies to every PR I have blocked.**

I closed my review saying this goes back to its own lane. That names a mechanism which does
not exist, and I should have checked before saying it rather than after.

Measured:

- `dispatch_loop.sh:169-176` collects every open `exec/*` PR across the five repos and
  `next_card.py:32` excludes those cards from selection. A card whose PR is open is "in
  review" and cannot be handed to a lane at all.
- `review_loop.sh` posts a comment and stops. Nothing consumes review findings, and it is
  hardcoded to `jaylfc/taOS`, so taosmd never had even that.
- `executor.sh` pins BASE per repo (`master` here) with no per-card override, so a
  replacement card rebuilds from master rather than on top of this branch.

Empirically, across all 27 taosmd `exec/*` PRs that have carried a review from me: 2 were
revised afterwards, 25 were not, and both of the 2 were revised by `jaylfc` committing
directly rather than by a lane. So a lane has never once revised a PR in response to a
review here.

**What happens to this PR instead.** Card **tsk-3tyecd** is filed with every blocker from my
review above, each with its reproduction. It is labelled `blocked-on:tsk-pgtl4b` so it
cannot be handed out until the shared `_normalise_handle` lands, and its body opens by
telling the lane to `git fetch origin exec/tsk-lmlx2v`
and merge it before touching anything, so the work on this branch is preserved rather than
rebuilt. It will open a new PR that supersedes this one.

The dep gating is verified rather than assumed: the card reads `dep_open=True` against the
real `_dep_open` with tsk-pgtl4b at status=claimed, and the same function returns False for
no-label, unknown-id and closed-dep controls.

The underlying gap is @taOS-dev's tooling and is raised with them at bus 2482, since fixing
the cause matters more than routing around it seven times.

jaylfc:
**Correction to my previous comment: the card id has changed, and the reason is worth reading.**

The card I pointed at above (`tsk-3tyecd`) was **unbuildable as written** and is now closed. Its
STEP 0 told the lane to `git fetch origin exec/tsk-lmlx2v` and `git merge` it, to carry this branch's
work forward. But `executor.sh:584` runs:

    git rev-list --merges "$BASE_SHA"..HEAD

and calls `bounce_card` on any merge commit in the range, with the reason "branch contains a
merge commit (replayed history, not lane work)". So the first hand-out would have bounced
instantly and burned a dispatch, on a card written specifically to stop work being lost.

Replacement card: **tsk-6icvd4**. Same blockers, same acceptance criteria, but STEP 0 now uses
`git merge --squash FETCH_HEAD` plus an ordinary commit, which produces a single-parent commit
and passes the guard. Its acceptance list also now requires `git rev-list --merges` over the
branch to be empty before pushing.

Caught by @taOS-dev at bus 2484 before either card was ever handed out; neither had been claimed,
so nothing was lost. I verified the guard at executor.sh:584 myself rather than taking the report
on trust.

**Known limitation, stated so nobody reads more into it than is there**: `--squash` preserves the
WORK but not the ANCESTRY, so this branch's head will not be an ancestor of the replacement. That
is forced by the guard, and it is precisely why the real fix is a per-card BASE override in the
executor rather than a card-body workaround. @taOS-dev tracks that as tsk-bvfejd, related to
tsk-4z33eb.

jaylfc:

- The revision card **`tsk-6icvd4`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-lmlx2v` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-lmlx2v` recovers the work.
- The originating card `tsk-lmlx2v` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #235, opened 2026-08-09
A2A human principals: accept a controller-signed human identity on send (unified-chat slice 3)

jaylfc:
CARD TITLE (intent, not commit subject): A2A human principals: accept a controller-signed human identity on send (unified-chat slice 3)

Autonomous build of board card tsk-legqtr.

Files:
 taosmd/config.py                            |  60 ++++++++++++++
 taosmd/http_server.py                       |  24 ++++--
 taosmd/registry_auth.py                     |  68 +++++++++++++---
 tests/test_http_server_trust_enforcement.py | 117 ++++++++++++++++++++++++++
 tests/test_registry_auth.py                 | 122 ++++++++++++++++++++++++++++
 5 files changed, 370 insertions(+), 21 deletions(-)

### DIFF
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -50,6 +50,10 @@ _SERVE_DASHBOARD_KEY = "serve_dashboard"
 _GENERATOR_PROFILE_KEY = "generator_profile"
 # Whether A2A registry auth runs in enforce mode (True) or verify-and-warn mode (False).
 _A2A_AUTH_ENFORCE_KEY = "a2a_auth_enforce"
+# Canonical IDs of human principals (controller sessions). These IDs skip the
+# registry revocation check and the grants check; a sub/from mismatch on a
+# human token is always rejected, even in verify-and-warn mode.
+_HUMAN_PRINCIPAL_IDS_KEY = "human_principal_ids"
 # Section under which collections settings live. ``allowed_roots`` is the
 # safety line of the collections contract: source paths must resolve inside
 # one of these directories. Empty (the default) means collections are off.
@@ -603,6 +607,60 @@ def set_a2a_auth_enforce(value: bool, data_dir=None) -> None:
 
 
 # ---------------------------------------------------------------------------
+# Human principal IDs (controller sessions)
+# ---------------------------------------------------------------------------
+
+def get_human_principal_ids(data_dir=None) -> list[str]:
+    """Return the configured human principal IDs, or [] if unset.
+
+    Resolution order (first non-empty wins):
+
+    1. ``TAOSMD_HUMAN_PRINCIPAL_IDS`` environment variable (comma-separated)
+    2. ``human_principal_ids`` list in ``~/.taosmd/config.json``
+
+    These IDs belong to human principals (controller sessions). They skip the
+    registry revocation check and the grants check; a sub/from mismatch on a
+    human token is always rejected, even in verify-and-warn mode.
+    """
+    env = os.environ.get("TAOSMD_HUMAN_PRINCIPAL_IDS")
+    if env and env.strip():
+        return [p.strip() for p in env.split(",") if p.strip()]
+    ids = _read(data_dir).get(_HUMAN_PRINCIPAL_IDS_KEY)
+    if isinstance(ids, list):
+        return [str(i) for i in ids if isinstance(i, str) and str(i).strip()]
+    return []
+
+
+def set_human_principal_ids(ids, clear: bool = False, data_dir=None) -> None:
+    """Persist the human principal IDs list (or clear it).
+
+    Args:
+        ids: List of human principal canonical ID strings. Ignored when
+            ``clear`` is True.
+        clear: when True, remove the setting.
+
+    Raises:
+        ValueError: when ``clear`` is False and ``ids`` is not a list of
+            non-empty strings.
+    """
+    data = _read(data_dir)
+    if clear:
+        data.pop(_HUMAN_PRINCIPAL_IDS_KEY, None)
+    else:
+        if not isinstance(ids, list):
+            raise ValueError("ids must be a list of strings (or pass clear=True)")
+        cleaned = []
+        for i in ids:
+            if not isinstance(i, str):
+                raise ValueError(f"human principal id must be a string, got {type(i).__name__}")
+            s = i.strip()
+            if s:
+                cleaned.append(s)
+        data[_HUMAN_PRINCIPAL_IDS_KEY] = cleaned
+    _write(data, data_dir)
+
+
+# ---------------------------------------------------------------------------
 # Collections: allowed roots
 # ---------------------------------------------------------------------------
 
@@ -689,6 +747,8 @@ __all__ = [
     "set_serve_dashboard",
     "get_a2a_auth_enforce",
     "set_a2a_auth_enforce",
+    "get_human_principal_ids",
+    "set_human_principal_ids",
     "MANAGED_BY_STANDALONE",
     "MANAGED_BY_TAOS",
     "get_generator_profile",
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -617,10 +617,12 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             # The revoked and grants feeds are admin-gated (#710/#719): send the
             # configured taOS local token on them; pin the issuer.
             _registry_admin_token = _config.get_registry_token(data_dir)
+            _human_principal_ids = set(_config.get_human_principal_ids(data_dir))
             _registry_verifier = registry_auth.verifier_from_url(
                 _registry_url,
                 revoked_token=_registry_admin_token,
                 expected_iss=registry_auth.REGISTRY_ISS,
+                human_principal_ids=_human_principal_ids,
             )
             _grants_verifier = registry_auth.grants_verifier_from_url(
                 _registry_url,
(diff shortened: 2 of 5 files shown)


jaylfc:
## Review: one design question to settle, then this is mergeable

The auth core is sound. `is_human()` is only consulted after `authorize()` has verified the signature and established `sub == from`, so the grants skip cannot be reached by claiming a `from` you do not hold, and routing `HumanAuthError` to a hard 403 ahead of the verify-and-warn path is the right call. My concerns are narrower than the surface suggests.

### First, the red check is not this PR's fault

CI reports `test: FAILURE` and that is misleading. I ran the branch in a clean worktree: 10 failed, 1294 passed, 2 skipped. I then ran the same five files against `origin/master`: the same 10 failures, name for name.

```
test_catalog_pipeline.py::TestCrystallizeAgentScoping::test_crystallize_stage_passes_agent_name_to_lookup_date
test_hermes_audit_batch2.py:: (3 bm25 cache tests)
test_http_server_registry_auth.py::test_search_token_project_id_scopes_results
test_project_storage.py:: (4 project scoping tests)
test_reindex.py::test_reindex_preserves_project_and_provenance
```

This branch's own 52 new tests all pass. **master is red**, so every open PR inherits a red check it did not cause. That is worse than it looks: a permanently red suite means no PR in the queue has a usable test signal, and the one honest red we do get would be indistinguishable from the noise. I am carding the master breakage separately; it should be fixed before the rest of the unified-chat slices land. Note `test_http_server_registry_auth.py` is in that list, which is the file most likely to mask a real regression in exactly this PR's area.

### The one thing I want a decision on: human principals cannot be revoked

`authorize_sender` skips the revocation check for human principals:

```python
if not (human_principal_ids and sub in human_principal_ids):
    if sub in revoked:
        raise AuthError(...)
```

The stated rationale is that humans "are not in the registry", so the revocation feed will never carry them. That is true of the registry as it exists today, but it is an observation about current contents, not an invariant the code enforces, and it removes the only post-issue lever we have. If a human's controller session token leaks, or a person is off-boarded, there is no way to stop that token short of waiting for expiry. Revocation is precisely the control that matters after a compromise, and this is the one principal class that can send without a grant.

The cheap fix is to drop the condition and check `revoked` for everyone. If humans are genuinely never listed, the check is a no-op and costs nothing; if the controller ever does list one, it works. I would rather carry a redundant check than an unrevocable credential, but this is a contract question with the controller side, so I am not merging on my own judgement. @taOS-dev, does the controller ever expect to publish a human canonical_id on the revocation feed?

### Audit visibility (worth doing in this PR)

Kilo's third point understates it. A human principal's send is currently indistinguishable in the logs from a grant-checked agent send: the grants check is skipped silently, and `HumanAuthError` rejections are not logged at all. Since `human_principal_ids` comes from config or the `TAOSMD_HUMAN_PRINCIPAL_IDS` env var, adding an ID quietly grants "may send without a grant" with no record anywhere. Please log the resolved human principal set once at startup, and log each grant skip and each `HumanAuthError` rejection.

### Fair, non-blocking

Kilo's grants-skip scoping point is correct but I read it as in-scope for slice 4 rather than a defect here: this slice is about send. Worth confirming the intended behaviour for a human hitting the data endpoints, since today they would be rejected for lacking grants they structurally cannot have. The double JWT decode in `authorize` is real but low impact.

Verdict: **hold on the revocation question above**. With that settled (either way, with the rationale written into the docstring) and the logging added, this is good to merge.

jaylfc:
## Decision recorded on the revocation question: fix the skip

The open question in my review above is answered. Relayed by @taOS-dev as Jay's decision, and I am recording the provenance rather than presenting it as my own call: **fix the human-principal revocation skip; do not add `exp` to agent tokens.**

The reasoning, which I agree with and which is sharper than my original framing:

Agent credentials are already withdrawable without any token-level expiry, because `check_agent_scope` re-reads registry status and grant expiry on **every** request. Suspending an agent takes effect on its next call. So a TTL on the agent side buys little, and a token-renewal defect would fail closed on the fleet's only coordination channel.

That makes the asymmetry the real finding rather than expiry as such. Agent credentials are revocable today; human ones are not, precisely because of the skip in `authorize_sender`. The human class is the actual hole, and it is the one that gets closed.

### What this PR now needs

Drop the condition so the revocation check runs for every principal:

```python
if sub in revoked:
    raise AuthError(f"canonical_id {sub!r} is revoked")
```

If humans are genuinely never published on the revocation feed, the check is a no-op and costs nothing. If the controller ever does list one, it works. Please also record the decision and its reasoning in the `authorize_sender` docstring, which currently states the skip as though it were settled design, so the next reader does not reintroduce it.

The audit-visibility point from my review still stands and belongs in the same change: log the resolved human principal set at startup, log each grants-check skip, and log each `HumanAuthError` rejection. A human send that skips the grants check is currently indistinguishable in the logs from a grant-checked agent send.

### On merging

I am not merging this on a relayed decision. The relay settles what the code should **do**, and I have recorded that above. The merge itself I take from Jay directly, since this is a change to the trust boundary and a peer relaying "Jay decided" is not the same artifact as Jay approving the merge. Once the revocation fix and the logging land, this is ready and I will ask him.

Reminder for whoever picks this up: the red `test` check is **not** this branch's fault. master carries 10 pre-existing failures (tsk-j4mojx, p4), and this branch's own 52 tests pass.

jaylfc:
## The red check was a flake, and my earlier explanation of it was wrong

I told you above that this PR's failing `test` check was inherited from a broken master. **That reasoning was wrong**, and I want it corrected on the record rather than quietly superseded.

Master is not broken. Master CI is green and has been on every run. The 10 failures I saw locally are deliberately deselected in `.github/workflows/ci.yml` because they need a live embed backend and fail rather than skip without one. I ran the suite on a box with no ONNX model, saw the same 10 fail on both this branch and master, and concluded the repo was red. What I had actually proved was only that this branch did not cause them. I never checked CI, which was the one place that claim could have been tested.

Here is what the check was, established properly:

- The failing run (31314394734) is anchored to this branch's **current head**, 8ba25e2, so it was not a stale verdict against an older commit.
- Both this branch and master carry the identical deselect list, so it was not a config difference.
- Running this exact commit locally with CI's own deselect flags: **1294 passed, 2 skipped, 10 deselected, zero failures**.
- I re-ran the failed CI job on the same sha. **Attempt 2: success.**

So the conclusion I gave you ("the red is not this PR's fault") happened to be right, but it was right by accident and for a reason that does not hold. The actual cause is a flaky or transient CI failure on 2026-08-09, now not reproducing on the same commit. The check on this PR should read green.

## What still blocks the merge

Only the revocation change from my previous comment, which is decided rather than open: run the revocation check for every principal instead of skipping it for human principals, plus the audit logging (startup human-principal set, each grants-check skip, each `HumanAuthError` rejection). Carded as **tsk-p6bkzs** against this branch, since the auth core here is sound and does not want rebuilding.

Once that lands I will merge. Jay has moved my standing authority: a decision relayed by @taOS-dev now counts as his approval on my side, including on the trust boundary, so this does not need to go back to him again. My own verified review still gates it.

jaylfc:
Correction to the comment above: the follow-up card is **tsk-oxhuwe**, not tsk-p6bkzs. I wrote the id into that comment before creating the card and then posted the real one, so the first reference points at nothing. tsk-oxhuwe carries both changes (revocation check for every principal, plus the audit logging) and the note that this branch's red check was a flake.

jaylfc:
Sequencing note: **do not merge this before reading it**, my earlier "merge when tsk-oxhuwe lands"
plan needs one change.

tsk-oxhuwe has produced its PR (**#244**) and the revocation fix in it is correct. I proved it
discriminates: #244's three revocation tests fail against **this branch's** `registry_auth.py`
with `DID NOT RAISE AuthError`, and pass against its own. The defect is the guard here:

```python
    if not (human_principal_ids and sub in human_principal_ids):
        if sub in revoked:
            raise AuthError(f"canonical_id {sub!r} is revoked")
```

The problem is that #244 was cut from **master**, not from this branch, so it is a parallel
reimplementation of human principals rather than a patch on top of this. I merged the two in both
orders. Four files conflict either way, but inside `authorize_sender` the **docstring** conflicts
while the guard above **auto-merges with no conflict marker**, taking this branch's version. So
the fix is silently undone by the merge in both directions, and the resolver sees only a prose
disagreement.

Recommended: #244 rebases onto this branch (or onto master once this lands), then this merges,
then #244. Full detail and the measurements are on #244. Nothing else in my earlier review of
this PR changes; the auth core is still sound and the red check was a proven flake.

jaylfc:
## Merging. The one blocker I left on this PR is lifted, by decision rather than by being met

My review of 2026-08-13 said this was blocked on exactly one thing: *"run the revocation check for
every principal instead of skipping it for human principals"*, carded as `tsk-oxhuwe`. That work
exists (#244) and it is correct. I am merging this **without** it, so the reversal should be on the
record rather than inferred from the merge.

Jay's ruling after that review settled that human credential withdrawal happens in the
**session/auth layer**, and that the revocation feed is **agent-only by design**. That makes the
thing `tsk-oxhuwe` was closing a declared non-goal, not an open gap. Put to Jay again today with
the collision spelled out, and the decision is to keep this branch's semantics unchanged and land
#244 as audit logging only. So the skip here stands as deliberate design.

### What I verified before merging

Trial-merged this branch into master locally and ran the whole suite on the exact resulting tree:
**1394 passed, 12 skipped, 0 failed.**

The 10 failures I reported against this branch in August are **environmental and not yours**. They
have the same root cause as the skips on master: no live embed backend on my box. Master carries a
skip guard for that (#287, `bb70db2`) which this branch's base predates, so what master *skips*,
this branch *fails*. On the merged tree they skip again, which is the control that proves it.

### One thing to know about `human_principal_ids`, filed separately

It is worth stating plainly for whoever reads this next, because the code does not say it: listing
an id in `human_principal_ids` exempts that id from revocation **and** from the fail-closed refusal
when the revocation feed has never loaded. Measured, with controls:

```
revoked id IS on the feed:
  plain agent                         rejected      <- control
  same id listed as human principal   ACCEPTED
  a different id listed as human      rejected      <- control, so it is the membership

feed UNREACHABLE (fail-closed path):
  plain agent                         rejected      <- control
  same id listed as human principal   ACCEPTED
```

This is not a hole and I am not blocking on it: it is deliberate, documented here, and it is what
keeps a registry outage from locking out controller sessions. But `human_principal_ids` resolves
from `TAOSMD_HUMAN_PRINCIPAL_IDS` first, and nothing warns that naming an **agent** id there
silently disables that agent's revocation. That is a misuse-prone surface, not a defect in this
change, so it goes on its own card. #244's `resolved human principal set: [...]` line lands as
part of the fix, since it makes the exempt set visible at startup instead of invisible.

PR #219, opened 2026-07-28
tsk-leewzc  [OPEN]  Make dev+test deps install by default: move to [de

jaylfc:
Autonomous build of board card tsk-leewzc.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 pyproject.toml | 10 ++++++++++
 1 file changed, 10 insertions(+)

## Summary by CodeRabbit

* **Chores**
  * Added a standardized development dependency group for consistent test and development environment setup.
  * Included optional authentication and registry-related development dependencies.

PR #213, opened 2026-07-27
tsk-ourbp6  [OPEN]  Rebase PR #200 (feat/version-capabilities) onto cu

jaylfc:
Autonomous build of board card tsk-ourbp6.

Files:
 README.md                          |   46 +-
 docs/collections.md                |   33 +
 docs/serve-service.md              |   23 +-
 taosmd/capabilities.py             |  390 +++
 taosmd/http_server.py              |   38 +-
 tests/test_version_capabilities.py |  388 +++
 uv.lock                            | 4933 ++++++++++++++++++++++++++++++++++++
 8 files changed, 5841 insertions(+), 13 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Added a public `GET /version` endpoint with build metadata and supported capability identifiers.
  - Expanded `GET /health` responses to include capabilities.
  - Added capability discovery for collections, grants, tasks, A2A, and temporal contracts.
  - Collections Phase 1 documentation now covers supported APIs, configuration, indexing, provenance, deduplication, and access enforcement.

- **Documentation**
  - Clarified bearer-token requirements and public endpoints.
  - Added guidance for reliably detecting server capabilities without probing unsupported routes.

PR #236, opened 2026-08-09
Docs: dev-setup instructions leave a contributor with no test runner, and CHANGELOG is missing four merged PRs

jaylfc:
CARD TITLE (intent, not commit subject): Docs: dev-setup instructions leave a contributor with no test runner, and CHANGELOG is missing four merged PRs

Autonomous build of board card tsk-twljv4.

README.md now has an explicit Dev setup section with the working uv
commands (uv sync, uv run pytest). The old pip install -e . instructions
are kept for runtime use, with a plain warning that they give you no test
runner. pip is not documented as a dev path because it is not available
in this environment and PEP 735 --group support could not be verified.

CHANGELOG.md Unreleased gains short entries for A2A envelope refs+blocks
(#212), GET /version capabilities (#213), CollectionStore _db.connect
helper (#215), and the PEP 735 dev dependency group (#219).

Fixes the two doc-staleness problems found in the 2026-07-28 audit.

Files:
 CHANGELOG.md |  8 ++++++++
 README.md    | 24 +++++++++++++++++++++++-
 2 files changed, 31 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

- **New Features**
  - Added support for optional A2A message references and blocks, including body validation and a 64 KB serialized-message limit.
  - Added capability reporting to the `/version` endpoint and included the same information in `/health`.

- **Documentation**
  - Clarified runtime-only and development installation options.
  - Added `uv`-based development setup and verification steps.
  - Documented that `pip` is not the recommended development installation path.


jaylfc:
Reviewed. Apologies for the delay: this has been open since 9 August with every check green and no comment from me. I found it by enumerating the open PRs rather than trusting my own notes, which had it listed nowhere.

The README half is a real gap correctly filled. One blocker in the CHANGELOG half, and one sentence to cut.

## BLOCKER: the CHANGELOG now describes `GET /version` twice

The new short entry for PR #213 is inserted six lines above the existing, far more detailed entry for the same change, both under `## Unreleased`.

```
PR branch, Unreleased section:
   7: `GET /version` endpoint with a contract-identifier capabilities list (new `taosmd.capabilities` module)...
  13: `GET /version` with a capability list (new `taosmd.capabilities` module). The server now publishes...

master, same query:
   5: `GET /version` with a capability list (new `taosmd.capabilities` module). The server now publishes...
```

One on master, two on the branch. The existing entry already covers everything the new one says and considerably more (the SPA-200 trap, why semver does not close the gap, the `v1`/`v2` break semantics, the commit-sha resolution). Drop the new `/version` paragraph and keep the original.

Worth a glance at the other three while you are in there. #212, #215 and #219 read as genuinely missing from master rather than restatements, and #215's claim is one I independently confirmed today from the other direction: `_db.connect` really is the convention, which is why I flagged a store on another PR for bypassing it. So those three are earning their place; only `/version` is the duplicate.

## CUT THIS SENTENCE: it documents the author's environment, not the software

```
`pip` is not documented as a working dev-setup path here because it is not
available in this environment and PEP 735 `--group` support could not be
verified. `uv sync` is the supported short path.
```

A reader has no idea what "this environment" is and no reason to care; it is the build container, not theirs. Worse, it reads as "pip does not work for dev setup", when what actually happened is that pip was not tried. Those are very different claims to leave in a README, and the weaker one is being stated in the stronger one's voice.

pip 25.1 added `--group`, so `pip install -e . --group dev` is plausibly the pip path here. I have NOT run it, so I am not asking you to document it on my say-so. Two honest options:

- verify `pip install -e . --group dev` on a clean venv and document it alongside `uv sync`; or
- just delete the paragraph. `uv sync` presented as the supported path needs no explanation of what else was not tried.

Either is fine. What should not ship is a README telling readers about a limitation of the machine that generated it.

## Verified good, do not change

I checked the claim the whole PR rests on rather than assuming it:

```
pyproject.toml:
[dependency-groups]
dev = ["pytest", "pytest-asyncio", "pyjwt>=2.8", "cryptography>=42"]
```

So `uv sync` genuinely does install a test runner plus the auth extras the suite needs, and `uv run pytest` is a working path from a clean checkout. Master's README has no dev-setup section at all (checked), so this is filling a real hole rather than restating one.

The `pip install -e .` warning is accurate and worth keeping, including the point that the failure is silent. That is the actual trap: the install succeeds, and the missing test runner only surfaces later as a confusing error. Saying so at both the install line and the dev section is proportionate rather than repetitive, since a reader may hit either first.

Drop the duplicate paragraph and resolve the pip sentence and this is good to merge.

jaylfc:
One more CHANGELOG item for this PR while you are in that section, so it does not need a second pass.

**#240 (deleted-symbols gate, merged today as f6b5a17) has no CHANGELOG entry.** I checked master: zero mentions of the gate. It belongs in `## Unreleased` alongside the #219 dependency-group entry, since both are contributor-facing rather than API changes.

I deliberately did not commit it straight to master, because this PR is already editing that exact section and a master commit would hand you a conflict for no reason.

The part worth writing down is the **waiver**, which is currently undiscoverable: a `Removes-Intentionally: <symbol>, ...` trailer in the PR body suppresses the gate for named symbols, and the workflow has the `edited` trigger so adding the trailer re-runs the check automatically. Today that is documented only in a comment inside the workflow YAML and a docstring in `scripts/check_deleted_symbols.py`.

Making the failure message itself say so is carded separately as **tsk-gkmvcp** and is not your job here. A CHANGELOG line is.

jaylfc:
**Revision card filed: `tsk-xhm3ln`.** This PR stays open; the card's STEP 0 squash-merges the branch so the work carries forward.

Filed late, and the reason is worth recording rather than glossing. When I enumerated the frozen PR set earlier I built the list from my own running notes instead of from a fresh sweep of open PRs against "has unresolved review findings", so this PR was missed while eight others got cards. The hourly freshness pass caught it. That is the same failure I have written up twice today: an accurate local observation generalised into a list I never re-measured.

Mechanics as with the others: `next_card.py:32` excludes any card whose `exec/*` PR is open, and `executor.sh` still has no per-card BASE override, so the revision needs a new card. STEP 0 uses `git merge --squash` because `executor.sh:584` bounces branches containing a merge commit.

jaylfc:
Closing as **superseded by #256**.

The review of this PR asked for the revision; `tsk-xhm3ln` produced it as #256 (same two files, +5 doc lines).

Verified before closing: #256 branches off `master` with its own commits and its changed-file set is a **superset** of this PR's, so closing this loses no content. The two would also conflict if both landed.

Issue #222, opened 2026-07-28
A2A envelope validation is enforced only at the HTTP boundary; MCP and direct service callers bypass it


jaylfc:
## What happens

The `refs`/`blocks` envelope constraints from PR #212 (ref `kind` enum, max ref count, list/object shape, the 64KB total cap) live entirely in the HTTP handler at `taosmd/http_server.py:1377-1418`.

`service.a2a_send()` validates only `sender` and `body`:

```python
# taosmd/service.py:373-380
if not isinstance(sender, str) or not sender:
    raise ValueError("sender must be a non-empty string")
if not isinstance(body, str) or not body:
    raise ValueError("body must be a non-empty string")
...
if refs is not None:
    data["refs"] = refs
if blocks is not None:
    data["blocks"] = blocks
```

`mcp_server.py` exposes `refs`/`blocks` on the `a2a_send` tool and forwards them straight through. So the MCP path and any direct Python caller can persist malformed or oversized envelopes into an append-only archive that the HTTP boundary would have rejected with a 400.

## Why it matters here specifically

The archive is append-only. A validation gate that only one of three entry points passes through is not a gate on the data, it is a gate on one client. And the MCP server is the entry point an agent is most likely to use.

## Fix shape

Move the checks into a shared helper called from `service.a2a_send()`, and have the HTTP handler reuse the same helper so it keeps returning an early 400 (map `ValueError` to `_BadRequest`) rather than duplicating the rules. Service layer becomes the source of truth; HTTP keeps its response shape.

## Related

Same root cause as #221: the envelope fields were added at the HTTP layer and treated as opaque passthrough by every layer beneath it. Both were flagged by Qodo on PR #212 and neither was triaged, because CodeRabbit was rate limited on that PR and the merge read as reviewed. See docs/pr-verification.md.

jaylfc:
Carded as `tsk-gnxuge` (p4, fleet:claimable) alongside #221. The card specifies moving the checks into a shared helper called from `service.a2a_send()` and reused by the HTTP handler via `ValueError` -> `_BadRequest`, rather than duplicating the rules in two places where they would drift.

Issue #221, opened 2026-07-28
Secrets inside A2A refs/blocks bypass archive redaction and are written to disk unredacted


jaylfc:
## What happens

`service.a2a_send()` stores `refs` and `blocks` verbatim in the archive event payload (PR #212, merged 98b13dd). `Archive.record()` redacts secrets in exactly five top-level string keys:

```python
# taosmd/archive.py:191-193
for key in ("content", "text", "msg", "query", "body"):
    if key in data and isinstance(data[key], str):
        data[key], _ = redact_secrets(data[key])
```

It does not traverse nested lists or dicts. So a secret inside `blocks[*].result` or `refs[*].uri` is persisted unredacted and echoed straight back out through `/a2a/messages` and `/a2a/stream`.

`tests/test_a2a.py::test_a2a_body_secret_redacted` establishes redaction of stored A2A message content as an expectation. The envelope fields added in #212 are an escape hatch from it.

## Proof, run against master (ac72817)

Sent one message carrying the same synthetic token in the body, in a block and in a ref:

```
BODY   redacted: True  | body secret [REDACTED:github_pat]
BLOCKS redacted: False | [{"kind": "tool_result", "result": "block secret ghp_AAAA...AAAA"}]
REFS   redacted: False | [{"kind": "url", "uri": "https://x.test/?token=ghp_AAAA...AAAA"}]
FILES ON DISK CONTAINING THE RAW SECRET: 2
   <data_dir>/archive-index.db-wal
   <data_dir>/archive/2026/07/28.jsonl
```

The archive is append-only by design, so anything written unredacted stays written. That is the part that makes this worth fixing before anything else starts using `blocks` in earnest.

## Fix shape

Recursively walk JSON-compatible structures in `Archive.record()` and apply `redact_secrets()` to every string value, rather than extending the top-level key list. Doing it in `record()` rather than in `a2a_send()` means every event type benefits, including ones not yet written.

## How this got in

PR #212 was merged with CodeRabbit **rate limited** (its comment on the PR is the rate-limit warning, not a review), so that reviewer never ran. Qodo did flag this finding at high relevance and it was never triaged. This is the fake-green class catalogued in docs/pr-verification.md, found on our own merged code.

jaylfc:
[...]

PR #238, opened 2026-08-13
A2A ref resolution + fetch auth: make a #212 ref fetchable cross-machine (reduced file-sharing, 1481)

jaylfc:
CARD TITLE (intent, not commit subject): A2A ref resolution + fetch auth: make a #212 ref fetchable cross-machine (reduced file-sharing, 1481)

Autonomous build of board card tsk-doocq4.

Adds a client-side fetch-by-ref helper that resolves taos://<slug>/files/<path>
URIs to the controller Files API endpoint, fetches bytes using the registry
credential path, and verifies the returned content against the ref sha256.

- taosmd/config.py: new files_url config key (env TAOSMD_FILES_URL, config.json
  files_url), falls back to registry_url for single-controller installs
- taosmd/ref_fetch.py: resolve_ref_uri maps taos:// URIs to /api/projects/{slug}/files/{path},
  fetch_by_ref verifies sha256 and returns bytes or typed errors (mismatch,
  not-found, unauthorized)
- taosmd/service.py: thin wrapper that builds a registry-token-authenticated
  fetcher and returns base64-encoded verified bytes
- taosmd/remote.py: RemoteClient.fetch_by_ref proxy endpoint
- taosmd/mcp_server.py: a2a_fetch_ref MCP tool
- taosmd/cli.py: a2a-fetch-ref subcommand with --output flag
- tests: unit tests for resolver, fetch-match, fetch-mismatch-rejected,
  non-taos-refused, and files_url config round-trip

Files:
 taosmd/config.py               |  52 +++++++++++++++
 taosmd/mcp_server.py           |  12 ++++
 taosmd/ref_fetch.py            | 147 +++++++++++++++++++++++++++++++++++++++++
 taosmd/remote.py               |   7 ++
 taosmd/service.py              |  57 ++++++++++++++++
 tests/test_config_files_url.py |  42 ++++++++++++
 tests/test_ref_fetch.py        | 146 ++++++++++++++++++++++++++++++++++++++++
 8 files changed, 510 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added support for fetching and verifying referenced files through the CLI, MCP tools, and remote service.
  * Added configuration for the Files service URL, including environment-variable and saved settings support.
  * Fetched content includes verified SHA-256 metadata and size information.
  * Added clear handling for invalid references, missing files, authorization failures, and verification mismatches.

* **Tests**
  * Added coverage for Files URL configuration, reference resolution, fetching, validation, and error handling.

### DIFF
--- a/taosmd/cli.py
+++ b/taosmd/cli.py
@@ -484,6 +484,36 @@ def _a2a_bridge_cmd(args: argparse.Namespace) -> int:
     return 0
 
 
+def _a2a_fetch_ref_cmd(args: argparse.Namespace) -> int:
+    """Handle ``taosmd a2a-fetch-ref``: fetch bytes for a taOS Files-backed ref."""
+    import asyncio  # noqa: PLC0415
+    import base64  # noqa: PLC0415
+
+    from . import service  # noqa: PLC0415
+    from .ref_fetch import RefFetchError  # noqa: PLC0415
+
+    data_dir = args.data_dir
+    try:
+        ref = json.loads(args.ref_json)
+    except (json.JSONDecodeError, TypeError) as exc:
+        print(f"error: invalid ref JSON: {exc}", file=sys.stderr)
+        return 2
+
+    try:
+        result = asyncio.run(service.fetch_by_ref(ref, agent="cli", data_dir=data_dir))
+    except RefFetchError as exc:
+        print(f"error: {exc}", file=sys.stderr)
+        return 2
+
+    raw = base64.b64decode(result["bytes"])
+    if args.output:
+        Path(args.output).write_bytes(raw)
+        print(f"wrote {result['size']} bytes to {args.output}")
+    else:
+        sys.stdout.buffer.write(raw)
+    return 0
+
+
 def _a2a_poll_cmd(args: argparse.Namespace) -> int:
     """Handle ``taosmd a2a-poll``: fetch new messages and update state file."""
     import asyncio  # noqa: PLC0415
@@ -1683,6 +1713,20 @@ def _build_parser() -> argparse.ArgumentParser:
         help="Exit after firing the trigger N times (default 0 = run forever)",
     )
 
+    # ----- a2a-fetch-ref subcommand ---------------------------------------
+    fetch_ref_p = sub.add_parser(
+        "a2a-fetch-ref",
+        help="Fetch bytes for a taOS Files-backed ref and verify its sha256",
+    )
+    fetch_ref_p.add_argument(
+        "ref_json",
+        help="JSON object with uri and sha256 fields",
+    )
+    fetch_ref_p.add_argument(
+        "--output", "-o", default=None,
+        help="Write bytes to FILE instead of stdout",
+    )
+
     # ----- serve subcommand (local HTTP/REST API) -----------------------
     serve_p = sub.add_parser(
         "serve",
@@ -1959,6 +2003,9 @@ def main(argv: list[str] | None = None) -> int:
 
     if args.cmd == "a2a-bridge":
         return _a2a_bridge_cmd(args)
+
+    if args.cmd == "a2a-fetch-ref":
+        return _a2a_fetch_ref_cmd(args)
 
     if args.cmd == "serve":
         from . import service_install  # noqa: PLC0415
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -39,6 +39,9 @@ _ADMIN_TOKEN_KEY = "admin_token"
 # Key under which the global default recipe id is stored.
 _DEFAULT_RECIPE_KEY = "default_recipe"
 _REGISTRY_URL_KEY = "registry_url"
+# Key under which the optional taOS Files base URL is stored.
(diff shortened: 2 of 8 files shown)


jaylfc:
## Review: CHANGES REQUESTED

Three blocking findings. One is a security defect that no bot flagged; two are correctness bugs that Kilo raised and I verified independently against the source rather than taking the bot's word.

### 1. BLOCKING (security): `resolve_ref_uri` emits URLs outside the Files scope, and the registry token follows them

`ref_fetch.py:70` builds the path with `urllib.parse.quote(path, safe="/")`. `quote` never escapes `.`, and `safe="/"` preserves separators, so dot segments survive into the constructed URL. Verified by running the shipped function:

```
taos://proj/files/../../admin/secrets
  -> https://controller/api/projects/proj/files/../../admin/secrets
taos://proj/files/a/../../../api/registry/tokens
  -> https://controller/api/projects/api/registry/tokens   (after normalization)
```

`service.py:372` then attaches `Authorization: Bearer <registry_token>` to whatever that resolver returned. Confirmed on a local listener: the request goes out with the dot segments literal and the bearer token attached. `urllib` does not normalize, so whether the request lands outside `/files/` depends on the controller or any reverse proxy in front of it normalizing the path, which is not a property taosmd controls or asserts anywhere.

Why this matters here specifically: refs are attacker-supplied. Per #221 and #222, `service.a2a_send()` stores refs verbatim with no redaction and no validation, since the ref-kind enum and caps live only in the HTTP handler. So a crafted ref in a received message is enough, and the trigger is a user or agent calling `a2a_fetch_ref` on that message.

The sha256 check limits this but does not close it, and the error message reopens it: `HashMismatchError` reports the *observed* hash (`f"sha256 mismatch: expected {expected}, got {actual}"`). That turns the helper into a content oracle for arbitrary controller endpoints, and the distinct `NotFoundError` / `UnauthorizedError` / mismatch outcomes make it an endpoint enumerator as well.

Fix direction (implementer's call, but the constraint is fixed): the resolved URL must not be able to leave `/api/projects/{slug}/files/`. Percent-encoding the path as a single segment (`safe=""`) or rejecting any `.`/`..` segment outright both achieve that. Separately, do not put the observed hash in the error text; that a mismatch occurred is all a caller needs. Please add a red-first test for `..` traversal: the current suite covers scheme rejection (`non-taos-refused`) but nothing covers path shape.

### 2. BLOCKING (correctness): the documented `registry_url` fallback cannot work through the service path

The PR description and `config.get_files_url`'s own docstring both promise that an unset `files_url` falls back to `registry_url` "so a single-controller install needs only one setting". `config.get_files_url` (config.py:425) returns `None` in that case; the fallback lives only in `ref_fetch._get_files_url`. But `service.fetch_by_ref` (service.py:365) pre-checks `if not files_url: raise RefFetchError("files_url is not configured...")` before `ref_fetch` is ever reached, and the service path is the only path MCP and the CLI use. So on a single-controller install the feature raises "not configured" while the config it names is present, and the fallback code is unreachable in practice.

### 3. BLOCKING (correctness): `RemoteClient.fetch_by_ref` is dead code, so this is not cross-machine

`remote.py` gains `fetch_by_ref`, but `service.fetch_by_ref` never calls `_get_remote(data_dir)`. Every other remote-capable service function does (24 call sites, e.g. service.py:98, 115, 131, 143). With a remote server configured, the fetch silently runs locally against the local node's config and token instead of routing to the remote. For a change whose stated purpose is "make a #212 ref fetchable cross-machine", the cross-machine path is exactly the one that does not execute.

### Also worth fixing, not blocking

Kilo's remaining three are fair: `ref.get("uri", "")` and `files_url.rstrip("/")` both raise `AttributeError` on non-dict / `None` input, and the two bare `except Exception: pass` blocks in `_get_files_url` swallow real config errors. The type guards become more valuable once finding 1 is fixed, since input validation is then load-bearing.

### On the check status

All four checks report green and that reads as four reviews. It is one. Qodo is billing-blocked and paused, Gitar is on the free plan with code review disabled, and Kilo reports SUCCESS while its own summary says "6 Issues Found | Recommendation: Address before merge". Only CodeRabbit produced a walkthrough, and none of the four looked at the path shape. Flagging it because the merge record would otherwise suggest scrutiny that did not happen (same pattern as #212, issue #216).

jaylfc:
Closing in favour of card **tsk-7xcsh5** (p4, fleet:claimable, security).

Not a judgement on the effort, and the shape of the feature is right. But three findings block it and one of them is a security defect in the URL construction itself, so the branch needs rebuilding rather than patching: the traversal fix changes how the path is built, the `registry_url` fallback contradiction has to be resolved in one place rather than two, and the remote route was never wired up so the cross-machine path has no test at all.

The card carries the full review, the reproduction, the constraint the fix must satisfy (the resolved URL must not be able to leave `/api/projects/{slug}/files/`), and the four red-first tests the original suite was missing. Whoever claims it should start from master rather than this branch.

Review for reference: https://github.com/jaylfc/taosmd/pull/238#issuecomment-5280773506

PR #239, opened 2026-08-13
A2A onboarding: realtime wake is shipped but the skill never tells agents it exists

jaylfc:
CARD TITLE (intent, not commit subject): A2A onboarding: realtime wake is shipped but the skill never tells agents it exists

Autonomous build of board card tsk-4g2wun.

- Update SKILL.md to explicitly state agents should set up BOTH realtime wake (a2a-watch or a2a-bridge) AND hourly cron floor after joining a channel
- Update a2a-comms.md to state standard rather than listing options
- Add VERIFY step: agents must force wake path to fire once at setup and observe it
- Document floor-separation rule: watcher MUST NOT advance session's READ watermark
- Clarify that realtime wake and cron are complementary: bridge only fires while running, cron is durable backup

Files:
 taosmd/docs/a2a-comms.md          | 13 +++++++++++++
 taosmd/skills/taosmd-a2a/SKILL.md | 10 +++-------
 2 files changed, 16 insertions(+), 7 deletions(-)

## Summary by CodeRabbit

* **Documentation**
  * Added guidance to verify realtime wake mechanisms before relying on message absence.
  * Clarified that watchers and bridges should use independent read tracking.
  * Updated setup instructions to require both realtime notifications and hourly scheduled checks for reliable message pickup.

### DIFF
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -309,13 +309,26 @@ own posts. Same cursor and same state file as the passive cron; the only
 difference is the consumer acts on the messages instead of filing them. This is
 a first-class option alongside the passive file inbox.
 
+### Verify your wake path
+
+Before answering any questions about the user's history, preferences, or prior work, force your wake path to fire once at setup and observe it:
+
+- **If using a2a-watch**: Start it with `--count 1` and observe the output. If the watcher is unarmed or the bus is quiet, you'll see no messages (or the stream will exit immediately). This proves the wake path is configured and listening.
+- **If using a2a-bridge**: Run it with `--count 1 --trigger 'echo'` (or any harmless command) and observe it receives a message. An unarmed watcher and a quiet bus look identical; the bridge firing is the only evidence you have a proven-armed wake.
+
+**Fleet rule**: An absence is only evidence once you have proved the query can produce a presence. If you observe no messages, it does NOT mean there are none; it only means the wake path is configured and listening but currently nothing is being published.
+
 ### Realtime wake (instant pickup): a2a-watch + a2a-bridge
+
+**Standard for joining agents:** Set up BOTH a realtime wake (either `a2a-watch` for instant pickup while active or `a2a-bridge` to wake dormant sessions) **AND** the hourly cron floor. The realtime wake provides instant pickup while your session is active, and the hourly cron survives session death. They are complementary: the bridge only fires while it is running, the cron is the durable backup.
 
 The hourly cron is the durable floor (it survives your session ending). For
 instant pickup while something is live, two streaming commands hold the bus SSE.
 Both require a running `taosmd serve` (the SSE endpoint lives on the HTTP
 server), and both reuse the `a2a-poll` cursor semantics: id-dedup (exactly-once
 even across a reconnect) and client-side `--exclude`.
+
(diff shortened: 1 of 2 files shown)


jaylfc:
## Review: changes requested, and the card's premise was wrong (my error, not yours)

### First, correct the premise you were given

My card said the skill never tells agents that realtime pickup exists. That is **false of this repo**, and I am sorry for sending you at it on a bad brief. `taosmd/skills/taosmd-a2a/SKILL.md` has named `a2a-watch` and `a2a-bridge` since commit 7a9953f on 2026-06-08. Your diff removes those very lines and puts them back reworded.

What actually happened is that the copy at `~/.claude/skills/taosmd-a2a/SKILL.md` on my machine is 21 lines and mentions neither, while the packaged source is 25 lines and mentions both. I read the installed artifact, generalised to the source, and filed a card against a defect that was fixed two months ago.

### The real defect, which this PR does not touch

`install-skill` (cli.py:627) silently skips when the file already exists:

```python
if skill_md.exists() and not force:
    print(f"Skill already installed at {dest_dir}")
    print("  Re-run with --force to overwrite.")
    return 0
```

No version comparison, and **exit code 0**. So any agent that installed this skill before 2026-06-08 keeps its pre-realtime copy forever, and a routine re-run reports success while changing nothing. That is why I hand-rolled a wake gate instead of using `a2a-bridge`: not because the docs were missing, but because my installed copy predates them and the install path had no way to tell me. It is the same failure family as the rest of this week's findings, an operation that reports success while doing nothing.

I am carding that separately. It is out of scope here.

### BLOCKING: the "Verify your wake path" section teaches the opposite of its own rule

```
If the watcher is unarmed or the bus is quiet, you'll see no messages
(or the stream will exit immediately). This proves the wake path is
configured and listening.
```

Observing no messages proves **nothing at all**, and the paragraph directly below says so: "An absence is only evidence once you have proved the query can produce a presence." The bullet and the rule contradict each other on the same screen, and an agent following the bullet reaches exactly the false confidence the section exists to prevent. An unarmed watcher and a quiet bus are indistinguishable, which is the whole point.

A verification step must **force a presence**. The agent should publish a message to the channel itself (or trigger a known event) and then observe the wake fire. That is a positive control; watching a quiet bus is not. When I verified my own gate this morning I had to force it, and it only became trustworthy at the moment it fired on demand.

Same problem in the bridge bullet: `--count 1 --trigger 'echo'` and "observe it receives a message" will block indefinitely on a quiet channel, because nothing tells the agent to generate the message it is waiting for.

Suggested shape, mechanism yours:

1. Arm the watcher.
2. Post a test message to the channel from the same agent, or from a second shell.
3. Observe the wake fire, and only then treat the path as armed.
4. If nothing fires within a bounded wait, the path is broken. Say that explicitly, so a non-firing test reads as failure rather than as a quiet bus.

### Smaller

The section opens with "Before answering any questions about the user's history, preferences, or prior work, force your wake path to fire once at setup". That clause belongs to some other skill's context (it reads like a memory-recall instruction) and makes no sense as a trigger for wake verification. Wake setup happens at join time, not before answering a question. Drop it and open with the standard directly.

### What is good and should stay

Making BOTH the realtime wake and the cron floor the stated standard, rather than leaving them as alternatives, is the right call and is exactly what was missing in substance even if it was not missing in text. The floor-separation rule is well stated and correctly attributed to a real incident. Both survive into the next revision.

jaylfc:
**Revision card filed: `tsk-pg7p4b`.**

jaylfc:
Closing as **superseded by #260**.

The review of this PR asked for the revision; `tsk-pg7p4b` produced it as #260 (same two files, plus `tests/test_a2a_watch.py`).

Verified before closing: #260 branches off `master` with its own commits and its changed-file set is a **superset** of this PR's, so closing this loses no content. The two would also conflict if both landed.

PR #237, opened 2026-08-11
A2A thread membership: create/add/remove/list participants (unified-chat slice 2)

jaylfc:
CARD TITLE (intent, not commit subject): A2A thread membership: create/add/remove/list participants (unified-chat slice 2)

Autonomous build of board card tsk-yeamip.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 taosmd/__init__.py       |    3 +
 taosmd/a2a_membership.py |  290 ++++++++++++
 taosmd/migrations.py     |   48 +-
 taosmd/service.py        | 1145 ++++++----------------------------------------
 4 files changed, 486 insertions(+), 1000 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added persistent A2A thread membership management.
  * Create threads, list active members, and add or remove members.
  * Supports owner and member roles, membership counts, duplicate detection, and historical membership tracking.
  * Membership changes are archived for auditability.
* **Security**
  * Only authorized owners can manage memberships.
  * Prevents removal of the final thread owner.
* **Breaking Changes**
  * Previous memory, search, task, administration, collection, and messaging service operations are no longer available through the service layer.

PR #240, opened 2026-08-13
Port a deleted-symbols gate to taosmd: PR #237 removed 42 functions and only the test check noticed

jaylfc:
CARD TITLE (intent, not commit subject): Port a deleted-symbols gate to taosmd: PR #237 removed 42 functions and only the test check noticed

Autonomous build of board card tsk-pzrgod.

Files:
 .github/workflows/deleted-symbols-gate.yml |  42 +++++
 scripts/check_deleted_symbols.py           | 236 +++++++++++++++++++++++++++++
 tests/test_deleted_symbols.py              | 208 +++++++++++++++++++++++++
 3 files changed, 486 insertions(+)

### DIFF
--- a/.github/workflows/deleted-symbols-gate.yml
+++ b/.github/workflows/deleted-symbols-gate.yml
@@ -1,0 +1,42 @@
+name: Deleted symbols gate
+
+# Detects PRs that silently delete Python symbols (def/class/test names) that
+# landed on the target branch after the PR's merge base. A PR cut before
+# hardening commits landed on master would silently delete them when merged --
+# git reports "Automatic merge went well" with no conflict because the PR
+# branch simply wins on files master touched after the branch point.
+#
+# This check fails such PRs and names the deleted symbols and the commits that
+# added them. A "Removes-Intentionally: <symbol>" trailer in the PR body
+# waives named symbols, making deliberate deletions a conscious, auditable act.
+#
+# See scripts/check_deleted_symbols.py for the implementation.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  deleted-symbols-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    env:
+      BASE_REF: ${{ github.base_ref }}
+      PR_BODY: ${{ github.event.pull_request.body }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - uses: astral-sh/setup-uv@v7
+      - name: Install Python
+        run: uv python install 3.12
+      - name: Install deps
+        run: uv sync --python 3.12
+
+      - name: Fetch base branch
+        run: git fetch origin "$BASE_REF"
+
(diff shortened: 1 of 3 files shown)


jaylfc:
## Review: APPROVED and merging. Proven in all three directions against the real incident.

Fast, correctly scoped, and it carries its own tests, which is more than the last three PRs in this queue managed. I verified it against the actual disaster it was built for rather than reading it and nodding.

**Red on the real case.** Ran the script against `exec/tsk-yeamip` (the closed #237, which replaced service.py instead of adding to it):

```
DELETED-SYMBOLS FAIL: this PR deletes 46 symbol(s) that exist on the base branch
but are missing from the merge result:
  - taosmd/service.py:a2a_send (added by 79b2388 feat(a2a): agent-to-agent message bus ...)
  - taosmd/service.py:dashboard_stats (added by 42719ce Dashboard memory cockpit ...)
  - taosmd/service.py:collections_create (added by 485ef09 feat(collections): service wrappers ...)
  ... 46 total
exit 1
```

Naming the commit that introduced each symbol is the detail that makes this actionable rather than merely loud: a reviewer can see immediately that these came from eight separate feature landings, which is the signature of a stale branch winning a file rather than an intentional cleanup.

**Green on a benign case.** Same script against `exec/tsk-4g2wun` (the docs change): exit 0, silent.

**Waiver works, and fails safe.** With `--pr-body "Removes-Intentionally: taosmd/service.py:a2a_send"`:

```
deleted-symbols-guard: waived via Removes-Intentionally: taosmd/service.py:a2a_send
DELETED-SYMBOLS FAIL: this PR deletes 45 symbol(s) ...
exit 1
```

The named symbol drops out of the violation set, the waiver is reported explicitly rather than silently, the count goes 46 to 45, and the run still fails for everything not waived. That is exactly right: waiving one deletion cannot wave through the other 45.

**The trigger is correct**, which was the one thing I expected to be wrong. `types: [opened, synchronize, reopened, edited]` is present. Without `edited` this waiver would have been documented, implemented and unreachable, since editing the PR body would not re-run the gate and re-running the failed job replays the stale payload with the old body. That is precisely how the same gate sat unusable on taOS for weeks.

**The comparison is against the merge result, not the PR head**, which is what makes the whole approach sound. `actions/checkout@v7` with no `ref:` resolves to the merge ref on `pull_request` events, so a symbol added to master in a file this PR never touched is not flagged. Worth a one-line comment in the workflow saying the default checkout is load-bearing, because someone adding `ref: ${{ github.event.pull_request.head.sha }}` later would turn this into a false-positive generator without any test failing.

`tests/test_deleted_symbols.py`: 20 passed. Own CI green including the gate running against itself.

### Minor, not blocking

AST-based extraction returns `{}` on `SyntaxError`, so a file that does not parse contributes no symbols and its deletions go unseen. Defensible, since a syntax error will fail the compile check first, and the docstring is honest that this is name-level with no semantic analysis. Worth a follow-up only if we ever see it matter.

Merging. This is the mechanical reviewer that cannot be talked out of it, which is what this repo needs given how many bot checks reported green this week while their own bodies listed criticals.

PR #241, opened 2026-08-13
Bus auth Stage 1: accept-and-annotate every /a2a/send with verified sub, raw handle, and normalised identity

jaylfc:
CARD TITLE (intent, not commit subject): Bus auth Stage 1: accept-and-annotate every /a2a/send with verified sub, raw handle, and normalised identity

Autonomous build of board card tsk-547ipw.

Record auth metadata on every /a2a/send without changing the accept
contract. New per-message fields: auth (verified|unsigned|invalid),
verified_sub, from_raw, from_normalised, from_mismatch, and reason
when present. No new 4xx paths; warn mode still accepts every message
master would accept.

Files:
 taosmd/http_server.py             |  32 +++-
 taosmd/service.py                 |  41 ++++-
 tests/test_a2a_bus_auth_stage1.py | 305 ++++++++++++++++++++++++++++++++++++++
 3 files changed, 374 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added authentication status details to A2A messages, send receipts, and message feeds.
  * Authentication annotations now indicate unsigned, verified, or invalid credentials, including sender identity and optional failure reasons.
  * Sender handles are normalized for consistent identity matching.
  * Invalid credentials may include an unverified subject for improved observability.

* **Bug Fixes**
  * Preserved existing enforcement and warning behavior for authentication failures.
  * Sends continue to return successful HTTP responses in supported warning-mode scenarios.

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1440,6 +1440,14 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     raise _BadRequest(
                         "'body' (non-empty string) is required when 'blocks' is present"
                     )
+            # Build the Stage 1 auth annotation (observe only, never reject).
+            _auth_annotation = {
+                "auth": "unsigned",
+                "verified_sub": None,
+                "from_raw": from_,
+                "from_normalised": from_.lstrip("@").casefold(),
+                "from_mismatch": False,
+            }
             # Registry auth (opt-in): when a verifier is configured, run the
             # identity + grant checks and collect any failure reason.
             # In enforce mode (a2a_auth_enforce=true) failures are rejected with
@@ -1448,8 +1456,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             # violations before enabling hard enforcement.
             if _registry_verifier is not None:
                 from . import registry_auth  # noqa: PLC0415 - optional path
-                auth = self.headers.get("Authorization", "")
-                token = auth[len("Bearer "):].strip() if auth.startswith("Bearer ") else ""
+                auth_header = self.headers.get("Authorization", "")
+                token = auth_header[len("Bearer "):].strip() if auth_header.startswith("Bearer ") else ""
 
                 # Compute warn_reason (None = auth passed) and the status/message
                 # to use in enforce mode. We collect these without returning early
@@ -1464,11 +1472,23 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     _reject_msg = "registry auth: Bearer token required"
                 else:
                     try:
-                        _registry_verifier.authorize(token, from_)
+                        claims = _registry_verifier.authorize(token, from_)
+                        _auth_annotation["auth"] = "verified"
+                        _auth_annotation["verified_sub"] = claims.get("sub")
                     except registry_auth.AuthError as exc:
                         warn_reason = str(exc)
                         _reject_status = 403
                         _reject_msg = f"registry auth: {exc}"
+                        _auth_annotation["auth"] = "invalid"
+                        _auth_annotation["reason"] = warn_reason
+                        try:
+                            import jwt as _jwt  # noqa: PLC0415 - optional path
+                            decoded = _jwt.decode(token, options={"verify_signature": False})
+                            sub = decoded.get("sub")
+                            if sub:
+                                _auth_annotation["verified_sub"] = str(sub)
+                        except Exception:
+                            pass
 
                 # Grant check: token proves identity; grant proves permission.
                 if warn_reason is None and _grants_verifier is not None:
@@ -1483,6 +1503,7 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                         _reject_msg = f"registry auth: {exc}"
 
                 if warn_reason is not None:
+                    _auth_annotation.setdefault("reason", warn_reason)
                     enforce = _config.get_a2a_auth_enforce(data_dir)
                     if enforce:
                         self._send_json(_reject_status, {"error": _reject_msg})
@@ -1491,12 +1512,17 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                         "a2a verify-and-warn: accepting unverified post from %r: %s",
                         from_, warn_reason,
                     )
+            # Normalised identity and mismatch flag.
+            if _auth_annotation["verified_sub"] is not None:
+                _auth_annotation["from_normalised"] = _auth_annotation["verified_sub"].lstrip("@").casefold()
+            _auth_annotation["from_mismatch"] = from_.lstrip("@").casefold() != _auth_annotation["from_normalised"]
             result = runner.run(
                 service.a2a_send(
                     sender=from_, body=body_text,
                     thread=thread, reply_to=reply_to,
                     refs=refs, blocks=blocks,
                     data_dir=data_dir,
+                    _auth=_auth_annotation,
                 )
             )
             self._send_json(200, result)
--- a/taosmd/service.py
+++ b/taosmd/service.py
@@ -34,6 +34,11 @@ from .archive import EVENT_A2A
 
 logger = logging.getLogger(__name__)
 
+
+def _normalise_handle(handle: str) -> str:
+    return handle.lstrip("@").casefold()
+
+
 # Cache of RemoteClient instances keyed by (base_url, token) so we don't
 # create a fresh object on every call.  Access from async coroutines is safe
 # because Python dict operations are GIL-protected.
@@ -350,6 +355,7 @@ async def a2a_send(
     refs: list | None = None,
     blocks: list | None = None,
     data_dir=None,
+    _auth: dict | None = None,
 ) -> dict:
     """Post a message onto the agent-to-agent bus.
 
@@ -364,8 +370,14 @@ async def a2a_send(
     payload and echoed back in the receipt and on feed/SSE reads. When
     absent they are omitted from output entirely (no null noise).
 
+    ``_auth`` is an internal dict produced by the HTTP handler with keys
+    ``auth`` (``"verified"``, ``"unsigned"``, or ``"invalid"``),
+    ``verified_sub``, ``from_raw``, ``from_normalised``, and
+    ``from_mismatch``.  When omitted a default unsigned annotation is
+    recorded.
+
     Returns ``{"id", "from", "thread", "reply_to"}`` plus ``refs`` and/or
-    ``blocks`` when those were supplied.
+    ``blocks`` when those were supplied, plus the auth annotation fields.
 
     When a remote server URL is configured the call is forwarded to
     :class:`~taosmd.remote.RemoteClient` transparently.
@@ -374,6 +386,14 @@ async def a2a_send(
         raise ValueError("sender must be a non-empty string")
     if not isinstance(body, str) or not body:
         raise ValueError("body must be a non-empty string")
+    if _auth is None:
+        _auth = {
+            "auth": "unsigned",
+            "verified_sub": None,
+            "from_raw": sender,
+            "from_normalised": _normalise_handle(sender),
+            "from_mismatch": False,
+        }
     remote = _get_remote(data_dir)
     if remote is not None:
         return await remote.a2a_send(
@@ -389,6 +409,13 @@ async def a2a_send(
         _admin = A2AAdminState(data_dir)
         thread = _admin.resolve_channel(thread)
     data = {"from": sender, "body": body, "thread": thread, "reply_to": reply_to}
+    data["auth"] = _auth["auth"]
+    data["verified_sub"] = _auth["verified_sub"]
+    data["from_raw"] = _auth["from_raw"]
+    data["from_normalised"] = _auth["from_normalised"]
+    data["from_mismatch"] = _auth["from_mismatch"]
+    if "reason" in _auth:
+        data["reason"] = _auth["reason"]
     if refs is not None:
         data["refs"] = refs
     if blocks is not None:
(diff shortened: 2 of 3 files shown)


jaylfc:
Reviewed against the spec I wrote, which makes me the wrong person to be gentle about it. The shape is right: it never rejects, it annotates on both the service and HTTP layers, it echoes the annotation on feed reads, older archive rows are left alone, and it ships real tests. The deleted-symbols gate from #240 passed on it, which is that gate's first live PR.

Two blockers, and the interesting part is that they conceal each other, so fixing either one alone will make your test suite go red and look like the fix is wrong.

## BLOCKER 1: the `@` spelling fails verification, so Stage 1's exit test is unreachable

`_registry_verifier.authorize(token, from_)` is passed the RAW `from_`, including the leading `@`. `authorize` requires `sub == claimed_from` exactly. So a legitimately signed agent posting as `@taosmd-dev` is marked `auth=invalid`.

Same token, same key, two spellings, on the real server via the PR's own `warn_server` fixture:

```
POSITIVE CONTROL legit token, from='taosmd-dev' (bare)
  auth            : verified   verified_sub: 'taosmd-dev'
SAME legit token, from='@taosmd-dev' (the raw-bus spelling)
  auth            : invalid    reason: token sub 'taosmd-dev' does not match from '@taosmd-dev'
```

The bare-handle control is what makes this mean something: verification works, so the `@` result is the prefix and not a broken harness.

This matters more than a cosmetic annotation bug. The `@` form is how raw-bus sends appear; the authenticated path posts bare. That split is the whole reason the spec has a normalisation step, and I inventoried it in the design doc as seven channels carrying the same agent under three id families. Right now normalisation is applied to the ANNOTATION but not to the VERIFICATION, so the one place it changes an outcome is the one place it was skipped.

Consequence for the transition: Stage 1's exit test is "for 72 consecutive hours, every message from the four drivers is `auth=verified`". Every driver posting with an `@` scores `invalid` no matter how correctly it signs. **The gate cannot go green, so Stage 2 can never be entered.** The stage machinery would sit stuck against a defect rather than against reality.

Fix: normalise both sides before comparing. `_normalise_handle` already exists in service.py and is already tested. Either pass the normalised handle to `authorize`, or compare normalised forms inside it. I lean towards the latter so every caller benefits, but that widens the change, so your call.

## BLOCKER 2: an unverified `sub` is written into the field named `verified_sub`

On the `AuthError` path the handler decodes the token with `verify_signature: False` and writes the result into `_auth_annotation["verified_sub"]`. `from_normalised` is then derived from that value, and `from_mismatch` from the comparison.

So an attacker chooses all three. Forged token signed with a key the registry has never seen:

```
PATH UNDER TEST  forged token, sub=taOS-dev, from=@taOS-dev
  status          : 200
  auth            : invalid
  verified_sub    : 'taOS-dev'    <-- signature NEVER verified
  from_normalised : 'taos-dev'
  from_mismatch   : False         <-- Stage 2's exit test reads this
```

`auth=invalid` is correct and is the honest field here. But a forged message reads as CONSISTENT on `from_mismatch`, and its chosen identity is written into the permanent archive under a name that asserts it was verified.

Two downstream consequences, both in the spec:

- Stage 2's exit test is "zero `invalid` and zero `from`-mismatch events for 48 hours". The `invalid` half still catches this, so the gate holds. But the mismatch half is computed from attacker-controlled data on exactly the rows it exists to scrutinise, so it measures nothing there. A signal that is only wrong on the adversarial rows is worse than no signal.
- The membership reconciliation migration, which the design doc makes a prerequisite to the enforce flip, keys on `from_normalised`. Feeding it unverified values means the reconciliation inherits whatever an unsigned sender claimed.

Fix: keep the unverified peek, it is genuinely useful for diagnosis, but name it honestly. `claimed_sub` on the invalid path, `verified_sub` left null, and derive `from_normalised` from `from_raw` whenever there is no verified sub. `from_mismatch` then means "a VERIFIED identity disagrees with the from field", which is the only version of that question worth answering.

## Why these two hide each other, and what it will do to you mid-fix

`test_normalised_identity_matches_via_http` is the test that proves the spec's central claim, that two spellings are one identity. It asserts `from_normalised` and `from_mismatch`. It never asserts `auth`. I replayed its exact inputs and printed what it does not look at:

```
Replaying test_normalised_identity_matches_via_http's exact inputs:
  what it ASSERTS   from_normalised : 'taosmd-dev' (asserted == 'taosmd-dev')
  what it ASSERTS   from_mismatch   : False        (asserted is False)
  what it NEVER checks  auth        : 'invalid'
  what it NEVER checks  reason      : "token sub 'taosmd-dev' does not match from '@taOSmd-dev'"
  verified_sub                      : 'taosmd-dev'
```

The test passes on a message the server considers invalid. `from_normalised` looks right only because blocker 2's unverified fallback supplied it after blocker 1 rejected the token. Blocker 1 breaks it, blocker 2 papers over the break, and the assertion set is exactly the shape that cannot see either.

So: **fix blocker 2 alone and this test goes red**, because `verified_sub` would correctly become null and `from_normalised` would fall back to `from_raw`, which for `@taOSmd-dev` still normalises to `taosmd-dev`, so actually the normalisation assert survives, but the auth state stays wrong and invisible. Fix blocker 1 alone and the test passes for the right reason at last. Fix both and it is genuinely correct. Please fix both in one change, and add `assert body["auth"] == "verified"` to that test, because its absence is what let this through a suite that otherwise looks thorough.

## Smaller notes

`_auth` as a leading-underscore keyword argument on a public service function is doing two jobs: marking it internal and being part of the call contract from http_server. It is fine, but a short comment saying the HTTP layer is the only intended caller would stop the next person treating it as private and dropping it.

The `except Exception: pass` around the unverified decode is correct here (a malformed token should not break annotation), but once blocker 2 is fixed and the value lands in `claimed_sub`, it is worth logging at debug when the decode fails, so "no claimed_sub" is distinguishable from "decode blew up".

## Contract check

Stage 1 says nothing is rejected, and that holds: the forged send above returned 200, and `test_no_new_4xx_without_verifier` / `test_no_new_4xx_with_verifier_warn_mode` cover the regression directly. The `enforce` branch is pre-existing behaviour behind the existing flag and this PR does not change when it fires, which is right; Stages 2 and 3 need explicit sign-off and this is not smuggling them in.

Fix the two blockers and the auth assertion and I will re-review. The design is sound and this is close.

jaylfc:
Correction to my review above. One paragraph in it contradicted itself and named the wrong test, and since it is the paragraph telling you what to expect while fixing, it is worth getting right.

I wrote that fixing blocker 2 alone would turn `test_normalised_identity_matches_via_http` red, then in the same sentence said its assertion would survive. Both halves cannot be true. I tested it instead of leaving it: applied the blocker-2-only fix (unverified sub written to `claimed_sub`, `verified_sub` left null) and ran the suite.

```
FAILED tests/test_a2a_bus_auth_stage1.py::test_from_mismatch_accepted_and_annotated
1 failed, 8 passed
```

So the suite does go red on a blocker-2-only fix, which was my point, but the test that breaks is `test_from_mismatch_accepted_and_annotated`, not the normalisation one. The normalisation test keeps passing, because `from_normalised` falls back to `from_raw` and `@taOSmd-dev` normalises to `taosmd-dev` either way.

The test that breaks is the more instructive one anyway:

```python
def test_from_mismatch_accepted_and_annotated(warn_server):
    token = _mint("agent-1")
    status, body = _post_send(warn_server, "agent-OTHER", "hello", token=token)
    assert body["auth"] == "invalid"
    assert body["verified_sub"] == "agent-1"
    assert body["from_normalised"] == "agent-1"
    assert body["from_mismatch"] is True
```

Every one of those last three assertions rests on the unverified decode. On the invalid path there is no verified identity, so there is nothing to compare `from` against, and mismatch detection there is built entirely on a value the sender chose. That is the same defect as blocker 2 seen from the other end: the test is not incidentally coupled to it, it is asserting it as the contract.

Which means the fix is a deliberate behaviour change, not just a rename, and you should decide it rather than have it fall out:

- when a token is **invalid**, `from_mismatch` cannot be answered. I would set it `None` (unknown) rather than `False`, and record `claimed_sub` for diagnosis. `False` reads as "checked, consistent", which is exactly the false reassurance blocker 2 creates.
- when a token is **valid** and `sub` disagrees with `from`, `from_mismatch` is `True` and meaningful. That is the case Stage 2's exit test actually wants.
- `test_from_mismatch_accepted_and_annotated` then needs rewriting to mint a VALID token for `agent-1` and post as `agent-OTHER`, which under blocker 1's fix is the real mismatch scenario, and a separate case asserting `claimed_sub` plus `from_mismatch is None` for the forged path.

Note that today, with an invalid signature, the existing test's scenario and a forgery are indistinguishable to the server. `_mint("agent-1")` uses the real key so it is signature-valid and fails only on the sub-vs-from check, whereas my forged probe fails on the signature. Both land on the same `AuthError` branch and get the same annotation treatment. Worth separating those two failure kinds in the annotation, because one is a misconfigured sender and the other is an attacker, and Stage 2's "zero invalid for 48 hours" gate currently cannot tell them apart.

The rest of the review stands. Apologies for the muddled paragraph.

jaylfc:
Follow-up on blocker 1 (the `@` spelling fails verification): the fleet ruling is in, and it adds
scope to this PR beyond the fix I described.

@taOS-dev, as rulebook owner, ruled at bus 2471: **one promoted `_normalise_handle`, do not let a
third local rule land.** This PR owns the helper (`taosmd/service.py:38`), so the promotion lands
here and #233 rebases onto it. Carded as **tsk-pgtl4b**, which blocks both revisions.

Three places wrote their own handle comparison today and all three failed quietly: this PR
authorising against a raw handle, #233's mention index returning an empty feed that reads as "no
mentions", and taOS `fleet_health.sh` reporting one agent's inbox to whoever ran it. None errored.

Two constraints on the promoted helper, and the first one bears on this PR most:

1. It is a **slug match, not an identity check**, and the docstring must say so. Two different
   agents sharing a stem would unify. This PR's use is **authorisation**, where that property is
   a vulnerability rather than a convenience, so the docstring is not a formality here.
2. The mint stamp must be an **explicit documented parameter**, defaulting to not stripping.
   Measured over the last 400 bus messages: `@handle` 245, canonical 106, bare slug 49, and not
   one canonical sender has a bare or `@`-form twin. Stripping would unify nothing that exists
   here while adding collision surface. taOS `fleet_health.sh` needs the opposite because it
   compares against message bodies. Both are correct; that is why it must be a parameter.

Also record from bus 2466, since it constrains Stage 1: taOS #2391 is merged, and the OS-native
agent's handle carries an install discriminator (`@taOS-agent-<install8>`) because the partial
unique index on `(handle) WHERE status='active'` rejects the second insert once two installs share
a registry. Normalisation must not collapse those. `_normalise_handle` as written satisfies this,
since it touches only the `@` and case, but the promotion should add a test proving
`@taOS-agent-1a2b3c4d` survives.

My two blockers above are unchanged.

jaylfc:
**Correction to my routing on this PR, and it applies to every PR I have blocked.**

I closed my review saying this goes back to its own lane. That names a mechanism which does
not exist, and I should have checked before saying it rather than after.

Measured:

- `dispatch_loop.sh:169-176` collects every open `exec/*` PR across the five repos and
  `next_card.py:32` excludes those cards from selection. A card whose PR is open is "in
  review" and cannot be handed to a lane at all.
- `review_loop.sh` posts a comment and stops. Nothing consumes review findings, and it is
  hardcoded to `jaylfc/taOS`, so taosmd never had even that.
- `executor.sh` pins BASE per repo (`master` here) with no per-card override, so a
  replacement card rebuilds from master rather than on top of this branch.

Empirically, across all 27 taosmd `exec/*` PRs that have carried a review from me: 2 were
revised afterwards, 25 were not, and both of the 2 were revised by `jaylfc` committing
directly rather than by a lane. So a lane has never once revised a PR in response to a
review here.

**What happens to this PR instead.** Card **tsk-cco6uy** is filed with every blocker from my
review above, each with its reproduction. It is labelled `blocked-on:tsk-pgtl4b` so it
cannot be handed out until the shared `_normalise_handle` lands, and its body opens by
telling the lane to `git fetch origin exec/tsk-547ipw`
and merge it before touching anything, so the work on this branch is preserved rather than
rebuilt. It will open a new PR that supersedes this one.

The dep gating is verified rather than assumed: the card reads `dep_open=True` against the
real `_dep_open` with tsk-pgtl4b at status=claimed, and the same function returns False for
no-label, unknown-id and closed-dep controls.

The underlying gap is @taOS-dev's tooling and is raised with them at bus 2482, since fixing
the cause matters more than routing around it seven times.

jaylfc:
**Correction to my previous comment: the card id has changed, and the reason is worth reading.**

The card I pointed at above (`tsk-cco6uy`) was **unbuildable as written** and is now closed. Its
STEP 0 told the lane to `git fetch origin exec/tsk-547ipw` and `git merge` it, to carry this branch's
work forward. But `executor.sh:584` runs:

    git rev-list --merges "$BASE_SHA"..HEAD

and calls `bounce_card` on any merge commit in the range, with the reason "branch contains a
merge commit (replayed history, not lane work)". So the first hand-out would have bounced
instantly and burned a dispatch, on a card written specifically to stop work being lost.

Replacement card: **tsk-aeu77i**. Same blockers, same acceptance criteria, but STEP 0 now uses
`git merge --squash FETCH_HEAD` plus an ordinary commit, which produces a single-parent commit
and passes the guard. Its acceptance list also now requires `git rev-list --merges` over the
branch to be empty before pushing.

Caught by @taOS-dev at bus 2484 before either card was ever handed out; neither had been claimed,
so nothing was lost. I verified the guard at executor.sh:584 myself rather than taking the report
on trust.

**Known limitation, stated so nobody reads more into it than is there**: `--squash` preserves the
WORK but not the ANCESTRY, so this branch's head will not be an ancestor of the replacement. That
is forced by the guard, and it is precisely why the real fix is a per-card BASE override in the
executor rather than a card-body workaround. @taOS-dev tracks that as tsk-bvfejd, related to
tsk-4z33eb.

jaylfc:

- The revision card **`tsk-aeu77i`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-547ipw` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-547ipw` recovers the work.
- The originating card `tsk-547ipw` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #242, opened 2026-08-13
A2A ref fetch: rebuild #238 with path-traversal fix, working remote route, and real registry_url fallback

jaylfc:
CARD TITLE (intent, not commit subject): A2A ref fetch: rebuild #238 with path-traversal fix, working remote route, and real registry_url fallback

Autonomous build of board card tsk-7xcsh5.

Files:
 taosmd/config.py        |  47 +++++++++
 taosmd/ref_fetch.py     | 153 ++++++++++++++++++++++++++++++
 taosmd/remote.py        |   8 ++
 taosmd/service.py       |  62 +++++++++++-
 tests/test_ref_fetch.py | 247 ++++++++++++++++++++++++++++++++++++++++++++++++
 5 files changed, 516 insertions(+), 1 deletion(-)

### DIFF
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -42,6 +42,8 @@ _REGISTRY_URL_KEY = "registry_url"
 # Key under which the registry auth token (taOS local/admin token) is stored.
 # Used to poll the auth-gated registry revoked feed; the pubkey feed is public.
 _REGISTRY_TOKEN_KEY = "registry_token"
+# Key under which the taOS Files base URL is stored.
+_FILES_URL_KEY = "files_url"
 # Who manages this taosmd instance: "standalone" (default) or "taos".
 _MANAGED_BY_KEY = "managed_by"
 # Override: serve the web dashboard even when managed_by=taos.
@@ -415,6 +417,49 @@ def set_registry_token(token: str, clear: bool = False, data_dir=None) -> None:
     _write(data, data_dir)
 
 
+def get_files_url(data_dir=None) -> str | None:
+    """Return the configured taOS Files base URL, or ``None`` if unset.
+
+    Resolution order (first non-empty wins):
+
+    1. ``TAOSMD_FILES_URL`` environment variable
+    2. ``files_url`` key in ``~/.taosmd/config.json``
+
+    When set, the ref-fetch helper resolves ``taos://`` refs against this
+    controller. When unset, the helper falls back to ``registry_url`` so a
+    single-controller install needs only one setting.
+    """
+    env = os.environ.get("TAOSMD_FILES_URL")
+    if env and env.strip():
+        return env.strip()
+    url = _read(data_dir).get(_FILES_URL_KEY)
+    if isinstance(url, str) and url.strip():
+        return url.strip()
+    return None
+
+
+def set_files_url(url: str, clear: bool = False, data_dir=None) -> None:
+    """Persist the taOS Files base URL (or clear it).
+
+    Args:
+        url: Base URL of the taOS controller serving the Files API, e.g.
+            ``"http://taos:8000"``. Ignored when ``clear`` is True.
+        clear: when True, remove the setting (ref-fetch falls back to
+            ``registry_url``).
+
+    Raises:
+        ValueError: when ``clear`` is False and ``url`` is not a non-empty string.
+    """
+    data = _read(data_dir)
+    if clear:
+        data.pop(_FILES_URL_KEY, None)
+    else:
+        if not isinstance(url, str) or not url.strip():
+            raise ValueError("url must be a non-empty string (or pass clear=True)")
+        data[_FILES_URL_KEY] = url.strip()
+    _write(data, data_dir)
+
+
 # ---------------------------------------------------------------------------
 # Remote server bearer token
 # ---------------------------------------------------------------------------
@@ -681,6 +726,8 @@ __all__ = [
     "set_admin_token",
     "get_registry_url",
     "set_registry_url",
+    "get_files_url",
+    "set_files_url",
     "get_registry_token",
     "set_registry_token",
     "get_managed_by",
--- a/taosmd/ref_fetch.py
+++ b/taosmd/ref_fetch.py
@@ -1,0 +1,153 @@
+"""Resolve and fetch taOS Files-backed refs with hash verification.
+
+A ref uri of the form ``taos://<project-slug>/files/<path>`` is resolved to
+``GET /api/projects/{slug}/files/{path}`` on the configured controller. The
+fetch helper verifies the returned bytes against the ref's ``sha256`` and
+returns the verified bytes or a typed error.
+
+No new server storage is introduced: this is purely a client-side helper.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import os
+import urllib.parse
+
+
+class RefFetchError(Exception):
+    """Base error for ref fetch failures."""
+
+
+class HashMismatchError(RefFetchError):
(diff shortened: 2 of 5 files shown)


jaylfc:
## Review: blocked, 3 blockers. The traversal fix is sound; the token-scope hole is not closed

I closed #238 because it could send our registry bearer token outside the Files scope. So I
reviewed this rebuild by attacking it rather than reading it, and I checked the specific thing
the card exists to fix first.

**The path-traversal fix holds.** Both controls reject, and the accepted cases are neutralised:

```
PROBE: what does resolve_ref_uri accept?
  ACCEPT  CONTROL, benign            -> .../files/docs/spec.md
  reject  CONTROL, literal traversal (dot segment '..')
  reject  CONTROL, single-encoded    (dot segment '..')
  reject  encoded slash in segment   (dot segment '..')
  ACCEPT  double-encoded             -> .../files/%25252e%25252e/etc
  ACCEPT  backslash traversal        -> .../files/..%255c..%255cetc
  ACCEPT  overlong UTF-8             -> .../files/%25c0%25ae%25c0%25ae/etc
```

The three ACCEPTs are safe, but for a reason worth writing down: `quote(path, safe="/")`
re-encodes the `%`, so `%252e` reaches the server as the literal text `%25252e` and decodes once
to `%252e`, never to a dot. The defense is therefore two mechanisms, `_reject_dot_segments` plus
the re-quoting, and only the first is documented. An optimisation that skips re-encoding an
already-encoded path would silently reopen the encoded variants. Please put that in a comment.
26/26 tests pass, and the deleted-symbols gate is green (this is its second live PR).

The problem is that the token-scope defect did not go away. It moved.

---

### Blocker 1: the registry bearer token follows redirects to any host

`_fetcher` builds a `urllib.request.Request` with `Authorization: Bearer <registry_token>` and
calls `urlopen`, which follows 3xx by default. CPython's `HTTPRedirectHandler` copies every
header except content-length and content-type onto the redirected request, so `Authorization`
crosses to the new origin. Reproduced with the fetcher copied verbatim from this branch:

```
CONTROL direct GET to attacker host:
   captured Authorization -> Bearer SECRET-REGISTRY-BEARER-TOKEN

TEST controller 302 -> attacker host:
   bytes returned to caller       -> b'payload-from-attacker-host'
   attacker received Authorization -> Bearer SECRET-REGISTRY-BEARER-TOKEN
```

The ref uri here is entirely well-formed, no dot segments, and `resolve_ref_uri` accepts it. The
traversal fix is not bypassed; it is simply not on this path.

Two things make this worse than it first looks:

- **The sha256 check protects the data, not the credential.** The attacker's bytes fail
  verification and the caller gets a `HashMismatchError`, so the operator sees "fetch failed"
  and nothing else. The token left the building before the hash was ever computed. That is
  fail-quiet on exactly the half that matters.
- It needs the configured controller to emit the redirect, so it is not reachable from a hostile
  ref uri alone. But "the controller can be made to redirect" covers an open redirect on the
  controller and a compromised or misconfigured one, and the whole reason this card exists is
  that the token must not leave the Files scope.

Fix: build an opener with a redirect handler that drops `Authorization` when the redirect target
changes origin, or refuse redirects outright and treat a 3xx as an error. Either is a few lines,
and the test is the probe above.

### Blocker 2: `POST /refs/fetch` has no server route

`RemoteClient.fetch_by_ref` posts to `/refs/fetch`. That path exists nowhere else on the branch:

```
$ git grep -n "refs/fetch" exec/tsk-7xcsh5
  taosmd/remote.py:405:  """POST /refs/fetch: proxy a ref fetch to the remote server."""
  taosmd/remote.py:409:  return await self._run("POST", "/refs/fetch", {"ref": ref, "agent": agent})
```

No `http_server.py` change is in the diff at all. So a remote-configured install calling
`service.fetch_by_ref` takes the `_get_remote` branch and gets a 404 from its own server. The PR
title says "working remote route"; the client half is written and the server half is not. Same
shape as #234's unrouted endpoints and #232's dispatched-but-unwritten handler, which is now
three PRs in a row, so it may be worth a checklist item rather than three separate review notes.

### Blocker 3: the guard and the fetch resolve two different base URLs

`service.fetch_by_ref` computes `files_url = _config.get_files_url(data_dir)` and raises if it is
unset. It then never passes that value on. `ref_fetch.fetch_by_ref` calls `_get_files_url()`
again internally, with **no** `data_dir`. The local variable is dead and the two disagree:

```
CONTROL: guard passes but the fetch resolves elsewhere?
    guard get_files_url(d2)    : https://only-in-datadir.example   (guard PASSES)
    _get_files_url() (no dir)  : RefFetchError: files_url is not configured
```

The HTTP server always passes `data_dir` (see the `_get_remote` docstring, "as the http_server
always does"), so on a server this diverges every time: the guard says configured, the fetch says
not configured. Only the `TAOSMD_FILES_URL` env path makes them agree, which is why the tests do
not catch it.

This also undercuts the third item in the title. The `registry_url` fallback is real, but it
lives in `_get_files_url()` and resolves against the default config location, ignoring the
`data_dir` the caller asked about. A data_dir install with `files_url` set can still fall through
to the default config's `registry_url` and fetch from the registry host, carrying the bearer
token there. Given the card is about token scope, that fallback deserves an explicit decision
rather than arriving as a side effect: pass the resolved `files_url` into
`ref_fetch.fetch_by_ref` as a parameter and delete the second resolution.

---

### Smaller things

- **`agent` is threaded through and never used.** `_fetcher(url, agent)` ignores it; every fetch
  uses the one install-wide registry token. If the intent is per-agent authorisation, it is not
  implemented, and if it is not, the parameter invites the belief that it is.
- **No response size limit.** `resp.read()` is unbounded and the result is base64'd at 1.33x
  before returning, so a large or hostile file is two copies in memory. A `max_bytes` guard costs
  a line.
- **The sha256 comparison is case-sensitive** against a hex string supplied in the ref. An
  uppercase hash fails as a mismatch rather than matching. `.lower()` on both sides.
- `resolve_ref_uri` accepts `....//etc`, which is safe on a server that does not normalise but is
  worth rejecting alongside the dot segments since it costs nothing.

### What I would do with it

The rebuild does the thing it was carded for and does it correctly, which I want to say plainly
because the traversal fix is the part I was most sceptical of. Blocker 1 is the same defect
class that closed #238 reappearing on a path the fix does not cover, and I would not land it
with the token still crossing origins. Blockers 2 and 3 are both "written but not wired", and
both have a one-line reproduction above.

Revision goes back to the lane on tsk-7xcsh5. The test file is well built and each blocker above
converts directly into a failing test, so please add those first.

jaylfc:
**Revision card filed: `tsk-revvkw`.**

jaylfc:
Superseded and closed: **#264 merged as `ac8f714`.**

Same reasoning as #234, and the same hazard: #264 is a carry-forward revision of this PR (`carry forward tsk-7xcsh5`), so nothing auto-closed this one either. Leaving it open means a later merge silently reverts the redirect fix that just landed.

Verified on master after the merge rather than taken on trust:
- `_NoRedirect` at `service.py:375-384` - `redirect_request` returns `None` and the opener is built with it, so the bearer token can no longer follow a 3xx to a second origin. Control: absent on pre-merge master `6de8798`.
- `POST /refs/fetch` now has a real server route at `http_server.py:1020` - that was blocker 2 here.

One known gap carried forward openly: #264's `test_fetcher_uses_no_redirect_opener` is vacuous (a do-nothing `HTTPRedirectHandler` subclass passes it while the token leaks). The fix is proven by probe, not by the suite. Carded as a follow-up rather than left implicit.

The review findings here stand as history. Branch untouched.

PR #243, opened 2026-08-13
deleted-symbols gate: the failure message never tells you the waiver exists

jaylfc:
CARD TITLE (intent, not commit subject): deleted-symbols gate: the failure message never tells you the waiver exists

Autonomous build of board card tsk-gkmvcp.

The waiver trailer is documented only in a workflow comment and the
module docstring, neither of which a contributor reads while CI is red.
Print a ready-to-paste trailer line built from the symbols that actually
failed right in the FAIL message so the escape hatch is discoverable.

Add tests asserting the hint appears on failure and is absent from the
clean output.

Files:
 scripts/check_deleted_symbols.py |  9 +++++++++
 tests/test_deleted_symbols.py    | 25 +++++++++++++++++++++++++
 2 files changed, 34 insertions(+)

### DIFF
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -226,6 +226,15 @@ def main(argv: list[str] | None = None) -> int:
         )
         for v in violations:
             print(f"  - {v.symbol} (added by {v.added_by})")
+
+        trailer_symbols = ", ".join(v.symbol for v in violations)
+        print()
+        print(
+            "If these deletions are intentional, add this trailer to the PR body and "
+            "the gate will re-run automatically:"
+        )
+        print()
+        print(f"    {TRAILER} {trailer_symbols}")
         return 1
 
     print("deleted-symbols-guard: clean")
--- a/tests/test_deleted_symbols.py
+++ b/tests/test_deleted_symbols.py
@@ -206,3 +206,28 @@ class TestDeletedSymbolsIntegration:
         captured = capsys.readouterr()
         assert "DELETED-SYMBOLS FAIL" in captured.out
         assert "taosmd/service.py:a2a_send" in captured.out
+
+    def test_main_prints_waiver_hint_on_failure(self, tmp_path, monkeypatch, capsys):
+        repo = _init_repo(tmp_path)
+        monkeypatch.chdir(repo)
+        monkeypatch.setattr(cds, "REPO_ROOT", repo)
+
+        rc = main(["--base", "HEAD~1"])
+        assert rc == 1
+        captured = capsys.readouterr()
+        assert "Removes-Intentionally:" in captured.out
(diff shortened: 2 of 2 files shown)


jaylfc:
## Approved. The waiver round-trips, proven at the case the tests do not cover

This is my card (tsk-gkmvcp), so I reviewed it against the thing that made me file it: #240
shipped a waiver that was documented only in a YAML comment and a docstring, and taOS shipped a
sibling gate whose documented waiver was unreachable for weeks (bus 2440). A hint that prints the
wrong format would be the same bug wearing a fix.

The two tests here cover one deleted symbol. The interesting case is the one where the hint is
actually constructed, `", ".join(...)` over several violations, so I ran that:

```
=== 1 deleted symbol(s) ===
  rc without waiver         : 1  (expect 1)
  hint the gate prints      : Removes-Intentionally: taosmd/service.py:sym_0
  rc with that hint pasted  : 0  PASS
  CONTROL wrong symbol      : 1  good, still fails

=== 3 deleted symbol(s) ===
  rc without waiver         : 1  (expect 1)
  hint the gate prints      : Removes-Intentionally: taosmd/service.py:sym_0, taosmd/service.py:sym_1, taosmd/service.py:sym_2
  rc with that hint pasted  : 0  PASS
  CONTROL wrong symbol      : 1  good, still fails
  CONTROL partial waiver    : 1  good, still fails for the rest
```

The round trip is real in both shapes: the exact string the gate prints, pasted back verbatim,
turns rc 1 into rc 0. The comma-joined form matches the parser, which splits on `,` at
`check_deleted_symbols.py:145`. Both controls matter, because without them a rc=0 proves nothing:
a wrong symbol still fails, and waiving one of three still fails for the other two, which is the
#240 property (46 down to 45 and still red).

I also checked the claim the new message makes, since it is an assertion about CI rather than
about this script. `.github/workflows/deleted-symbols-gate.yml` triggers on
`types: [opened, synchronize, reopened, edited]`, so "the gate will re-run automatically" on a
body edit is true. That is the exact trap from bus 2440, and this repo is on the right side of it.

22/22 pass. `test`, `deleted-symbols-gate`, CodeRabbit and Gitar all green, 0 commits behind
master. Merging.

### One follow-up, not a blocker

The parser accumulates across lines, so multiple `Removes-Intentionally:` lines work as well as
one comma-joined line. For a large deletion (the incident that motivated #240 named 46 symbols)
the single joined line gets unwieldy. Printing one trailer per line above a threshold would read
better and parses identically.

Worth considering too: telling the reader NOT to use "re-run job" after adding the trailer. Per
@taOS-dev at bus 2440, a re-run replays the stale payload with the old body, so it fails again and
looks like the waiver is broken. The body edit is what re-triggers it. Since this PR exists to
make the waiver discoverable, saying which action works is in scope for a later pass.

PR #244, opened 2026-08-13
PR #235 follow-up: run the revocation check for human principals, and log the human-principal path

jaylfc:
CARD TITLE (intent, not commit subject): PR #235 follow-up: run the revocation check for human principals, and log the human-principal path

Autonomous build of board card tsk-oxhuwe.

Files:
 taosmd/config.py            |  30 +++++++++
 taosmd/http_server.py       |  27 +++++---
 taosmd/registry_auth.py     |  52 +++++++++++++--
 tests/test_registry_auth.py | 158 ++++++++++++++++++++++++++++++++++++++++++++
 4 files changed, 254 insertions(+), 13 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring recognized human principal IDs through environment variables or application configuration.
  * Human principals can authenticate without grant checks.
  * Human identity mismatches now receive an immediate authorization failure.
  * Existing agent authentication and grant validation behavior remains unchanged.

* **Bug Fixes**
  * Improved handling of revoked or mismatched human identities during authentication.

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1575,6 +1575,7 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     try:
                         _registry_verifier.authorize(token, from_)
                     except registry_auth.HumanAuthError as exc:
+                        logger.warning("human principal %r rejected: %s", from_, exc)
                         self._send_json(403, {"error": f"registry auth: {exc}"})
                         return
                     except registry_auth.AuthError as exc:
@@ -1596,6 +1597,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                             warn_reason = str(exc)
                             _reject_status = 403
                             _reject_msg = f"registry auth: {exc}"
+                    else:
+                        logger.info("grants check skipped for human principal %r", from_)
 
                 if warn_reason is not None:
                     enforce = _config.get_a2a_auth_enforce(data_dir)
--- a/taosmd/registry_auth.py
+++ b/taosmd/registry_auth.py
@@ -81,6 +81,8 @@ def authorize_sender(token: str, claimed_from: str, *, public_key: str,
         raise AuthError("token has no 'sub' (canonical_id) claim")
     if sub != claimed_from:
         if human_principal_ids and sub in human_principal_ids:
+            logger.warning("human principal %r sub/from mismatch: sub=%r, from=%r",
+                           sub, sub, claimed_from)
             raise HumanAuthError(
                 f"human token sub {sub!r} does not match from {claimed_from!r}"
             )
@@ -129,6 +131,8 @@ class RegistryVerifier:
         self._clock = clock
         self._expected_iss = expected_iss
         self._human_principal_ids = human_principal_ids or set()
+        logger.info("resolved human principal set: %s",
+                    sorted(self._human_principal_ids))
         self._pubkey: str | None = None
         self._revoked: set[str] = set()
         self._revoked_fetched_at: float | None = None
--- a/tests/test_registry_auth.py
+++ b/tests/test_registry_auth.py
@@ -524,3 +524,28 @@ def test_decode_and_verify_is_single_entry_point_for_human_and_agent():
         registry_auth.decode_and_verify(bad_human, pub_pem)
     with pytest.raises(registry_auth.AuthError):
         registry_auth.decode_and_verify(bad_agent, pub_pem)
+
+
+def test_authorize_sender_logs_human_auth_error(caplog):
+    priv_pem, pub_pem = _keypair()
+    token = _sign(priv_pem, {"sub": "human-1"})
+
+    with pytest.raises(registry_auth.HumanAuthError):
+        registry_auth.authorize_sender(
+            token, "human-2", public_key=pub_pem, revoked=set(),
+            human_principal_ids={"human-1"},
+        )
+    assert any("human principal 'human-1' sub/from mismatch" in r.message
+               for r in caplog.records)
+
+
+def test_registry_verifier_logs_human_principal_set(caplog):
+    import logging
+    with caplog.at_level(logging.INFO, logger="taosmd.registry_auth"):
+        v = registry_auth.RegistryVerifier(
+            pubkey_loader=lambda: "pk",
+            revoked_loader=lambda: set(),
+            human_principal_ids={"human-1", "user-alice"},
+        )
+    assert any("resolved human principal set: ['human-1', 'user-alice']" in r.message
+               for r in caplog.records)


jaylfc:
## The fix is correct and proven. Do not merge it yet: it is built on the wrong base and the merge silently undoes it

This is tsk-oxhuwe, the card that #235 has been waiting on. The fix itself is right, and I proved
it discriminates rather than trusting the test names.

**Red, green, green.** #244's three revocation tests, run against #235's version of
`registry_auth.py` (the one with the skip) and then against this branch's:

```
GREEN control  #244 tests vs #244 registry_auth.py   3 passed
RED            #244 tests vs #235 registry_auth.py   3 failed
                 test_authorize_sender_rejects_revoked_human    Failed: DID NOT RAISE AuthError
                 test_verifier_fetches_revoked_for_human
                 test_registry_verifier_rejects_revoked_human
GREEN restored #244 tests vs #244 registry_auth.py   3 passed
```

The defect is exactly where the card said it was. #235 has:

```python
    if not (human_principal_ids and sub in human_principal_ids):
        if sub in revoked:
            raise AuthError(f"canonical_id {sub!r} is revoked")
```

and this PR makes it unconditional. 37/37 tests pass here, and the docstring explaining the
asymmetry (agents are already withdrawable live via `check_agent_scope`, humans had no path at
all) is the clearest writing on this in the file.

---

### Blocker: this branch is cut from master, not from #235, and merging both restores the hole

`exec/tsk-oxhuwe` branches from master (merge-base `53d147b`, 0 behind) and master has **no**
human-principal handling at all. So this is not a follow-up patch on top of #235; it is a
parallel reimplementation of the same feature, differing from #235 by the revocation guard, two
log lines and the docstring.

I merged them both ways to see what comes out. Both orders conflict in four files, so far so
normal. The problem is *which parts* conflict:

```
=== ORDER: #244 then #235 ===
  conflicts: taosmd/config.py taosmd/http_server.py taosmd/registry_auth.py tests/test_registry_auth.py
  the conflict hunks inside authorize_sender:
    <<<<<<< HEAD
          * ``sub`` must not be in the registry revocation set (this applies to
            every principal, including humans: ...)
    =======
          * for agent principals, ``sub`` must not be in the registry revocation set;
    >>>>>>> origin/exec/tsk-legqtr
  ...and then, with NO conflict marker around it:
    if not (human_principal_ids and sub in human_principal_ids):
        if sub in revoked:
            raise AuthError(f"canonical_id {sub!r} is revoked")
```

**Git flags the docstring that describes the fix and silently takes the code that undoes it.**
Same result in the other order, checked by marker position rather than by eye:

```
=== ORDER B (#235 then #244): is the guard inside a conflict marker? ===
  line  94: *** AUTO-MERGED, no marker ***  if human_principal_ids and sub in human_principal_ids:
  line 104: *** AUTO-MERGED, no marker ***  if not (human_principal_ids and sub in human_principal_ids):
  line 106: *** AUTO-MERGED, no marker ***  raise AuthError(f"canonical_id {sub!r} is revoked")
```

So whoever resolves this sees a prose disagreement, picks a side, and ships an auth hole. The
three tests above would catch it, but only if the resolver runs them, and the test file is itself
one of the conflicts.

Worth noting what does *not* save us here: the deleted-symbols gate (#240) detects deleted
symbols, and nothing is deleted. A conditional getting weaker inside a surviving function is
invisible to it. I am not proposing to extend that gate, but it is a concrete limit of it.

**I was initially wrong about this and it is worth saying why**, because it nearly became the
finding: my first merge test looked like `registry_auth.py` auto-merged with no conflict at all.
It had conflicted, and I had grepped a file containing conflict markers, so the grep printed
#235's side and I read it as the merged result. A conflict region contains both versions, so that
output was identical under both hypotheses. Re-run counting marker positions rather than grepping
for the line, which is where the sharper version of the finding came from.

### Recommended sequence

1. **Rebase this branch onto #235** (`exec/tsk-legqtr`), or onto master once #235 has landed. Its
   diff should then be roughly the guard, the two log lines, the docstring, and the tests, which
   is what a follow-up to #235 should look like.
2. Merge #235, then this.
3. Do not hand-resolve the current conflict instead. If anyone does, the three tests named above
   are the acceptance check and they must be run *after* resolution, not before.

The live risk while #235 sits on master without this is small but should be stated rather than
assumed: with no `human_principal_ids` configured the guard's condition is always true, so
revocation applies to everyone and there is no hole. It opens only once human principals are
actually configured.

### One open question, which decides whether the card's goal is actually met

This makes humans revocable **if the controller publishes them on the revoked feed**. The
docstring is honest that it may not: "If humans are never published on the revocation feed the
check is a no-op; if the controller ever lists one, it works."

That is the right way to write it, but it means the end-to-end property the card exists for
("a leaked or off-boarded human token can be withdrawn") is not established by this PR alone. I
am asking @taOS-dev on the bus whether `/api/agents/revoked` lists human principals. If it does
not, the correct fix is on the controller and this is the client half of it.

### Smaller things

- `_human_principal_ids` is resolved **once** at server construction, while the revoked set
  refreshes every `refresh_interval` (300s). So revoking a human takes effect within 5 minutes,
  but adding or removing one from the human set needs a server restart. Probably fine; worth a
  line in the docstring so nobody discovers it during an incident.
- `is_human` is a plain set membership on the configured ids, so a human who is *not* configured
  falls through to the grants check and is rejected for having no grant. That is the safe
  direction, but it means the human path fails closed on a config omission with an error message
  about grants, which will read as the wrong problem. A distinct message would save someone an
  hour.

jaylfc:
## Update: the controller answer is NO, and it is worse than the docstring assumes

I asked @taOS-dev whether the controller publishes human principals on the revocation feed, since
this PR's own docstring says the check is a no-op if it never lists one. Answered from source at
bus 2475: **no, and it cannot.**

The feed is `AgentRegistryStore.list_revoked()`:

```sql
SELECT canonical_id, revoked_at FROM agent_registry
WHERE revoked_at IS NOT NULL ORDER BY revoked_at
```

`agent_registry` holds AGENT identities; its insert path writes the agent's OWNER as `user_id`,
not a principal row, and `human_principal` appears nowhere in the controller. So a human's `sub`
can never appear in `revoked`. The honest reading is stronger than the docstring's: this is not a
check awaiting data, it is a check whose data source is **structurally incapable** of carrying
the case.

**So the docstring must change before this merges.** It currently says "If humans are never
published on the revocation feed the check is a no-op; if the controller ever lists one, it
works", which reads as a configuration contingency. It should say the feed is agent-only by
construction, and name the controller-side work as the blocker. @taOS-dev owns that half, has
read the feed half but not the human session-revocation path in `auth.py`, and is deliberately
not carding it from half a read. The thing that must not happen is this card closing as though
human token withdrawal now works.

To be clear about what still stands: the fix in this PR is correct and I proved it discriminates.
Removing the skip is right regardless, and an honest "this cannot fire yet, here is why and here
is who owns it" is worth having in the tree.

**One correction to that bus exchange, in my favour and against my own message**: @taOS-dev noted
the revocation path "in your docstring" is wrong, `/api/agents/registry/revoked` rather than
`/api/agents/revoked`. That wrong path was in **my bus message**, not in this PR. Checked both
refs:

```
wrong path anywhere in taosmd/ or tests/ on this branch or master:  NOT PRESENT
CONTROL, correct path present on this branch:                       1 hit (registry_auth.py:26)
```

`REVOKED_PATH = "/api/agents/registry/revoked"` is already correct and has been. **Do not
"fix" it.** My sloppy shorthand, corrected on the bus.

### The two bot findings are both genuine, and they support rebasing

kilo-code-bot flagged two things and I verified both rather than taking them at face value.

**1. Missing `set_human_principal_ids` (it calls this CRITICAL; I would call it moderate, but it
is real).** This branch has the getter and no setter:

```
#244 config.py:  613 get_human_principal_ids            (no setter)
#235 config.py:  613 get_human_principal_ids
                 634 set_human_principal_ids            + exported in __all__
```

So there is no supported way to persist the human set, on a config key that gates authorisation.
Note what this is evidence of: #235 **has** the setter, so this is not a design choice here, it
is a piece lost in the reimplementation. Together with the `tests/test_http_server_trust_
enforcement.py` file (117 lines) that #235 has and this does not, #244 is a **subset** of #235
plus the fix, which is exactly why the rebase in my earlier review is the right move rather than
a stylistic preference. Rebasing recovers both for free.

**2. `human_principal_ids` stored by reference** (`registry_auth.py:137`,
`self._human_principal_ids = human_principal_ids or set()`). Not exploitable today, because
`http_server` passes a freshly built `set(...)` and never mutates it. But this is the set that
decides whether the grants check is skipped, so a defensive copy is a one-word change on a
security-relevant field and worth taking.

jaylfc:
**Update to my earlier docstring finding: the reason has changed, and the fix is now easier to state.**

I flagged above (comment 5284297425) that this PR's docstring is too generous, because the revocation feed cannot carry human principals and the docstring frames the check as pending data that is on its way. That was right, but I described it as a **structural limitation**.

**Jay has now ruled** (2026-08-13, put to him by @taOS-dev, relayed at bus 2499): human credential withdrawal happens in the **session/auth layer**, and the revocation feed is **agent-only by design**. The feed is not being extended, no principal store is being added to `agent_registry`, and `human_principal` is not becoming a controller concept.

So the docstring should not say the check is pending, and it should not say the feed cannot express humans either. Both readings invite a future contributor to close the "gap". It should say the feed is agent-only **by decision**, and that human withdrawal lives elsewhere.

Concretely, the thing to avoid is any wording that leaves a reader thinking this is unfinished. The failure mode is not that the check is wrong today; it is that someone six weeks from now reads "pending" as a TODO and wires human principals into `agent_registry`, which is explicitly the wrong table.

Recorded on master in `docs/specs/a2a-bus-auth-transition.md` at 2151279: Target state item 3 is now a deliberate scope boundary that says plainly "if you are reading this because you noticed humans are missing, that absence is the design", and open question 1 is closed by decision with the investigation kept.

**Unchanged from my review**: the revocation fix in this PR is correct and proven to discriminate (its three tests fail against #235's `registry_auth.py`, pass against its own, pass again restored), and the merge-order hazard still blocks. #244 is cut from master and is a parallel reimplementation of #235, and in either order the docstring describing the fix conflicts while the guard that undoes it auto-merges silently. Rebase onto `exec/tsk-legqtr` first.

jaylfc:
## Rebased, and scoped down to the audit logging only. Read this before the diff

This PR now contains **32 added lines and nothing else**: three log statements and two tests for
them. The unconditional revocation check it was built around has been dropped. Your work was not
wrong, and the reason it is not shipping is worth having in writing.

### Why the revocation change is gone

I asked for it, on #235, and I was the one who carded it. After that, Jay ruled that human
credential withdrawal happens in the **session/auth layer** and that the revocation feed is
**agent-only by design**: humans are not on it and are not going to be. That turns "close the human
revocation gap" from an open gap into a declared non-goal, and #235's skip from an oversight into
deliberate design. Put back to Jay today with both sides laid out; the decision is to keep #235's
semantics and land your logging. So the code was correct for the requirement as it stood when you
wrote it, and the requirement moved underneath it.

Your docstring is also why this got caught properly. It said that if humans are never published the
check is a no-op, *"if the controller ever lists one, it works"*. That honesty is what made the
collision legible instead of me merging two plausible-looking guards together.

### The rebase, and what it nearly did

This branch was cut from master, so it was a parallel reimplementation of #235 rather than a patch
on it. Rebasing it produced conflicts in 4 files. **Every conflict git raised was prose or a log
line. The guard that actually decides the auth outcome merged silently to #235's version**, and the
two branches' tests contradict each other outright:

```
#235  test_authorize_sender_skips_revoked_check_for_human    revoked human ACCEPTED
#244  test_authorize_sender_rejects_revoked_human            revoked human REJECTED
```

So a resolver who took HEAD on the test conflicts would have got a fully green build in which this
PR's fix was dead and only the logging survived, with nothing anywhere reporting a semantic change.
That is the shape to remember: the docstring describing a fix conflicts loudly while the line
undoing it merges quietly.

### Verified before merge

- Full suite on the rebased tree: **1396 passed, 12 skipped, 0 failed** (1394 on master plus your 2).
- Both logging tests proven **red** by deleting the log lines they target, then green when restored.
  Neither is a test that passes without its subject.
- Diff is additive only, 0 deletions, which is the independent check that no behaviour moved.

### Your commit subject was changed

It read *"enforce revocation for human principals and add audit logging"*. It no longer enforces
anything, and a commit claiming a capability the build lacks is worse than a missing one, so it is
now *"add audit logging for human-principal auth decisions"* with the reasoning in the body.

One of your log lines turned out to matter more than as an audit trail: `resolved human principal
set: [...]` makes the revocation-exempt set visible at startup, which is the only place that
exemption currently surfaces at all. That is being carded separately.

jaylfc:
## Merging without `Kilo Code Review`. Recording the decision rather than letting it be inferred

`Kilo Code Review` has been `in_progress` on this PR since 17:04:43Z, 36 minutes at the time of
writing, against a measured band of 8-10 minutes today (#298 completed in 10m, #299 in 8m21s).
It is not an isolated slow run:

```
#297  in_progress since 14:50:50Z   (2h50m)
#244  in_progress since 17:04:43Z   (36m)   <- this PR
#258  in_progress since 17:11:45Z   (29m)
```

Every Kilo run started after 16:58Z is hung, and #297 proves a permanent hang is a real state
rather than a slow one. So this is not a wait with a knowable end.

**I am not lowering a gate the repo enforces.** I checked instead of assuming: `master` has no
required status checks at all (`required_status_checks` is absent from the branch protection
object, and `rulesets` returns `[]`). Kilo has always been advisory here. The bar being applied
is therefore my own, and I am choosing to meet it a different way rather than pretending the
check is a blocker I am entitled to skip because waiting is tedious.

### What stands in its place

- **`test` green, `deleted-symbols-gate` green** on `69a93ba`.
- **Full suite on the exact tree that will land**, not on the branch in isolation. Trial-merged
  `69a93ba` into current `master` (`9d4fe4e`, which has moved since the rebase): clean merge,
  **1397 passed, 12 skipped, 0 failed**. That is the 1396 from the rebase plus the one drift-guard
  test #299 added.
- **The diff against current master is 32 insertions and 0 deletions** across three files. Zero
  deletions is the independent check that no existing behaviour moved, which is the property that
  matters most on a file in the auth path.
- **The two new tests were proven red before being trusted**, by deleting the log lines they
  target and watching them fail, then restoring them.
- No file overlap with any other open PR.

### On the scope of this PR

Recorded on #235 (comment 5317995299) and repeated here so the record is complete: my original
block on #235 required running the revocation check for every principal, and Jay's ruling turned
that into a declared non-goal, with human withdrawal living in the session/auth layer. This PR was
then scoped down to **audit logging only**. It does not change the authorization outcome for any
principal, which is consistent with the 0-deletion diff.

Merging on that basis. If Kilo completes later and raises anything real, it goes to a follow-up
card rather than a revert.

PR #246, opened 2026-08-13
A2A: reject implausible since values on stream/messages (id-vs-timestamp footgun, taOS-dev 1447)

jaylfc:
CARD TITLE (intent, not commit subject): A2A: reject implausible since values on stream/messages (id-vs-timestamp footgun, taOS-dev 1447)

Autonomous build of board card tsk-jpioif.

Files:
 taosmd/http_server.py | 14 ++++++++--
 tests/test_a2a.py     | 71 +++++++++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 83 insertions(+), 2 deletions(-)

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -104,8 +104,8 @@ Endpoints
 ``POST /a2a/send``         ``{"from", "body", "thread"?, "reply_to"?, "refs"?, "blocks"?}`` -> send receipt
                            ``refs``: optional list (<=8) of ``{"kind": doc|report|spec|log, "title", "uri", "sha256"?, "doc_id"?, "version"?, "for"?, "summary"?}``
                            ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
-``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line)
-``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream)
+``GET  /a2a/messages``     ``?thread=&since=<epoch_ts>&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line; ``since`` is an epoch timestamp in seconds, values below 1_000_000_000 are rejected)
+``GET  /a2a/stream``       ``?thread=&since=<epoch_ts>``             -> SSE stream (text/event-stream; ``since`` is an epoch timestamp in seconds, values below 1_000_000_000 are rejected)
 ``GET  /a2a/channels``                                     -> ``{"channels": [...]}``
 ``GET  /a2a/members``      ``?channel=<name>``             -> ``{"members": [...]}``
 ``POST /tasks``            ``{"title", "body"?, "project"?, "assignee"?, "priority"?, "depends_on"?: [...], "created_by"}`` -> task object
@@ -1511,6 +1511,11 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                 since = float(since_raw) if since_raw is not None else None
             except (TypeError, ValueError) as exc:
                 raise _BadRequest("'since' must be a float timestamp") from exc
+            if since is not None and since < 1_000_000_000:
+                raise _BadRequest(
+                    "'since' is an epoch timestamp in seconds; got "
+                    f"{since_raw} which looks like a message id - use the message's ts field, or omit since"
+                )
             try:
                 limit_i = int(limit_raw)
             except (TypeError, ValueError) as exc:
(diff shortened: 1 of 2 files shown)


jaylfc:
Reviewed by running it, not reading it. This is my own card (tsk-jpioif) and the work is **good**: 8/8 of its tests pass, the guard is placed correctly on both handlers (the stream raises before `send_response(200)`, which is the easy thing to get wrong), the docstring table is updated on both rows, and the boundary is exact.

Verified boundary, since an off-by-one here would be invisible:

    since=999999999   -> 400   (below threshold)
    since=1000000000  -> 200   (at threshold)

**BLOCKER: `nan` and `inf` walk straight through the new guard.**

    CASE                      messages        stream
    1444        (control)     400             400
    <valid epoch> (control)   200             200 streaming
    nan                       200 {"messages": []}   200 streaming
    inf                       200 {"messages": []}   200 streaming
    -inf                      400             400

Cause: `float("nan") < 1_000_000_000` is **False** (every comparison with nan is false), and `inf < 1_000_000_000` is False too. So both skip the check, and the downstream filter then matches nothing and returns a cheerful empty window.

That is precisely the defect class this card exists to close. The card is about a `since` value that silently returns the wrong window rather than erroring; `nan` is the purest example of it, because an agent whose cursor arithmetic produced a nan gets a permanently silent feed and a 200 forever. `-inf` being correctly rejected is what proves the guard is running and that only these two evade it.

Fix: reject non-finite values explicitly, e.g. `if not math.isfinite(since): raise _BadRequest(...)` before the magnitude check. Please add `nan` and `inf` to the test list on **both** endpoints; the current tests cover only finite cases, which is why this survived.

**SECOND FINDING: the two endpoints now disagree on the same input, and the silent one is the stream.**

    /a2a/messages?since=abc  -> 400 {"error": "'since' must be a float timestamp"}
    /a2a/stream?since=abc    -> 200, streams from now

`_handle_a2a_stream` still has `except (TypeError, ValueError): last_ts = time.time()`, so unparseable input is swallowed and silently reinterpreted as "now". The new guard cannot catch it, because by then `last_ts` is `time.time()`, which is comfortably above the threshold. So this PR makes `since` strict on `/a2a/messages` while leaving `/a2a/stream` permissive for the garbage case, and a caller cannot tell from the response that its cursor was discarded.

Worth fixing in this PR rather than later: `/a2a/stream` is the realtime path, it is the one an agent polls unattended, and @hermes is building an SSE consumer against it right now. A second silent failure mode in that endpoint is expensive. Suggest raising `_BadRequest` from the except branch instead of defaulting, matching `/a2a/messages` exactly.

**Not blocking, noted only**: values far above the threshold (`1e20`) are accepted and also return empty. Same silent-empty shape, but a plausible far-future timestamp is a much weaker footgun than nan, and an upper bound invites its own bikeshedding. Leaving it out is defensible; say so explicitly if you agree rather than leaving it unaddressed.

**No regression on the intended path**: omitting `since` returns everything, and a recent epoch still filters correctly. Both are covered by the PR's own tests and I re-ran them.

Blocked on the nan/inf gap only. Everything else here is sound, and the boundary work is better than the card asked for.

jaylfc:
**Revision card filed: `tsk-dx6fok`.**

jaylfc:
Closing as **superseded by #253** (`e3ea7c3`), on the same measurement as #269.

**This PR and #269 are the same change set.** Hashing the changed lines (additions and deletions, excluding file headers) of each branch against its merge-base:

```
#246 (exec/tsk-jpioif) 72ccddbf949f281f
#269 (exec/tsk-dx6fok) 72ccddbf949f281f   IDENTICAL
CONTROL #268 (exec/tsk-aildfj) b0297918c3a56c3b   differs, so the hash discriminates
```

So the STEP 0 carry-forward on #269 was faithful - it carried this work forward exactly - and the full analysis I posted on #269 applies here unchanged. In short:

1. **The `since < 1_000_000_000` guard does not catch `nan` or `inf`**, because both comparisons are `False`. Probed live on the branch: `/a2a/messages?since=nan` and `?since=inf` both return **200**. `-inf` is rejected only incidentally (`-inf < 1e9` is `True`), which is what proves the mechanism rather than the symptom.
2. **`/a2a/stream?since=abc` returns 200 here and 400 on master** - the stream handler keeps `except (TypeError, ValueError): last_ts = time.time()`, so a malformed cursor silently becomes "now" and skips history.
3. **Master already does all of it.** `_parse_since` (http_server.py:267) applies `math.isfinite` **and** the 1e9 floor and is wired into both call sites (:1617, :1661) - the exact two sites patched inline here. On master, `nan`/`inf`/`-inf`/`0`/`1444`/`abc` are all 400 and a valid epoch is 200.
4. **The settling check:** this branch test file, lifted onto master unchanged, gives `8 passed`. Every acceptance test here is already green on master **without** this branch code.

This PR also conflicts with master as it stands.

The one thing master genuinely lacks - a `since`-rejection test on `/a2a/stream` - is carried forward as **tsk-q6fjux**, which points the lane at this branch tests as the starting material. Nothing is discarded. Branch untouched.

jaylfc:
Closed as superseded by #253 (e3ea7c3). Identical change set to #269 (verified by hash, with a discriminating control); full measurement in the comment above. Stream-test gap carried to tsk-q6fjux.

PR #248, opened 2026-08-13
deleted-symbols gate: multi-line waiver for large deletions, and say which action re-runs it

jaylfc:
CARD TITLE (intent, not commit subject): deleted-symbols gate: multi-line waiver for large deletions, and say which action re-runs it

Autonomous build of board card tsk-xuuzct.

Files:
 scripts/check_deleted_symbols.py | 10 +++--
 tests/test_deleted_symbols.py    | 83 ++++++++++++++++++++++++++++++++++++++++
 2 files changed, 90 insertions(+), 3 deletions(-)

### DIFF
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -227,14 +227,18 @@ def main(argv: list[str] | None = None) -> int:
         for v in violations:
             print(f"  - {v.symbol} (added by {v.added_by})")
 
-        trailer_symbols = ", ".join(v.symbol for v in violations)
         print()
         print(
             "If these deletions are intentional, add this trailer to the PR body and "
-            "the gate will re-run automatically:"
+            "edit the PR body, do not use Re-run job:"
         )
         print()
-        print(f"    {TRAILER} {trailer_symbols}")
+        if len(violations) > 5:
+            for v in violations:
+                print(f"    {TRAILER} {v.symbol}")
+        else:
+            trailer_symbols = ", ".join(v.symbol for v in violations)
+            print(f"    {TRAILER} {trailer_symbols}")
         return 1
 
     print("deleted-symbols-guard: clean")
--- a/tests/test_deleted_symbols.py
+++ b/tests/test_deleted_symbols.py
@@ -134,6 +134,27 @@ def _init_repo(tmp_path):
     return tmp_path
 
 
+def _init_repo_with_n_symbols(tmp_path, n):
+    """Create a git repo with n top-level defs, all deleted in head."""
+    subprocess.run(["git", "init"], cwd=tmp_path, check=True, capture_output=True)
+    subprocess.run(["git", "config", "user.email", "test@test.com"], cwd=tmp_path, check=True, capture_output=True)
+    subprocess.run(["git", "config", "user.name", "Test"], cwd=tmp_path, check=True, capture_output=True)
+
+    base_file = tmp_path / "taosmd" / "service.py"
+    base_file.parent.mkdir(parents=True, exist_ok=True)
+    lines = "\n".join(f"def s{i}():\n    pass\n" for i in range(n))
+    base_file.write_text(lines)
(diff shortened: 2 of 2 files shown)


jaylfc:
My card (tsk-xuuzct). **The functional half is correct and I proved it by round-tripping rather than reading**, which is the same method that cleared #240 and #243. 27/27 tests pass, and independently:

    n=6, no waiver                    rc 1   (control: the gate can go red)
    printed 6 trailer lines, pasted
      back VERBATIM as the PR body    rc 0   (the hint actually works)
    waive 3 of the 6                  rc 1   (the #240 property still holds)
    n=5 boundary                      1 line, single-line form, round-trips to rc 0

Worth stating for the record, because it is the thing that could have been wrong and was not: the **parser already accepted multi-line trailers** (`_waived_symbols` iterates every line and unions into a set, lines 141-149). So this PR correctly changes only the HINT, which previously offered a form the parser accepted but the message never advertised. That is exactly the shape of the defect #243 fixed, one level down, and the fix is in the right place. The >5 threshold is a sensible split and the tests pin both sides of it.

**One fix before merge, and it is the half the card actually asked for.**

The new sentence reads:

> If these deletions are intentional, add this trailer to the PR body and edit the PR body, do not use Re-run job:

Two problems. It is self-redundant, since adding the trailer to the body IS editing the body, so the instruction appears to ask for two actions when there is one. More importantly it DROPPED the old clause "and the gate will re-run automatically", which was true and was the reassurance that stops someone reaching for Re-run job in the first place. The card asked the message to say *which action re-runs it*; this version says only which action does not.

Suggested replacement:

    If these deletions are intentional, add this trailer to the PR body.
    Editing the body re-runs this gate automatically. Do not use "Re-run job":
    it replays the old body and will fail again.

The final clause is the bus-2440 finding, and it is the reason the whole card exists: taOS shipped a gate whose documented waiver was unreachable for weeks because re-running the job replays the stale payload. Naming that consequence is what makes the warning stick rather than read as arbitrary.

Verified separately and not a defect here: `deleted-symbols-gate.yml` does carry `types: [..., edited]`, so "editing re-runs it automatically" is a true statement and safe to put back in the message.

Nothing else. Fix the wording and this is good to merge.

jaylfc:
**Revision card filed: `tsk-hyqmvc`.**

jaylfc:
Superseding card: **`tsk-zln2uf`** replaces `tsk-hyqmvc`.

Same fix, and the do-not-regress evidence is carried over unchanged. What differs is the mechanism: the new card uses the per-card `BASE: exec/tsk-xuuzct` override that @taOS-dev installed today, and drops the STEP 0 squash entirely. Squashing preserves the work but not the ancestry, so a squashed revision can never contain this branch's head; cutting from the branch makes it an ancestor by construction, and the executor now asserts that at runtime.

This card is also the deliberate first end-to-end trial of that override, chosen because it was unclaimed (so no lane lost work) and because the failure mode it guards against would be invisible here: if the cut silently fell back to master, the entire multi-line waiver work would vanish and the PR would still open clean and green.

jaylfc:
## Closing: this PR is now a REVERT of what already landed.

Its revision, PR #259, was cut from this branch with the `BASE:` override and merged as `8b20e09`. Master therefore already contains this branch's work plus the wording fix. What remains on this branch is only the OLD text, so merging it now would undo the fix:

```
git diff origin/master origin/exec/tsk-xuuzct -- scripts/check_deleted_symbols.py

-            "If these deletions are intentional, edit the PR body to add this trailer. "
-            "Editing the PR body re-runs this gate automatically; Re-run job does not "
-            "work because it replays the stale payload with the old body:"
+            "If these deletions are intentional, add this trailer to the PR body and "
+            "edit the PR body, do not use Re-run job:"
```

Note the branch still reports `1 commit ahead of master`, because #259 was squash-merged and the commit objects did not transfer. A stale branch after a squash-merge looks identical to a branch with genuine unmerged work, which is why this is worth stating rather than leaving for someone to infer from the ahead-count.

Nothing is lost: the functional half (the multi-line waiver) and the improved hint are both on master.

PR #249, opened 2026-08-13
CI gate: RemoteClient call paths must have a matching http_server route

jaylfc:
CARD TITLE (intent, not commit subject): CI gate: RemoteClient call paths must have a matching http_server route

Autonomous build of board card tsk-bwgr26.

Files:
 .github/workflows/remote-routes-gate.yml |  35 +++++
 scripts/check_remote_routes.py           | 206 ++++++++++++++++++++++++++++
 tests/test_remote_routes_gate.py         | 228 +++++++++++++++++++++++++++++++
 3 files changed, 469 insertions(+)

### DIFF
--- a/.github/workflows/remote-routes-gate.yml
+++ b/.github/workflows/remote-routes-gate.yml
@@ -1,0 +1,35 @@
+name: Remote routes gate
+
+# Ensures every RemoteClient call path in taosmd/remote.py has a matching
+# routed literal in taosmd/http_server.py.  A missing route is invisible to
+# review-by-reading and invisible to the local test suite because the tests
+# only exercise the local (non-remote) path.  This gate catches it
+# mechanically before it ships.
+#
+# Paths built by interpolation (f-strings, %s) cannot be extracted by a
+# literal scan.  They are counted as unparseable call sites and reported so
+# a silent drop in coverage is visible.
+#
+# A Missing-Route-Intentionally trailer in the PR body waives named missing
+# routes, making deliberate omissions a conscious, auditable act.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  remote-routes-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    env:
+      PR_BODY: ${{ github.event.pull_request.body }}
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v7
+      - name: Install Python
+        run: uv python install 3.12
+      - name: Install deps
+        run: uv sync --python 3.12
+      - name: Check remote routes
+        run: python scripts/check_remote_routes.py
--- a/scripts/check_remote_routes.py
+++ b/scripts/check_remote_routes.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python3
+"""RemoteClient routes gate.
+
+Extracts every ``self._run("<METHOD>", "<path>"`` literal from
+``taosmd/remote.py`` and requires each path to appear as a routed literal
+in ``taosmd/http_server.py``.  Fails with the method and path named.
+
+Paths built by interpolation (f-strings, ``%s``) cannot be extracted by a
+literal scan.  They are counted as unparseable call sites and reported so
+a silent drop in coverage is visible.
+
+A ``Missing-Route-Intentionally:`` trailer in the PR body waives named
+missing routes, making deliberate omissions a conscious, auditable act.
+"""
+from __future__ import annotations
+
+import argparse
+import ast
+import os
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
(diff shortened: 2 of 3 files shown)


jaylfc:
My card (tsk-bwgr26). I pre-registered the red/green for this gate **before** carding it, so this review is a reproduction rather than a judgement. **The gate itself passes every criterion I set, exactly.**

Ran the PR's script against the three refs I measured when filing the card:

```
origin/master              clean                        rc 0   (expected 19 paths / 0 missing)
origin/exec/tsk-7xcsh5     FAIL, POST /refs/fetch       rc 1   (expected 1 missing, named)
                             at remote.py:409
origin/exec/tsk-lmlx2v     clean                        rc 0   (expected 0 missing)
```

Every one matches. The #233 row is the one that matters most: it is **green despite #233 having a real bug** (the dropped `reader` argument), which is what proves the gate discriminates rather than failing everything in sight. A gate that goes red on all three would have looked equally "working".

Waiver round-trip and its controls, the method that cleared #240 and #243:

```
printed trailer pasted back verbatim   rc 0
wrong path in the trailer              rc 1
empty PR body                          rc 1
```

**And it honours the scope limit I wrote into the card**: interpolated paths are *reported*, not silently skipped. `remote-routes-gate: 3 unparseable call site(s)` on every ref. That line is doing real work, and it is the difference between a known coverage boundary and a silent one.

## BLOCKER: a test that can only pass on the machine that wrote it

`tests/test_remote_routes_gate.py:226`:

```python
calls, _ = _parse_remote_calls(
    Path("/tmp/exec-tsk-bwgr26/taosmd/remote.py").read_text(encoding="utf-8")
)
```

That is an absolute path into **the lane's own build worktree**. It is not in the repo, it is not in the checkout, and it is gone now:

```
$ ls -d /tmp/exec-tsk-bwgr26   -> absent
$ uv run pytest -k twenty_literal
FAILED tests/test_remote_routes_gate.py::TestCurrentRepo::test_at_least_twenty_literal_paths
```

The workflow at `.github/workflows/remote-routes-gate.yml` only runs `python scripts/check_remote_routes.py`, so the gate job stays green and hides this; the main `test` job runs the suite and will not. Fix: read the repo's own `taosmd/remote.py` via `REPO_ROOT`, the way the rest of the file already does.

Worth naming the irony, because it is the same defect class this gate exists to catch: the PR that stops a client half shipping without its server half shipped a test bound to a path that only ever existed on one box.

## Bot findings adjudicated, and four of them are wrong in DIRECTION

Kilo reports 4 CRITICAL, all variants of "unhandled `FileNotFoundError`/`OSError` on `read_text()`" and "unhandled `SyntaxError` in `ast.parse()`". **Do not fold these.** Measured:

```
missing file   -> rc 1
syntax error   -> rc 1
```

The gate already **fails closed**, which is correct and is what this document's own fail-safe rules demand. Catching those exceptions so the script can continue would convert a fail-closed gate into a fail-open one, which is precisely the failure this repo has been hunting all week. The only change worth making is a clearer error message, and even that is cosmetic.

Kilo's line 226 finding is the real one and I have credited it above; it independently matched what I found by running the suite.

**Latent, not blocking**: the waiver splits on commas, so a route literal containing a comma would print a trailer that does not round-trip. No such route exists today. Same shape as the #248 multi-line waiver issue. Worth a comment in the code rather than a fix.

## Residual coverage, stated so nobody over-reads the green

The 3 unparseable sites are all f-string task paths:

```
remote.py:374  self._run("POST", f"/tasks/{task_id}", ...)
remote.py:385  self._run("POST", f"/tasks/{from_id}/edges", ...)
remote.py:398  self._run("POST", f"/tasks/{from_id}/edges/remove", ...)
```

Those are exactly the family **PR #185** touches. So the one open PR most likely to add a task-edge route is in the gate's blind spot. Not a defect in this PR, and the card said so, but it is the concrete reason the reported count must never be allowed to drift upward unnoticed.

Fix the hardcoded path and this merges.

jaylfc:
**Revision card filed: `tsk-hb2y7n`.** This PR stays open; the card's STEP 0 squash-merges the branch so the work carries forward.

Filed late, and the reason is worth recording rather than glossing. When I enumerated the frozen PR set earlier I built the list from my own running notes instead of from a fresh sweep of open PRs against "has unresolved review findings", so this PR was missed while eight others got cards. The hourly freshness pass caught it. That is the same failure I have written up twice today: an accurate local observation generalised into a list I never re-measured.

Mechanics as with the others: `next_card.py:32` excludes any card whose `exec/*` PR is open, and `executor.sh` still has no per-card BASE override, so the revision needs a new card. STEP 0 uses `git merge --squash` because `executor.sh:584` bounces branches containing a merge commit.

jaylfc:
Closing as **superseded by #255**.

The review of this PR asked for the revision; `tsk-hb2y7n` produced it as #255 (same three files, +2 test lines).

Verified before closing: #255 branches off `master` with its own commits and its changed-file set is a **superset** of this PR's, so closing this loses no content. The two would also conflict if both landed.

PR #247, opened 2026-08-13
research-report: log the #184 granite ship in the revision log (finding recorded, ship is not)

jaylfc:
CARD TITLE (intent, not commit subject): research-report: log the #184 granite ship in the revision log (finding recorded, ship is not)

Autonomous build of board card tsk-kb2rgf.

Files:
 docs/research-report.md | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

PR #250, opened 2026-08-13
Measure identity spellings in CHANNEL MEMBERSHIP rows before bus-auth Stage 2

jaylfc:
CARD TITLE (intent, not commit subject): Measure identity spellings in CHANNEL MEMBERSHIP rows before bus-auth Stage 2

Autonomous build of board card tsk-rf5gwb.

Files:
 docs/specs/tsk-rf5gwb-membership-stems.md |  79 +++++++++++
 scripts/measure_membership_stems.py       | 219 ++++++++++++++++++++++++++++++
 tests/test_measure_membership_stems.py    | 114 ++++++++++++++++
 3 files changed, 412 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added a command-line measurement tool for analyzing membership identity stems from available archive data.
  * Reports normalized spellings, timestamp variants, canonical identity twins, and potential stem collisions.
  * Preserves install-specific identity distinctions during reconciliation.

* **Documentation**
  * Added a specification documenting measurement methods, observed spelling variations, and reconciliation findings.

* **Tests**
  * Added coverage for normalization, timestamp handling, identity grouping, and install discriminator separation.

### DIFF
--- a/docs/specs/tsk-rf5gwb-membership-stems.md
+++ b/docs/specs/tsk-rf5gwb-membership-stems.md
@@ -1,0 +1,79 @@
+# tsk-rf5gwb: Membership identity-stem measurement
+
+Status: CLOSED
+Scope: channel membership principals from bus-spool.jsonl (499 sender/channel pairs, 753 raw lines)
+
+## Why this exists
+
+Stage 2's mismatch gate and the membership reconciliation migration both key on
+normalised identity. The `from`-field measurement (last 400 bus messages) is
+scoped to `from` only. Channel membership is a different field and was not
+re-measured under the stem grouping. Carded as tsk-rf5gwb so Stage 2 does not
+assume the `from` conclusion carries over.
+
+## Method
+
+1. Extract every distinct `(sender, channel)` pair from the bus-spool.
+   Lines without an unambiguous `[bus/<channel>] <sender>:` or
+   `<sender>: [AUTO-ACK]` header are excluded so the measurement stays on
+   observed channel membership, not inferred routing.
+2. Compute two stem groupings for each sender:
+   - **without mint stripping**: strip `@`, casefold
+   - **with mint stripping**: strip `@`, casefold, then strip a trailing
+     `-YYYYMMDD-HHMMSS` mint stamp
+3. Do not collapse `@taOS-agent-<install8>` install discriminators: the
+   partial unique index on `(handle) WHERE status='active'` rejects the
+   second insert, so merging two installs would be a measurement bug, not a
+   finding. The mint-strip regex only matches `-YYYYMMDD-HHMMSS`, so install
+   IDs are preserved.
+
+## Measured numbers
+
+| Question | Answer |
+|---|---|
+| Total distinct principals | 14 |
+| Stems with >1 spelling, no mint stripping | 1 |
+| Stems with >1 spelling, with mint stripping | 1 |
+| Canonical entries with bare or @-form twin | 0 |
+| Distinct principals collapsing to one stem (no mint) | 1 |
+
+### Stems carrying more than one spelling
+
+Both groupings see the same single multi-spelling stem:
+
+```
+taosmd-dev: ['@taOSmd-dev', 'taosmd-dev']
+```
+
+No other stem carries more than one spelling.
+
+### Canonical twin check
+
+Zero canonical (mint-stamped) membership entries have a bare or `@`-form twin.
+
+This is the question that decides whether mint-stamp stripping is safe for
+membership. The answer is yes: stripping the mint stamp would unify nothing
+that is actually split, and it adds no collision surface on the measured
+data.
+
+### Collapse check (distinct principals to one stem)
+
+One pair of distinct raw senders collapses to the same stem without mint
+stripping:
+
+```
+taosmd-dev: ['@taOSmd-dev', 'taosmd-dev']
+```
+
+These are two spellings of the same agent. The slug match is safe for
+membership because no two **different agents** share a stem.
+
+## Conclusion
+
+Mint-stamp stripping is safe for membership under the measured scope.
+The `from`-field conclusion carries over: `_normalise_handle` (strip `@`,
+casefold, mint stamp stripped only when explicitly requested) is sufficient
+for Stage 1 and satisfies the install-discriminator constraint.
+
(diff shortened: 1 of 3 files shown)


jaylfc:
## Review: BLOCKED on 3 blockers. The tool is close to right, the data source is not.

This card is a **Stage 2 entry gate** (the bus-auth doc gates Stage 2 on "read-path fix landed and deployed **+ tsk-rf5gwb closed**"), so the standard here is higher than an ordinary card: if this lands with the wrong numbers, Stage 2 begins on a false measurement. 412/0, 3 files, `test` and `deleted-symbols-gate` both green.

The good half is real: `measure()` is sound, the stem functions are correct, and the install-discriminator constraint from bus 2466 is honoured and tested (`@taOS-agent-abc12345-20260813-192605` keeps its install id).

### BLOCKER 1: it measures the wrong population, and every headline number is wrong

`_collect_from_bus_spool` regexes senders out of message **bodies** in `~/.taosmd/bus-spool.jsonl`. Channel membership is already defined on master, by `service.a2a_channels` (service.py:516), whose docstring is explicit: `members` is the sorted set of unique senders observed on that channel, derived from EVENT_A2A archive rows. That endpoint was never consulted.

I ran **this PR's own `measure()`**, unmodified, against the live `/a2a/channels` membership rows:

```
                                   PR (spool)   real membership rows
total distinct principals              14              34
multi-spelling stems, no mint           1               3
multi-spelling stems, with mint         1               4
canonical entries with a twin           0               2
```

77 membership pairs across 17 channels, against the spool's 509. The doc's table is wrong in all five rows.

### BLOCKER 2: the conclusion inverts, and it is the exact conclusion Stage 2 keys on

Same code, right data, the script's own conclusion branch flips itself:

```
CONCLUSION: mint-stamp stripping may NOT be safe for membership.
   hermes-20260608-153000 -> ['hermes', 'hermes-20260727-001415']
   hermes-20260727-001415 -> ['hermes', 'hermes-20260608-153000']
```

`hermes-20260608-153000` and `hermes-20260727-001415` both stem to `hermes`, which also collides with the bare `hermes`. Those are **two distinct installs merged into one identity**, which is precisely what @taOS-dev's install-discriminator constraint forbids, and the reconciliation migration keys on normalised identity.

**Control, in the same run**: `@taOS-agent-abc12345` and `@taOS-agent-def67890` do **not** collapse. So `stem_with_mint` is not merging everything and this is a specific finding about the mint-stamp family, not a broken function.

The doc currently states the opposite ("mint-stamp stripping is safe for membership", "the `from`-field conclusion carries over"). Carrying the `from` conclusion over is the single assumption this card exists to prevent.

### BLOCKER 3: the fallback manufactures a confident wrong answer, silently

Run the documented invocation against a real data dir with a real archive:

```
$ uv run --extra dev python scripts/measure_membership_stems.py --data-dir ~/.taosmd
No EVENT_A2A rows found in /home/jay/.taosmd, falling back to .../bus-spool.jsonl   <- stderr
Scope: bus-spool.jsonl (509 sender/channel pairs)                                    <- stdout
CONCLUSION: mint-stamp stripping is safe for membership.                             <- stdout
```

The warning goes to stderr and the verdict goes to stdout, so a captured report is indistinguishable from a successful archive measurement. That is how the doc got written. Fix: do not fall back at all, or exit non-zero, and put the scope inside the conclusion rather than only in the header.

`_collect_from_archive` is nearly correct already (it reads `from` plus `thread`/`app_id`, which is `a2a_channels`' own derivation). It needs pointing at the Pi's `/opt/taosmd-data`, or simply fetching `/a2a/channels`.

### Defects

1. **Not reproducible.** The doc says 499 pairs, I measured 509 from the same command the same day, because the source is a local mutable file that grows and that nobody else has. A Stage 2 entry gate has to be reproducible by @taOS-dev.
2. **The channel dimension is discarded.** `measure()` opens with `{p for p, _ in pairs}` and drops the channel, so a card about *channel* membership cannot express a per-channel result. On real data, 4 channels carry more than one hermes spelling (`build` carries all three).
3. **`Status: CLOSED` in the doc header.** The card closes when the gate is satisfied, which is a separate judgement from the PR that proposes it.
4. **No collector has a test.** All 8 tests are synthetic input to `measure()`; neither collector is exercised, nor is the fallback. There is also no test for two canonicals sharing one stem, which is the case the real data actually contains.
5. **`is_install_discriminator` is defined and never called.**

### What I want on the revision

Re-run against real membership rows (`/a2a/channels`, or the archive on the Pi's data dir), regenerate the table from that output, and let the conclusion follow the data rather than the `from` result. If mint stripping is unsafe for membership, that is a genuinely useful finding for Stage 2 and the card has done its job.

### Correction to my own earlier claim, so the revision is not held to it

I recorded that hermes appears in membership under **all three** spelling families. Measured: there is no `@hermes` on any channel. It is **two** families (bare and canonical) with **two distinct canonical ids**. The "two canonical ids" half stands and is the load-bearing part; "three families" was mine and was wrong.

jaylfc:
**Revision card filed: `tsk-aildfj`.**

Not because this PR is being abandoned. The work in it is worth carrying forward, and the card says so explicitly. The reason it needs a *new* card rather than a revision of `tsk-rf5gwb` is mechanical: `next_card.py:32` excludes any card whose `exec/*` PR is open, so `tsk-rf5gwb` is unclaimable while #250 exists, and `executor.sh` still sets `BASE` per repo (`master` for taosmd) with no per-card override, so a replacement card would otherwise rebuild from master and discard this branch. I re-checked both in source today rather than assuming the tooling had landed.

The card opens with a required STEP 0 that squash-merges `exec/tsk-rf5gwb` (squash, not merge, because `executor.sh:584` bounces any card whose branch contains a merge commit), and its acceptance line requires `git rev-list --merges` to be empty before push.

This PR stays open. It is the thing the revision builds on.

jaylfc:

- The revision card **`tsk-hsph7e`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-rf5gwb` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-rf5gwb` recovers the work.
- The originating card `tsk-rf5gwb` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #251, opened 2026-08-13
Tests that need a live embed backend must SKIP, not FAIL, so the CI deselect list can be deleted

jaylfc:
CARD TITLE (intent, not commit subject): Tests that need a live embed backend must SKIP, not FAIL, so the CI deselect list can be deleted

Autonomous build of board card tsk-hfy6dg.

Files:
 .github/workflows/ci.yml                | 19 +-----------
 tests/conftest.py                       | 54 +++++++++++++++++++++++++++++++++
 tests/test_catalog_pipeline.py          |  2 +-
 tests/test_hermes_audit_batch2.py       |  6 ++--
 tests/test_http_server_registry_auth.py |  2 +-
 tests/test_project_storage.py           |  8 ++---
 tests/test_reindex.py                   |  2 +-
 7 files changed, 65 insertions(+), 28 deletions(-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,21 +18,4 @@ jobs:
       - name: Compile check (two merged branches previously shipped files that do not parse)
         run: uv run python -m compileall -q taosmd/
       - name: Tests
-        # The --deselect list below is the set of tests that require a live
-        # embed backend (ONNX model or qmd service) and FAIL rather than skip
-        # when none is configured. They still run in full on a configured dev
-        # box. Tracked to get proper skip-if-no-backend guards; remove the
-        # deselects when that lands. Do NOT grow this list to hide new
-        # failures.
-        run: >
-          uv run pytest tests/ -q
-          --deselect tests/test_catalog_pipeline.py::TestCrystallizeAgentScoping::test_crystallize_stage_passes_agent_name_to_lookup_date
-          --deselect tests/test_hermes_audit_batch2.py::test_bm25_cache_is_populated_after_search
-          --deselect tests/test_hermes_audit_batch2.py::test_bm25_cache_invalidated_on_add
-          --deselect tests/test_hermes_audit_batch2.py::test_bm25_cache_not_rebuilt_on_second_query
-          --deselect tests/test_http_server_registry_auth.py::test_search_token_project_id_scopes_results
-          --deselect tests/test_project_storage.py::TestIngestProject::test_ingest_different_agents_same_project
-          --deselect tests/test_project_storage.py::TestSearchProject::test_search_scoped_to_project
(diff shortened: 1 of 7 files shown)


jaylfc:
## Approved. The card's goal is met and I measured it rather than reading it.

My card (tsk-hfy6dg). The claim to test is "the deselect list can be deleted", so that is what I ran.

**Acceptance, on a box with NO embed backend** (verified: no `model.onnx` at either candidate path, and `localhost:7832` unreachable):

```
full suite, no --deselect flags:   1301 passed, 12 skipped, 0 failed  (137s)
```

The ten formerly-deselected tests skip instead of failing, and the CI workflow's entire deselect block plus its "do not grow this list" comment are gone. That was the point of the card.

### The check that actually matters, because "10 skipped" is an ambiguous result

Ten skips are equally consistent with "the guard works" and "the guard always skips", and the second would be a silent loss of ten tests' coverage while looking greener than before. So I ran both halves:

```
A. no backend (real state of this box)      -> ssssssssss   10 skipped
B. TAOSMD_ONNX_PATH -> a dir with model.onnx ->  2 failed   (tests RAN)
C. same two tests, no backend                ->  ss          2 skipped
```

B is the one that makes A meaningful: with the detector satisfied the tests execute rather than skip, so the skip is conditional on the backend and not unconditional.

**Coverage did not silently widen either.** Exactly 10 call sites request the fixture, distributed 1/3/1/4/1 across `test_catalog_pipeline` / `test_hermes_audit_batch2` / `test_http_server_registry_auth` / `test_project_storage` / `test_reindex`, which matches the deselect list file for file. No test that used to run is now skipped.

The skip message is also the right kind: it names both ways to get a backend (`scripts/setup.sh` or `TAOSMD_ONNX_PATH`, or `qmd serve`) rather than just reporting absence.

### Defect, not blocking, and carded as follow-up

`_has_onnx_model()` tests only that the file EXISTS, never that it loads. My step B above is the proof: an EMPTY `model.onnx` satisfied the detector, so the tests ran and FAILED rather than skipping, which is the exact failure mode this card exists to remove, reappearing in a narrower case.

An empty file is artificial, but a **truncated one is not**: an interrupted `scripts/setup.sh` leaves precisely that, and the user then sees two confusing failures instead of a skip that tells them to re-run setup. Worth noting the PR is still a strict improvement here, because before it those tests failed unconditionally.

Suggested fix in the follow-up card: treat an embed failure at fixture time as a skip too, so the guard keys on "can this backend actually embed" rather than "is there a file".

Minor, same card: `_has_qmd_service()` is a function-scoped fixture, so it re-probes per test. Connection-refused returns instantly (the 10 skips took 1.6s total), but on a network where the port is filtered rather than refused each probe waits the full 2s timeout. Session-scoping or caching the answer costs nothing.

Merging.

PR #253, opened 2026-08-13
Revise PR #234: unrouted endpoints, inf/nan through _parse_since, zero tests

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #234: unrouted endpoints, inf/nan through _parse_since, zero tests

Autonomous build of board card tsk-3fuapf.

Files:
 taosmd/http_server.py     | 150 ++++++++++++++++++++++++++++++++++++------
 taosmd/remote.py          |  19 ++++++
 taosmd/service.py         | 164 ++++++++++++++++++++++++++++++++++++++++++++++
 tests/test_http_server.py |  58 ++++++++++++++++
 4 files changed, 372 insertions(+), 19 deletions(-)

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -102,10 +102,12 @@ Endpoints
 ``GET  /pending?agent=``                                   -> ``{"pending": [...]}``
 ``POST /pending/resolve``  ``{"id", "decision", "note"?}`` -> resolve result
 ``POST /a2a/send``         ``{"from", "body", "thread"?, "reply_to"?, "refs"?, "blocks"?}`` -> send receipt
-                           ``refs``: optional list (<=8) of ``{"kind": doc|report|spec|log, "title", "uri", "sha256"?, "doc_id"?, "version"?, "for"?, "summary"?}``
-                           ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
-``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line)
-``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream)
+                            ``refs``: optional list (<=8) of ``{"kind": doc|report|spec|log, "title", "uri", "sha256"?, "doc_id"?, "version"?, "for"?, "summary"?}``
+                            ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
+``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line; ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400)
+``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream); ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400
+``GET  /a2a/threads``      ``?principal=``                  -> ``{"threads": [...]}`` thread list for the principal, ordered by latest activity desc; each entry has ``thread``, ``kind``, ``participants``, ``last_message: {id, ts, from, body_preview}`` (see notes below)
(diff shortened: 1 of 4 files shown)


jaylfc:
## All three blockers on #234 are fixed, each verified by running. Four minor defects, none blocking.

Reviewed on head `ed75622`. 63 tests pass in `tests/test_http_server.py`.

### Blocker 1, unrouted endpoints: FIXED

#234 shipped client methods with no server route. Both new `RemoteClient` methods here have a
live route, checked by hitting a running server rather than by reading the dispatcher:

```
GET /a2a/threads                     -> 200 {"threads": []}
GET /a2a/threads/build/messages      -> 200 {"thread": "build", "messages": []}
```

Both `service.a2a_threads` and `service.a2a_thread_messages` also carry the `_get_remote`
branch, so a remote-configured install does not silently answer from the local archive. That was
the #230 defect class and it is not repeated here.

### Blocker 2, `inf`/`nan` through `_parse_since`: FIXED, red/green

The three new guard tests pass on this branch. Swapping in master's `taosmd/http_server.py` and
re-running the same three tests turns them red, so they discriminate rather than asserting
something trivially true:

```
GREEN (this branch): 3 passed
RED   (master's http_server.py, same tests): 3 failed
      test_a2a_since_nan_rejected - assert 200 == 400
      test_a2a_since_inf_rejected - assert 200 == 400
      test_a2a_since_negative_inf_rejected - assert 200 == 400
```

Better than the tests claim: the guard also covers `/a2a/stream`, which no test exercises.
Measured live on both endpoints, with a valid timestamp as the control:

```
              /a2a/messages   /a2a/stream
since=nan         400             400
since=inf         400             400
since=-inf        400             400
since=12345       400             400
since=1786000000  200             200
```

`_BadRequest` is raised before the SSE headers go out and the outer handler converts it to a
400, so the stream endpoint does not half-open a connection and then fail. Worth a test, since
nothing currently pins that ordering.

### Blocker 3, zero tests: FIXED

Seven new tests covering both new endpoints and the three guard cases.

### The two documented limit claims are true, and I had to seed 250 messages to show it

At 6 messages, `limit=500` returning 6 is what both "clamped to 200" and "not clamped at all"
predict, so the first probe proved nothing. With 250 seeded:

```
no limit  -> 50     (default 50, as documented)
limit=200 -> 200
limit=201 -> 200    (clamped)
limit=500 -> 200
limit=1e6 -> 200
CONTROL /a2a/messages?limit=1000 -> 250   (all 250 really are stored)
```

### Defects (minor)

1. **The limit floor is not validated while the ceiling is.** `limit=0` and `limit=-1` both
   return 200 with exactly **1** message. Measured against a seeded thread where `limit=2`
   returns 2 and `limit=3` returns 3, so the probe discriminates. A client asking for zero
   messages gets one; a negative limit is silently coerced. Either reject with 400 (matching
   how `limit=1e20` is already handled) or clamp to the default, but not silently to 1.
2. **The `thread name required` 404 branch is unreachable.** Line 967 does
   `path = parts.path.rstrip("/") or "/"`, so `/a2a/threads/` normalises to `/a2a/threads` and
   is served by the thread-list handler: measured, it returns `200 {"threads": []}`. Nothing can
   reach `startswith("/a2a/threads/")` with an empty remainder, so `if not thread` is dead code.
3. **`/a2a/threads/{thread}` without the `/messages` suffix silently serves messages.**
   `GET /a2a/threads/build` returns `{"thread": "build", "messages": []}`. That undocumented
   alias is fine as a convenience but it should be documented or removed, not left implicit.
4. **The stream guard has no test**, as noted above.

### Merge-order hazard with #265, please read before either lands

**#265 (tsk-dx6fok) rewrites this same code path** and the two conflict in
`taosmd/http_server.py`. I merged both onto master in a scratch tree and checked the conflict by
marker position rather than grepping the file:

```
<<<<<<< HEAD                                  (this PR)
            since = _parse_since(since_raw)
=======                                       (#265)
            try:
                since = float(since_raw) if since_raw is not None else None
            ...
            if since is not None and not math.isfinite(since):
                raise _BadRequest("'since' must be a finite timestamp")
>>>>>>>
```

**Take this PR's side.** `_parse_since` here does the finite check **and** the epoch floor;
#265's side does the finite check only, and I have measured that #265 reverts the epoch floor to
master's behaviour (`since=12345` returns 200 on #265, 400 here). Details on #265.

**Recommendation: land this first.** It is the stronger implementation of the shared guard and
it makes #265's remaining scope small.

PR #255, opened 2026-08-13
Revise PR #249: the gate's own test reads the lane's build worktree, which no longer exists

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #249: the gate's own test reads the lane's build worktree, which no longer exists

Autonomous build of board card tsk-hb2y7n.

The test at line 226 in tests/test_remote_routes_gate.py read
Path("/tmp/exec-tsk-bwgr26/taosmd/remote.py"), the lane's own build
worktree, which no longer exists. Read the repo's own taosmd/remote.py
via REPO_ROOT instead so the test passes in any environment, matching
the behaviour the gate is meant to enforce.

Carry forward the tsk-bwgr26 work via squash merge.

Files:
 .github/workflows/remote-routes-gate.yml |  35 +++++
 scripts/check_remote_routes.py           | 206 +++++++++++++++++++++++++++
 tests/test_remote_routes_gate.py         | 230 +++++++++++++++++++++++++++++++
 3 files changed, 471 insertions(+)

### DIFF
--- a/.github/workflows/remote-routes-gate.yml
+++ b/.github/workflows/remote-routes-gate.yml
@@ -1,0 +1,35 @@
+name: Remote routes gate
+
+# Ensures every RemoteClient call path in taosmd/remote.py has a matching
+# routed literal in taosmd/http_server.py.  A missing route is invisible to
+# review-by-reading and invisible to the local test suite because the tests
+# only exercise the local (non-remote) path.  This gate catches it
+# mechanically before it ships.
+#
+# Paths built by interpolation (f-strings, %s) cannot be extracted by a
+# literal scan.  They are counted as unparseable call sites and reported so
+# a silent drop in coverage is visible.
+#
+# A Missing-Route-Intentionally trailer in the PR body waives named missing
+# routes, making deliberate omissions a conscious, auditable act.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  remote-routes-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    env:
+      PR_BODY: ${{ github.event.pull_request.body }}
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v7
+      - name: Install Python
+        run: uv python install 3.12
+      - name: Install deps
+        run: uv sync --python 3.12
+      - name: Check remote routes
+        run: python scripts/check_remote_routes.py
--- a/scripts/check_remote_routes.py
+++ b/scripts/check_remote_routes.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python3
+"""RemoteClient routes gate.
+
+Extracts every ``self._run("<METHOD>", "<path>"`` literal from
+``taosmd/remote.py`` and requires each path to appear as a routed literal
+in ``taosmd/http_server.py``.  Fails with the method and path named.
+
+Paths built by interpolation (f-strings, ``%s``) cannot be extracted by a
+literal scan.  They are counted as unparseable call sites and reported so
+a silent drop in coverage is visible.
+
+A ``Missing-Route-Intentionally:`` trailer in the PR body waives named
+missing routes, making deliberate omissions a conscious, auditable act.
+"""
+from __future__ import annotations
+
+import argparse
+import ast
+import os
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DEFAULT_REMOTE = REPO_ROOT / "taosmd" / "remote.py"
+DEFAULT_HTTP = REPO_ROOT / "taosmd" / "http_server.py"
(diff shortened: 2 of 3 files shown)


jaylfc:
## Blocker fixed, and the gate discriminates on real refs. Approving.

Reviewed on head `dbacd4e`. 16 tests pass.

### The blocker is gone

#249's test read the lane's own build worktree, a path that stopped existing the moment the
lane finished:

```
base exec/tsk-bwgr26, line 226:
    Path("/tmp/exec-tsk-bwgr26/taosmd/remote.py").read_text(encoding="utf-8")

this revision: no /tmp, /home, worktree or exec/tsk path anywhere in the test file
```

### The gate actually catches the thing it was carded for

A gate is only worth having if it can go red on a real defect and green on everything else, so I
ran it against five refs rather than trusting the tests:

```
MASTER      rc=0   green
#242-base   rc=1   RED: "1 RemoteClient call path(s) have no matching routed literal"
                        Missing-Route-Intentionally: POST /refs/fetch
#264        rc=0   green   (the same PR after the route was added)
#233        rc=0   green
#263        rc=0   green
```

That is the discrimination I wanted from tsk-bwgr26. It names the exact defect that closed #238
and forced the #242 rebuild, and the #242-base to #264 pair is the cleanest possible red/green:
same card, same client method, route absent then present. It does not fail everything, which was
the risk with a gate like this.

### The two scope limits from the card are honoured

1. **Interpolated paths are reported, not silently skipped.** Every run prints
   `N unparseable call site(s) (interpolated paths not checked)`, and the count moves with the
   ref (3 on most, 4 on #263). A drop in coverage is visible instead of reading as a clean pass.
   The module docstring says so as well.
2. **It catches "no route", not "wrong arguments".** Still true, and still worth stating: this
   gate is green on #233's dropped-`reader` blocker. Nobody should read a green here as "the
   remote path works".

### The waiver is reachable and discoverable

Both halves of the bus-2440 trap are avoided:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, edited]
```

`edited` is present, so adding the trailer re-runs the gate rather than leaving a waiver that
can never fire. And the trailer is printed **in the failure output itself**
(`Missing-Route-Intentionally: POST /refs/fetch`, pre-filled with the offending path), which is
the #243 lesson applied: the escape hatch is discoverable from CI redness alone, not from a YAML
comment.

No blockers. Merge it.

jaylfc:
## Withdrawing my approval: this now breaks master, and the test that breaks is brittle by design

I approved this on 2026-08-14 against head `dbacd4e`, which is still the head. That approval was correct at the time and is wrong now, so I am retracting it rather than merging on it.

**The PR's own test fails on the tree that would actually land.**

I trial-merged `exec/tsk-hb2y7n` into current master and ran the full suite:

```
FAILED tests/test_remote_routes_gate.py::TestCurrentRepo::test_master_is_clean
>       assert unparseable == 3
E       assert 4 == 3

1 failed, 1328 passed, 14 skipped in 139.03s
```

Merge tree clean, working tree clean, all three gate files present on it.

**Control, so this is not an artifact of my trial merge:** the same single test on `exec/tsk-hb2y7n` alone, with no master merged in, passes. It only fails once current master is in the tree. So this is master drift, not a broken branch.

All five checks on this PR are green, including `test`. They ran on 2026-08-14. That green is stale evidence, and it is the whole reason I ran the suite on the merged tree instead of trusting it.

### What drifted

`test_master_is_clean` hard-codes the number of unparseable call sites in the live repo. There were three on 2026-08-14. There are four now:

```
remote.py:277  f"/a2a/threads/{thread}/messages"    <- new since 2026-08-14
remote.py:393  f"/tasks/{task_id}"
remote.py:404  f"/tasks/{from_id}/edges"
remote.py:417  f"/tasks/{from_id}/edges/remove"
```

The A2A chat read work landed the first one after this branch was written.

### The real problem is the shape of the assertion, not the number

Bumping `3` to `4` fixes today and reintroduces the same failure the next time anyone adds an interpolated remote path. That is a recurring tax on unrelated PRs, paid by whoever merges next, for a number that carries no invariant.

The actual invariant this gate exists to enforce is the one on the line above it:

```python
assert missing == [], f"Expected clean but got missing: {missing}"
```

That one is worth asserting against the live repo. The unparseable count is diagnostic, not a contract. Either report it without asserting an exact value, or assert it against a fixture with known content so the number is pinned to something that does not move under you.

I would rather have that fixed now than watch it bounce a future PR that has nothing to do with remote routes.

### Not in dispute

The gate itself is sound and I re-verified it on the merged tree: `scripts/check_remote_routes.py` exits 0 and reports `remote-routes-gate: clean` against current master plus this branch. So merging it would not start blocking unrelated work, and the 471 lines here are worth having. The problem is confined to one assertion in one test.

---

Closing in favour of the revision card `tsk-vw6gq7`, for the same reason as #297 and #298: a blocked PR has no revision path here, and closing frees the throttle slot and clears the one-PR-per-task guard. The card branches from `exec/tsk-hb2y7n`, so all 471 lines survive untouched. **The branch is deliberately NOT deleted.** Please leave it in place.

PR #262, opened 2026-08-13
Revise PR #242: bearer token still crosses origins on redirect, plus a missing route

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #242: bearer token still crosses origins on redirect, plus a missing route

Autonomous build of board card tsk-revvkw.

Files:
 taosmd/config.py        |  47 +++++++++
 taosmd/ref_fetch.py     | 153 ++++++++++++++++++++++++++++++
 taosmd/remote.py        |   8 ++
 taosmd/service.py       |  62 +++++++++++-
 tests/test_ref_fetch.py | 247 ++++++++++++++++++++++++++++++++++++++++++++++++
 5 files changed, 516 insertions(+), 1 deletion(-)

### DIFF
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -42,6 +42,8 @@ _REGISTRY_URL_KEY = "registry_url"
 # Key under which the registry auth token (taOS local/admin token) is stored.
 # Used to poll the auth-gated registry revoked feed; the pubkey feed is public.
 _REGISTRY_TOKEN_KEY = "registry_token"
+# Key under which the taOS Files base URL is stored.
+_FILES_URL_KEY = "files_url"
 # Who manages this taosmd instance: "standalone" (default) or "taos".
 _MANAGED_BY_KEY = "managed_by"
 # Override: serve the web dashboard even when managed_by=taos.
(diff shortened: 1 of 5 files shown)


jaylfc:
## BLOCKED and closing. Byte-identical to the branch it revises.

```
git diff --stat origin/exec/tsk-7xcsh5 origin/exec/tsk-revvkw -- <files the original touched>   ->   empty
```

**Control**: the same command on `exec/tsk-xuuzct` vs `exec/tsk-zln2uf` reports a real change, so the empty result is a measurement rather than a broken command.

Fourth no-op revision today. STEP 0 carried the work forward and nothing after it ran, so all three blockers from my review of #242 persist: the bearer token still crosses origins on redirect, `POST /refs/fetch` still has no server route, and the guard and fetch still resolve different base URLs.

Closing rather than leaving it open, because an open `exec/<card>` PR makes that card unclaimable (`next_card.py:32`), so this PR would freeze the very card that needs redoing. Nothing is discarded: the content still exists on `exec/tsk-7xcsh5`, which has its own open PR. `tsk-revvkw` stays open and becomes dispatchable again, and I have added the non-empty-diff acceptance note to it.

PR #252, opened 2026-08-13
Revise PR #232: receipt identity is forgeable, and a routed admin endpoint has no handler

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #232: receipt identity is forgeable, and a routed admin endpoint has no handler

Autonomous build of board card tsk-qo6tpb.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 taosmd/api.py         |   4 ++
 taosmd/http_server.py | 110 ++++++++++++++++++++++++++++++++++-
 taosmd/migrations.py  |  18 ++++++
 taosmd/receipts.py    | 158 ++++++++++++++++++++++++++++++++++++++++++++++++++
 taosmd/service.py     |  98 +++++++++++++++++++++++++++++++
 5 files changed, 387 insertions(+), 1 deletion(-)

----
## Summary by Gitar

- **Receipt Store & Migrations:**
    - Added `ReceiptStore` in `taosmd/receipts.py` to manage message delivery and read receipts with SQLite
    - Added `a2a_receipts` database migration and mapping in `taosmd/migrations.py` and `taosmd/api.py`
- **API & HTTP Endpoints:**
    - Added service wrappers for receipt operations in `taosmd/service.py`
    - Added HTTP endpoints for managing and querying receipts, including admin prune route in `taosmd/http_server.py`

<sub>This will update automatically on new commits.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added A2A message read receipts, including delivery and seen timestamps.
  * Added endpoints to record, retrieve, and manage message receipts.
  * Authenticated message subscribers now automatically record successful deliveries.
  * Added administrative support for pruning older receipt records.

PR #263, opened 2026-08-13
Revise PR #232: receipt identity is forgeable, and a routed admin endpoint has no handler

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #232: receipt identity is forgeable, and a routed admin endpoint has no handler

Autonomous build of board card tsk-qo6tpb.

Files:
 taosmd/receipts.py                      | 160 ++++++++++++++++++++++++++
 taosmd/remote.py                        |  35 ++++++
 taosmd/service.py                       | 111 +++++++++++++++++-
 tests/test_admin_surface.py             |  10 ++
 tests/test_http_server_registry_auth.py |  32 ++++++
 tests/test_receipts_store.py            |  43 +++++++
 tests/test_remote.py                    |  53 +++++++++
 11 files changed, 671 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added A2A delivery and read receipts for tracking message status by agent.
  * Added APIs to record, retrieve, and prune receipts.
  * Added authenticated receipt acknowledgment and automatic delivery tracking for SSE subscribers.
  * Added administrative receipt cleanup with configurable retention and protected access.
* **Bug Fixes**
  * Receipt acknowledgments now reject invalid or unknown authentication tokens.
* **Tests**
  * Added coverage for receipt storage, remote operations, authorization, and administrative cleanup.

### DIFF
--- a/taosmd/admin.py
+++ b/taosmd/admin.py
@@ -419,6 +419,21 @@ async def a2a_admin_supersede_message(
     return {"superseded": True, "id": msg_id}
 
 
+async def a2a_admin_prune_receipts(
+    older_than_ts: float,
+    *,
+    data_dir: Path | str,
+    stores: dict,
+) -> dict:
+    """Prune receipts older than ``older_than_ts`` (by delivered_at).
+
+    Returns ``{"pruned": int}``.
+    """
+    receipt_store = stores["receipts"]
+    n = await receipt_store.prune(older_than_ts)
+    return {"pruned": n}
+
+
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
--- a/taosmd/api.py
+++ b/taosmd/api.py
@@ -144,12 +144,16 @@ async def _ensure_stores(data_dir=None) -> dict:
         from taosmd.claims.store import ClaimStore  # noqa: PLC0415
         claims = ClaimStore(db_path=str(path / "claims.db"))
         await claims.init()
+        from taosmd.receipts import ReceiptStore  # noqa: PLC0415
+        receipts = ReceiptStore(db_path=str(path / "a2a-receipts.db"))
+        await receipts.init()
 
         stores = {
             "archive": archive,
             "vector": vmem,
             "kg": kg,
             "claims": claims,
+            "receipts": receipts,
             "data_dir": str(path),
         }
         _stores_cache[resolved] = stores
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -108,6 +108,9 @@ Endpoints
 ``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream)
 ``GET  /a2a/channels``                                     -> ``{"channels": [...]}``
 ``GET  /a2a/members``      ``?channel=<name>``             -> ``{"members": [...]}``
+``PATCH /a2a/receipts``    ``{"message_id"}``              -> mark message seen by authenticated agent
+``GET  /a2a/messages/{id}/receipts``                       -> ``{"delivered": [...], "read": [...]}``
+``GET  /a2a/receipts``     ``?message_id=&agent=``         -> ``{"delivered_at", "seen_at"}`` or 404
 ``POST /tasks``            ``{"title", "body"?, "project"?, "assignee"?, "priority"?, "depends_on"?: [...], "created_by"}`` -> task object
 ``GET  /tasks``            ``?status=&project=&assignee=&limit=``  -> ``{"tasks": [...]}``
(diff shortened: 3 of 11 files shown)


jaylfc:
## All five blockers from #232 are fixed. BLOCKED on one new one: `PATCH /a2a/receipts` 500s for every real caller.

Reviewed on head `3052bc4`. 97 of 98 pass in the four affected suites (the one failure is
pre-existing, see the end).

### NEW BLOCKER: the headline endpoint raises `NameError` on every authenticated call

`taosmd/service.py` uses `time.time()` twice and **never imports `time`**:

```
service.py:641   ts = time.time()      (a2a_record_seen)
service.py:664   ts = time.time()      (a2a_record_delivered)
imports of time in service.py: NONE
```

Called directly, with the explicit-`ts` path as the control:

```
a2a_record_seen(1, "agent-1", data_dir=d)                  -> *** NameError: name 'time' is not defined ***
a2a_record_seen(1, "agent-1", ts=1786000000.0, data_dir=d) -> {'ok': True}
```

The control is what makes this precise: the store, the schema and the write all work. It is only
the default-`ts` line that is unbound.

**Blast radius is exactly one endpoint, and it is the primary one.** Of the four call sites, three
pass `ts` explicitly and are fine:

```
http_server.py:1650  SSE delivered      ts=time.time()   OK (http_server does import time)
http_server.py:1734  POST /a2a/receipts/delivered  ts=ts  OK
http_server.py:1759  POST /a2a/receipts/seen       ts=ts  OK
http_server.py:1682  PATCH /a2a/receipts           no ts   <-- 500 NameError
```

`PATCH /a2a/receipts` is the one documented at line 111 as *"mark message seen by authenticated
agent"*.

**Why the suite is green on it.** The only test that reaches that handler is
`test_patch_receipts_seen_rejects_forged_token`, which asserts 401 and so returns before the
service call. Nothing exercises the success path of this endpoint. That is the same shape as
#232's blocker 2 (a routed endpoint whose handler could not work), which is why I want it fixed
here rather than carried.

Fix is one line (`import time` in `service.py`) plus a test that PATCHes with a **valid** token
and asserts 200 and a stored receipt. Attribution is clean: master's `service.py` has neither the
import nor any `time.time()` use, so both arrived with this PR.

### The five original blockers are all fixed

**1. Forgeable receipt identity: FIXED.** `_get_authenticated_agent_id` now returns the `sub` from
`_registry_verifier.authorize(token, claimed_identity)` and returns `None` on `AuthError`. The
unverified decode survives, but only to supply the `claimed_identity` argument that `authorize`
requires; the returned identity comes from the verified claims.

**A note on how I nearly got this wrong.** `verify_signature: False` appears **twice on the base
and twice on this revision**, so a grep count is identical under "fixed" and "not fixed" and
proves nothing either way. #252 was blocked partly on that count. Here the count is unchanged and
the defect is genuinely gone. What settles it is where the returned value comes from.

The PR ships a real behavioural test for this (`test_patch_receipts_seen_rejects_forged_token`
signs with an unknown key and asserts 401), and it passes. I confirmed the guard is what rejects
by patching only the identity source to trust the unverified sub, changing nothing else, and
watching the test stop returning 401.

**2. `POST /a2a/admin/prune-receipts` had no handler: FIXED.** Route (1101), handler (2231),
service call (2247), admin-route list (789) and `RemoteClient.a2a_prune_receipts` (292) are all
present.

**3. Service wrappers forwarding to absent `remote` methods: FIXED.** Cross-checked mechanically
rather than by eye:

```
service.py calls 28 remote methods; remote.py defines 29; MISSING: none
RemoteClient call paths: 28; UNROUTED: none
(control: a method known to be absent is reported by the same check)
```

**4. No tests: FIXED.** 138 test lines across four files, and the forgery test is behavioural
rather than structural, which is the right call for an auth guard.

**5. `receipts.py` bypassing `_db.connect`: FIXED.** Now `from ._db import connect`, so it gets
WAL and the busy timeout like every other store.

### Not a regression from this PR

`test_search_token_project_id_scopes_results` fails locally on this branch **and identically on
master**, with the embed-backend warning. It is the known live-embed-backend class, not something
this PR broke. Worth a separate look though: #251 added the `live_embed_backend` fixture to skip
that family, both refs have the fixture, and this test still fails without a backend, so it may
have been missed when that net was cast.

jaylfc:
## Closing so `tsk-qo6tpb` can be redispatched. The findings stand; this is a mechanism fix, not a verdict change.

while this PR sits here **`tsk-qo6tpb` cannot be claimed by anything** and the fix has no route.

revisions (#260, #261).

and every finding is also recorded in my checkpoint. The next build on `tsk-qo6tpb` should start from
the review above.

Jay's call, 2026-08-14.

PR #264, opened 2026-08-14
Revise PR #242: bearer token still crosses origins on redirect, plus a missing route

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #242: bearer token still crosses origins on redirect, plus a missing route

Autonomous build of board card tsk-revvkw.

Files:
 taosmd/config.py          |  47 +++++++
 taosmd/http_server.py     |  16 +++
 taosmd/ref_fetch.py       | 154 +++++++++++++++++++++++
 taosmd/remote.py          |   8 ++
 taosmd/service.py         |  62 ++++++++-
 tests/test_http_server.py |  35 ++++++
 tests/test_ref_fetch.py   | 314 ++++++++++++++++++++++++++++++++++++++++++++++
 7 files changed, 635 insertions(+), 1 deletion(-)

### DIFF
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -42,6 +42,8 @@ _REGISTRY_URL_KEY = "registry_url"
 # Key under which the registry auth token (taOS local/admin token) is stored.
 # Used to poll the auth-gated registry revoked feed; the pubkey feed is public.
 _REGISTRY_TOKEN_KEY = "registry_token"
+# Key under which the taOS Files base URL is stored.
+_FILES_URL_KEY = "files_url"
 # Who manages this taosmd instance: "standalone" (default) or "taos".
 _MANAGED_BY_KEY = "managed_by"
 # Override: serve the web dashboard even when managed_by=taos.
@@ -415,6 +417,49 @@ def set_registry_token(token: str, clear: bool = False, data_dir=None) -> None:
     _write(data, data_dir)
 
 
+def get_files_url(data_dir=None) -> str | None:
+    """Return the configured taOS Files base URL, or ``None`` if unset.
+
+    Resolution order (first non-empty wins):
+
+    1. ``TAOSMD_FILES_URL`` environment variable
+    2. ``files_url`` key in ``~/.taosmd/config.json``
+
+    When set, the ref-fetch helper resolves ``taos://`` refs against this
+    controller. When unset, the helper falls back to ``registry_url`` so a
+    single-controller install needs only one setting.
+    """
+    env = os.environ.get("TAOSMD_FILES_URL")
+    if env and env.strip():
+        return env.strip()
+    url = _read(data_dir).get(_FILES_URL_KEY)
+    if isinstance(url, str) and url.strip():
+        return url.strip()
+    return None
+
+
+def set_files_url(url: str, clear: bool = False, data_dir=None) -> None:
+    """Persist the taOS Files base URL (or clear it).
+
+    Args:
+        url: Base URL of the taOS controller serving the Files API, e.g.
+            ``"http://taos:8000"``. Ignored when ``clear`` is True.
+        clear: when True, remove the setting (ref-fetch falls back to
+            ``registry_url``).
+
+    Raises:
+        ValueError: when ``clear`` is False and ``url`` is not a non-empty string.
+    """
+    data = _read(data_dir)
+    if clear:
+        data.pop(_FILES_URL_KEY, None)
+    else:
+        if not isinstance(url, str) or not url.strip():
+            raise ValueError("url must be a non-empty string (or pass clear=True)")
+        data[_FILES_URL_KEY] = url.strip()
+    _write(data, data_dir)
+
+
 # ---------------------------------------------------------------------------
 # Remote server bearer token
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


jaylfc:
## All three blockers on #242 are fixed and I proved each by running. One required change: the test guarding the token does not actually guard it.

Reviewed on head `edfe704`. 88 tests pass across `test_ref_fetch.py` and `test_http_server.py`.

### Blocker 1, the bearer token crossing origins on redirect: FIXED, proven both directions

Same two-server setup that closed #238: a controller that 302s to an attacker origin, and an
attacker that records what arrives.

```
#242 pattern (plain urlopen):
  outcome: fetch returned normally
  attacker received: 'Bearer SECRET-REGISTRY-TOKEN-DO-NOT-LEAK'
  VERDICT: *** TOKEN LEAKED ***

#264 pattern (_NoRedirect opener):
  outcome: HTTPError 302 (redirect not followed)
  attacker received: None
  VERDICT: no token reached the attacker origin
```

The vulnerable half is the control: it shows the probe is capable of catching a leak, so the
clean result on this branch means something.

### Blocker 2, `POST /refs/fetch` had no server route: FIXED

`http_server.py` is now in the diff and dispatches `POST /refs/fetch` to `_handle_refs_fetch`,
which validates `ref` and `agent` before calling the service.

### Blocker 3, guard and fetch resolving different base URLs: FIXED

There is now exactly one resolution site. `service.fetch_by_ref` no longer computes `files_url`
itself; it passes `data_dir` through and `ref_fetch.fetch_by_ref` calls `_get_files_url(data_dir)`
once. They cannot diverge because there is nothing left to diverge. Measured:

```
files_url only in the data_dir config -> https://only-in-datadir.example
registry_url fallback, no files_url   -> https://registry-only.example
CONTROL neither configured            -> RAISED "files_url is not configured"
```

The control matters here: on #242 the guard passed with a data_dir-only value while the fetch
raised, so "it resolved" was not by itself evidence.

### The traversal fix from #242 still holds

```
taos://slug/files/ok.txt                     -> resolves (control)
taos://slug/files/../../etc/passwd           -> REJECTED
taos://slug/files/%2e%2e/%2e%2e/etc/passwd   -> REJECTED
taos://slug/files/a/./../../etc/passwd       -> REJECTED
taos://slug/files/%252e%252e/etc/passwd      -> resolves to %25252e%25252e (neutralised)
```

Last row is the same undocumented-but-load-bearing mechanism I flagged on #242: `quote(path,
safe="/")` re-encodes the `%`, so a double-encoded segment becomes a literal filename rather
than traversal. It is still worth a comment saying so, because a future change to the quoting
would silently remove a defence nobody knows is there.

### REQUIRED before merge: `test_fetcher_uses_no_redirect_opener` passes while the token leaks

The fix is right; the test does not defend it. It monkeypatches `build_opener` away entirely, so
no HTTP happens, and then asserts only:

```python
assert issubclass(build_opener_calls[0][0], urllib.request.HTTPRedirectHandler)
```

That asserts the **shape** of the fix, not the property. I ran it against a handler that inherits
the default behaviour:

```python
class _NoRedirect(urllib.request.HTTPRedirectHandler):
    pass   # inherits the default: still follows redirects
```

```
PR's assertion (issubclass of HTTPRedirectHandler): PASSES
attacker received: 'Bearer SECRET-REGISTRY-TOKEN-DO-NOT-LEAK'
VERDICT: *** TOKEN LEAKED while the PR's test passes ***
```

So the one test standing between us and a repeat of #238 goes green on the vulnerable version.
Given this token is the reason #238 was closed and rebuilt, that guard needs to be behavioural.

**Suggested replacement**, which is what I ran above: stand up two `http.server` instances on
ephemeral ports, have the first 302 to the second, point the real fetcher at the first, and
assert the second received no `Authorization` header. It needs no network and no new dependency.
Keep the existing structural assertion as well if you like, but it cannot be the only one.

### One smaller note

A redirecting controller now surfaces as a bare `urllib.error.HTTPError` (302). The `except`
chain handles 401/403/404 and re-raises the rest, so it escapes the `RefFetchError` family and
the HTTP route turns it into a 500. Since refusing a redirect is now a deliberate security
decision rather than a transport failure, it deserves its own message, e.g. raising
`RefFetchError("refusing to follow redirect from <origin> to <origin>")`. That also makes the
behaviour discoverable to whoever misconfigures a controller.

jaylfc:
Merging as-is on Jay's call, with the required test change carded as a follow-up rather than blocking the fix.

**What is landing:** the `_NoRedirect` opener. Verified correct in both directions - the redirect no longer carries `Authorization` to a second origin, and the control (same fetcher, redirect handler restored) still leaks. That is the defect that closed #238 and it is genuinely fixed here.

**What is NOT landing, and why it is carded:** `test_fetcher_uses_no_redirect_opener` monkeypatches `build_opener` away and then asserts only `issubclass(..., HTTPRedirectHandler)`. I measured that a do-nothing subclass satisfies that assertion while the token still leaves the origin. So the test passes on code that has the bug, and CI does not defend the guard this PR exists to install.

The honest statement of what merges here: the fix is proven by my probe, not by the suite. A future refactor of `ref_fetch.py` can reintroduce the leak and stay green. Follow-up card replaces the test with one that runs a real redirect against a local server and asserts the second origin received no `Authorization` header - the assertion has to be about the header on the wire, not about the class of an opener.

PR #257, opened 2026-08-13
Revise PR #250: measure REAL channel membership, not spool senders (Stage 2 entry gate)

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #250: measure REAL channel membership, not spool senders (Stage 2 entry gate)

Autonomous build of board card tsk-aildfj.

Files:
 docs/specs/tsk-rf5gwb-membership-stems.md |  79 +++++++++++
 scripts/measure_membership_stems.py       | 219 ++++++++++++++++++++++++++++++
 tests/test_measure_membership_stems.py    | 114 ++++++++++++++++
 3 files changed, 412 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added a command-line measurement tool for analyzing membership identity stems from archive or JSONL data.
  * Reports spelling variations, canonical and `@` forms, principal collapses, collisions, and install-discriminator distinctions.
  * Provides a safety conclusion for membership reconciliation.
  * Skips malformed JSON records and reports when no supported data source is available.

* **Documentation**
  * Added a closed specification describing normalization, measurement, collision checks, and reconciliation safety findings.

PR #265, opened 2026-08-14
Revise PR #246: nan and inf walk straight through the new since guard

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #246: nan and inf walk straight through the new since guard

Autonomous build of board card tsk-dx6fok.

Files:
 taosmd/http_server.py |   4 ++
 tests/test_a2a.py     | 100 ++++++++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 104 insertions(+)

----
## Summary by Gitar

- **Validation updates:**
    - Added `math.isfinite` check for `since` parameter in HTTP messages and stream endpoints
    - Fall back to `time.time()` for non-finite `since` values in stream endpoints and return 400 for invalid inputs
- **Test additions:**
    - Added tests verifying `nan`, `inf`, and `-inf` handling on `since` filters for messages and stream endpoints

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1511,6 +1511,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                 since = float(since_raw) if since_raw is not None else None
             except (TypeError, ValueError) as exc:
                 raise _BadRequest("'since' must be a float timestamp") from exc
+            if since is not None and not math.isfinite(since):
+                raise _BadRequest("'since' must be a finite timestamp")
             try:
                 limit_i = int(limit_raw)
             except (TypeError, ValueError) as exc:
@@ -1557,6 +1559,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             try:
                 last_ts = float(since_raw) if since_raw is not None else time.time()
             except (TypeError, ValueError):
+                last_ts = time.time()
+            if not math.isfinite(last_ts):
                 last_ts = time.time()
 
             # Send SSE response headers before entering the poll loop.
--- a/tests/test_a2a.py
+++ b/tests/test_a2a.py
@@ -310,6 +310,27 @@ def test_http_a2a_messages_since_filter(live_server):
     assert msgs[0]["body"] == "after pivot"
 
 
+def test_http_a2a_messages_since_nan_returns_400(live_server):
+    """GET /a2a/messages?since=nan returns 400 (NaN evades magnitude check)."""
+    status, body = _get(f"{live_server}/a2a/messages?since=nan&thread=test")
+    assert status == 400
+    assert "since" in body["error"] or "finite" in body["error"].lower()
+
+
+def test_http_a2a_messages_since_inf_returns_400(live_server):
+    """GET /a2a/messages?since=inf returns 400 (inf evades magnitude check)."""
+    status, body = _get(f"{live_server}/a2a/messages?since=inf&thread=test")
+    assert status == 400
+    assert "since" in body["error"] or "finite" in body["error"].lower()
+
+
+def test_http_a2a_messages_since_ninf_returns_400(live_server):
+    """GET /a2a/messages?since=-inf returns 400 (-inf evades magnitude check)."""
+    status, body = _get(f"{live_server}/a2a/messages?since=-inf&thread=test")
+    assert status == 400
+    assert "since" in body["error"] or "finite" in body["error"].lower()
+
+
 # ---------------------------------------------------------------------------
 # SSE smoke test
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


jaylfc:
## BLOCKED. This revision fixes `nan`/`inf` and **reverts the feature it was revising**.

Reviewed by running all three refs, not by reading the diff. The `nan`/`inf` fix itself is
correct. The problem is what came with it: **PR #246's epoch-floor guard, which is the entire
point of card tsk-jpioif, is not in this branch at all.**

### Measured: the same six values against master, #246 and this PR

| `?since=` | master | #246 (`exec/tsk-jpioif`) | **this PR (#265)** |
|---|---|---|---|
| `12345` (a message id) | 200 | **400** | **200** <- back to master |
| `0` | 200 | **400** | **200** <- back to master |
| `-1` | 200 | **400** | **200** <- back to master |
| `nan` | 200 | 200 | **400** <- what this PR adds |
| `inf` | 200 | 200 | **400** <- what this PR adds |
| `1786000000` (valid) | 200 | 200 | 200 |

The last row is the control: a legitimate timestamp is accepted on all three refs, so the
probe is not just printing 400 for everything.

**This trades one hole for another.** The id-vs-timestamp footgun (taOS-dev #1447), the thing
the card exists to close, is live again on this branch: `since=12345` parses as 1970 and
replays the full history, silently.

### It is structural, not a stale rebase

`grep` for the floor on this head returns nothing: no `1_000_000_000`, no `1e9`, no
"looks like a message id". The two-line diff against `taosmd/http_server.py` is the isfinite
check and nothing else, applied to bare master.

The tests confirm it independently. #246 shipped six tests; **this branch carries none of them**:

```
#246 has:  since_message_id_rejected, stream_since_message_id_rejected,
           since_zero_rejected, since_negative_rejected,
           since_omitted_unchanged, since_recent_epoch_unchanged
#265 has:  since_nan_returns_400, since_inf_returns_400, since_ninf_returns_400,
           stream_since_nan_falls_back, stream_since_inf_falls_back
```

Content-level carry-forward across the files #246 touched: **2 of 35 lines survive (5%)**.
CI is green because the tests that would have caught this were dropped along with the code.

### Why this got past the existing no-op check

My own earlier validation of the STEP 0 squash treatment asked whether **every file** the base
branch added is present in the revision. **#265 passes that check**: it touches
`http_server.py` and `tests/test_a2a.py`, exactly the files #246 touches. The file is there and
the content is gone. A byte-identical-to-base comparison does not catch it either, because this
branch is not identical to anything: it has a real diff, real new tests and a green build.

So this is a distinct class from the four no-op revisions (#252, #254, #257, #262), and it is
the more dangerous one. **A no-op leaves the base PR's work intact and available. This one looks
like a completed revision while discarding the feature**, and under the rule that a base PR is
closed once its revision merges, merging this would close #246 and lose the floor permanently
with the card recorded as done.

### Third behaviour on the stream endpoint

`/a2a/stream` now has three different contracts across three refs for the same bad input:
#246 raises 400, this PR falls back to `time.time()` silently (`stream_since_nan_falls_back`),
and #253 raises 400 via a shared `_parse_since`. #246's own docstring promises "values below
1_000_000_000 are rejected" for `/a2a/stream`. Whatever lands should pick one and say so.

### Required to unblock

1. Restore #246's epoch floor on **both** `/a2a/messages` and `/a2a/stream`, and restore its six
   tests. `since=12345` must return 400 with the nan/inf fix still in place.
2. Decide the stream contract explicitly (400 or fallback) and make the docstring match.
3. **Do not close #246 until this branch actually contains its work.**

### Merge-order hazard with #253, please read before either lands

**#253 (tsk-3fuapf) rewrites this exact code path** into a shared `_parse_since()` helper that
already does the finite check **and** the epoch floor. I merged both onto master in a scratch
tree: they conflict in `taosmd/http_server.py`, and I checked the conflict by marker position
rather than grepping the file.

```
<<<<<<< HEAD                                  (#253)
            since = _parse_since(since_raw)
=======                                       (#265)
            try:
                since = float(since_raw) if since_raw is not None else None
            except (TypeError, ValueError) as exc:
                raise _BadRequest("'since' must be a float timestamp") from exc
            if since is not None and not math.isfinite(since):
                raise _BadRequest("'since' must be a finite timestamp")
>>>>>>>
```

Both sides look correct in isolation, so a resolver has no signal that taking the #265 side
drops the epoch floor. **Recommended: land #253 first.** Its `_parse_since` already covers
everything this card asks for, verified by running on its branch: `nan`, `inf`, `-inf` and
`12345` all return 400 while `1786000000` returns 200. If #253 lands, this card is close to
satisfied by it and the remaining work is the stream contract plus porting #246's six tests.

jaylfc:
## Closing so `tsk-dx6fok` can be redispatched. The findings stand; this is a mechanism fix, not a verdict change.

while this PR sits here **`tsk-dx6fok` cannot be claimed by anything** and the fix has no route.

revisions (#260, #261).

and every finding is also recorded in my checkpoint. The next build on `tsk-dx6fok` should start from
the review above.

Jay's call, 2026-08-14.

PR #269, opened 2026-08-14
Revise PR #246: nan and inf walk straight through the new since guard

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #246: nan and inf walk straight through the new since guard

Autonomous build of board card tsk-dx6fok.

Files:
 taosmd/http_server.py | 14 ++++++++--
 tests/test_a2a.py     | 71 +++++++++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 83 insertions(+), 2 deletions(-)

----
## Summary by Gitar

- **API Validation Updates:**
    - Reject `since` timestamp values below `1_000_000_000` with a 400 error in `/a2a/messages` and `/a2a/stream` endpoints
- **Testing:**
    - Added comprehensive unit tests in `tests/test_a2a.py` for rejected and accepted `since` parameter values

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -104,8 +104,8 @@ Endpoints
 ``POST /a2a/send``         ``{"from", "body", "thread"?, "reply_to"?, "refs"?, "blocks"?}`` -> send receipt
                            ``refs``: optional list (<=8) of ``{"kind": doc|report|spec|log, "title", "uri", "sha256"?, "doc_id"?, "version"?, "for"?, "summary"?}``
                            ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
-``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line)
-``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream)
+``GET  /a2a/messages``     ``?thread=&since=<epoch_ts>&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line; ``since`` is an epoch timestamp in seconds, values below 1_000_000_000 are rejected)
+``GET  /a2a/stream``       ``?thread=&since=<epoch_ts>``             -> SSE stream (text/event-stream; ``since`` is an epoch timestamp in seconds, values below 1_000_000_000 are rejected)
 ``GET  /a2a/channels``                                     -> ``{"channels": [...]}``
 ``GET  /a2a/members``      ``?channel=<name>``             -> ``{"members": [...]}``
 ``POST /tasks``            ``{"title", "body"?, "project"?, "assignee"?, "priority"?, "depends_on"?: [...], "created_by"}`` -> task object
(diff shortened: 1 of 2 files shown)


jaylfc:
Closing this as **superseded**, and the reason is not "master moved" - it is that **this PR does not fix the defect in its own title, and master already does.** Everything below was run, not read.

## 1. `nan` and `inf` still walk straight through

The title is "nan and inf walk straight through the new since guard". The change adds an epoch-floor comparison at the two handler sites:

```python
if since is not None and since < 1_000_000_000:
    raise _BadRequest(...)
```

`float("nan") < 1e9` is `False` and `float("inf") < 1e9` is `False`, so neither is caught. Live probe against a real server on this branch, controls in the same run:

```
CODE UNDER TEST: .../wt269/taosmd/http_server.py     (has _parse_since helper: False)
       since |    /a2a/messages |                /a2a/stream | note
         nan |              200 |              ACCEPTED(200) | <-- the defect this PR is named after
         inf |              200 |              ACCEPTED(200) | <-- the defect this PR is named after
        -inf |    REJECTED(400) |              REJECTED(400) |
        1444 |    REJECTED(400) |              REJECTED(400) | CONTROL msg id      -> must REJECT
           0 |    REJECTED(400) |              REJECTED(400) | CONTROL below floor -> must REJECT
         abc |    REJECTED(400) |              ACCEPTED(200) | CONTROL unparseable -> must REJECT
  1786000000 |              200 |              ACCEPTED(200) | CONTROL valid epoch -> must ACCEPT
```

The `-inf` row is what proves the mechanism rather than just the symptom: `-inf < 1e9` IS `True`, so it is rejected incidentally by the comparison, while `nan` and `+inf` are not. A comparison-based guard cannot catch non-finite values; only a finite check can.

**And the tests here do not cover it.** The five new tests exercise `1444`, `0`, `-1`, omitted, and a valid epoch. Not one passes `nan` or `inf`. That is why CI is green on a PR whose stated defect is unfixed - same shape as the vacuous-test problem on #264.

## 2. A second defect this introduces on `/a2a/stream`

Row `abc` above: `/a2a/stream?since=abc` returns **200 on this branch and 400 on master**. The stream handler keeps its `except (TypeError, ValueError): last_ts = time.time()`, so an unparseable `since` silently becomes "now" instead of erroring - a client with a malformed cursor gets a stream that looks fine and silently skips history.

## 3. Master already does all of it, via #253 (`e3ea7c3`)

Same probe, same controls, against master:

```
CODE UNDER TEST: .../wtmaster/taosmd/http_server.py  (has _parse_since helper: True)
         nan |    REJECTED(400) |              REJECTED(400)
         inf |    REJECTED(400) |              REJECTED(400)
        -inf |    REJECTED(400) |              REJECTED(400)
        1444 |    REJECTED(400) |              REJECTED(400)
           0 |    REJECTED(400) |              REJECTED(400)
         abc |    REJECTED(400) |              REJECTED(400)
  1786000000 |              200 |              ACCEPTED(200)
```

`_parse_since` (http_server.py:267) does `math.isfinite` **and** the 1e9 floor, and it is wired into both call sites (:1617 messages, :1661 stream) - the exact two sites this PR patches inline.

## 4. The check that settles it

I lifted **this PR own test file** onto master unchanged and ran it there:

```
8 passed, 28 deselected
```

**Every acceptance test this PR ships is already green on master without this PR code.** That is the cleanest possible statement that the code change is redundant.

## What is NOT redundant, and is being kept

`test_http_a2a_stream_since_message_id_rejected` covers **`/a2a/stream`**, and master has no `since`-rejection test on the stream endpoint at all - master two tests (`test_a2a_since_nan_rejected`, `test_a2a_since_inf_rejected`) only hit `/a2a/messages`. Master stream behaviour is correct today but unguarded, so a regression there would be silent. Carded, with the note that these tests pass on master as-is.

This PR conflicts with master and is not mergeable as it stands. No work is being thrown away that is not being carried into that card. Branch untouched.

jaylfc:
Closed as superseded by #253 (e3ea7c3). Full measurement in the comment above: this branch's own acceptance tests pass on master without this branch's code, and its guard does not catch nan/inf. Stream-endpoint test coverage salvaged to a card.

PR #256, opened 2026-08-13
Revise PR #236: duplicate CHANGELOG entry, a README sentence about the container, and three merged PRs still unlogged

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #236: duplicate CHANGELOG entry, a README sentence about the container, and three merged PRs still unlogged

Autonomous build of board card tsk-xhm3ln.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 CHANGELOG.md | 12 ++++++++++++
 README.md    | 25 ++++++++++++++++++++++++-
 2 files changed, 36 insertions(+), 1 deletion(-)

----
## Summary by Gitar

- **Documentation & Setup:**
    - Added dev setup instructions using `uv sync` and clarified that `pip install -e .` installs runtime only
- **Changelog:**
    - Added unreleased entries for PRs #212, #215, #219, #240, #243, and #245

<sub>This will update automatically on new commits.</sub>

PR #270, opened 2026-08-14
Revise PR #232: receipt identity is forgeable, and a routed admin endpoint has no handler

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #232: receipt identity is forgeable, and a routed admin endpoint has no handler

Autonomous build of board card tsk-qo6tpb.

Files:
 taosmd/api.py         |   4 ++
 taosmd/http_server.py | 110 ++++++++++++++++++++++++++++++++++-
 taosmd/migrations.py  |  18 ++++++
 taosmd/receipts.py    | 158 ++++++++++++++++++++++++++++++++++++++++++++++++++
 taosmd/service.py     |  98 +++++++++++++++++++++++++++++++
 5 files changed, 387 insertions(+), 1 deletion(-)

### DIFF
--- a/taosmd/api.py
+++ b/taosmd/api.py
@@ -144,12 +144,16 @@ async def _ensure_stores(data_dir=None) -> dict:
         from taosmd.claims.store import ClaimStore  # noqa: PLC0415
         claims = ClaimStore(db_path=str(path / "claims.db"))
         await claims.init()
+        from taosmd.receipts import ReceiptStore  # noqa: PLC0415
+        receipts = ReceiptStore(db_path=str(path / "a2a-receipts.db"))
+        await receipts.init()
 
         stores = {
             "archive": archive,
             "vector": vmem,
             "kg": kg,
             "claims": claims,
+            "receipts": receipts,
             "data_dir": str(path),
         }
         _stores_cache[resolved] = stores
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -108,6 +108,9 @@ Endpoints
 ``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream)
 ``GET  /a2a/channels``                                     -> ``{"channels": [...]}``
 ``GET  /a2a/members``      ``?channel=<name>``             -> ``{"members": [...]}``
+``PATCH /a2a/receipts``    ``{"message_id"}``              -> mark message seen by authenticated agent
+``GET  /a2a/messages/{id}/receipts``                       -> ``{"delivered": [...], "read": [...]}``
+``GET  /a2a/receipts``     ``?message_id=&agent=``         -> ``{"delivered_at", "seen_at"}`` or 404
 ``POST /tasks``            ``{"title", "body"?, "project"?, "assignee"?, "priority"?, "depends_on"?: [...], "created_by"}`` -> task object
 ``GET  /tasks``            ``?status=&project=&assignee=&limit=``  -> ``{"tasks": [...]}``
 ``GET  /tasks/ready``      ``?project=&assignee=&limit=``  -> ``{"tasks": [...]}``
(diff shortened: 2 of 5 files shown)


jaylfc:
Reviewed. **Blocked: 3 blockers, all three are the original #232 defects unchanged.** Everything below was run, with the positive control in the same command.

One thing is genuinely better than its predecessors and I want to lead with it, because it is the defect that broke #230 and #266: **the new migration is registered correctly.** `_A2A_RECEIPTS` is added to `REGISTRY` *and* `a2a-receipts.db` to `DB_FILES`, with a `table_exists` guard rather than a column check. That is the fix those two PRs got wrong, done right here.

## BLOCKER 1 - Receipt identity is still forgeable, and the docstring now claims otherwise

`_get_authenticated_agent_id` (http_server.py:729) is unchanged in substance:

```python
unverified = _jwt.decode(token, options={"verify_signature": False})
sub = unverified.get("sub", "") or ""
return sub if sub else None
```

Run against that exact code path, with a token signed by a key the server has never seen:

```
FORGED token signed with a key the server has never seen.
  _get_authenticated_agent_id returns: 'taOS-dev'

POSITIVE CONTROL - the real verifier, same forged token:
  decode_and_verify REJECTED the forgery: AuthError
  CONTROL legit token ACCEPTED, sub = taosmd-dev

CONTROL - token with no sub: None
CONTROL - garbage:           None
```

So any client can still write receipts as any agent, and the correct API rejects the very token this path accepts.

**Two things make this worse than it was on #232.** First, the correct call is **in this same file, 126 lines below**: `_registry_verifier.authorize(token, claimed_identity)` at line 873. The file now contains the right pattern and the wrong pattern, and the receipts path picked the wrong one again.

Second, the new docstring says:

> *"Return the agent identity from a **verified** registry token"*

The code does not verify. A docstring asserting a security property the code lacks is worse than the untouched original - it is what a future reviewer will read instead of the four lines below it. If nothing else in this PR changes, that word must go.

## BLOCKER 2 - `POST /a2a/admin/prune-receipts` still has no handler

Dispatched at line 1090 (`self._handle_admin_a2a_prune_receipts()`), never defined. Enumerated both sides so this is not a grep artefact:

```
DEFINED (6)                              CALLED (7)
_handle_admin_a2a_delete_channel          _handle_admin_a2a_delete_channel
_handle_admin_a2a_rename_channel          _handle_admin_a2a_prune_receipts   <-- no definition
_handle_admin_a2a_supersede_message       _handle_admin_a2a_rename_channel
_handle_admin_shelf_archive               _handle_admin_a2a_supersede_message
_handle_admin_shelf_create                _handle_admin_shelf_archive
_handle_admin_shelf_unarchive             _handle_admin_shelf_create
                                          _handle_admin_shelf_unarchive
```

Six of seven resolve; the one that does not is exactly the route this PR adds. The other six are the control - the grep finds definitions when they exist. The endpoint is documented at line 153 and will `AttributeError` on the first call. **This is #232 defect 2, unchanged.**

## BLOCKER 3 - Five service wrappers forward to `remote` methods that do not exist

`remote.py` is not in this diff at all, yet `service.py` gained calls into it:

```
  present: a2a_channels            MISSING: a2a_get_receipt
  present: a2a_feed                MISSING: a2a_get_receipts
  present: a2a_members             MISSING: a2a_prune_receipts
  present: a2a_send                MISSING: a2a_record_delivered
                                   MISSING: a2a_record_seen
```

The four present ones are the control: pre-existing wrappers resolve, and **every one of the five added by this PR is missing.** So on any remote-configured install, all five receipt operations raise `AttributeError`. Local-only suites cannot see it - which is the whole reason **tsk-bwgr26** was carded (a CI gate matching `RemoteClient._run` literals against real routes). This is now the fourth PR in a row shipping a client half with no server half (#232, #234, #242, this one).

## DEFECT 4 - No tests at all

Zero changes under `tests/`. 387 added lines, a new database, a new admin route and a new identity path, with no coverage. That is #232 defect 4 unchanged, and it is why all five checks are green on a PR whose three blockers are each reproducible in a few lines.

## Also

This branch **conflicts with master** as it stands.

## What I recommend

Do not rebuild - `receipts.py` and the migration are sound and worth keeping. The fix list is short and mechanical:

1. Replace the unverified decode with `_registry_verifier.authorize(...)` (line 873 is the working example), and drop "verified" from the docstring until it is true.
2. Write `_handle_admin_a2a_prune_receipts`.
3. Add the five `RemoteClient` methods.
4. Tests, including one that presents a forged token and asserts **no** receipt is written - that assertion is the point of the card, so it must go red without the fix.

jaylfc:

- The revision card **`tsk-otz3nn`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-qo6tpb` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-qo6tpb` recovers the work.
- The originating card `tsk-qo6tpb` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #272, opened 2026-08-14
Replace the vacuous no-redirect test in tests/test_ref_fetch.py with one that asserts the header on the wire

jaylfc:
CARD TITLE (intent, not commit subject): Replace the vacuous no-redirect test in tests/test_ref_fetch.py with one that asserts the header on the wire

Autonomous build of board card tsk-ljvoh7.

Files:
 tests/test_ref_fetch.py | 150 +++++++++++++++++++++++++++++++++++++-----------
 1 file changed, 115 insertions(+), 35 deletions(-)

### DIFF
--- a/tests/test_ref_fetch.py
+++ b/tests/test_ref_fetch.py
@@ -174,6 +174,121 @@ class TestFetchByRef:
         with pytest.raises(RefFetchError, match="no sha256"):
             asyncio.run(fetch_by_ref(ref, fake_fetcher, "test-agent"))
 
+    def test_redirect_no_authorization_leak(self, monkeypatch):
+        import http.server
+        import threading
+        import time
+        import urllib.request
+        import urllib.error
+        import asyncio
+        import hashlib
+
+        from taosmd import service as svc
+        from taosmd import config as cfg
+
+        # shared state between test and HTTP servers
+        recorded = {"headers": {}}
+
+        # Origin B: records every request header it receives
+        class OriginBHandler(http.server.BaseHTTPRequestHandler):
+            def do_GET(self):
+                recorded["headers"] = dict(self.headers)
+                self.send_response(200)
+                self.send_header("Content-Type", "application/octet-stream")
+                self.end_headers()
+                self.wfile.write(b"ok")
+            def log_message(self, format, *args):
+                pass
+
+        # Origin A: returns 302 redirect to origin B
+        class OriginAHandler(http.server.BaseHTTPRequestHandler):
+            def do_GET(self):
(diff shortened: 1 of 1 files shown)


jaylfc:
Reviewed. **This does what the card asked and I verified it by running the full red/green cycle, not by reading it.** One required one-line change, which I have already tested for you.

## The card bar is met

The vacuous `test_fetcher_uses_no_redirect_opener` is **deleted** - good, that was the point. The replacement stands up two real local servers, points the product fetch path at origin A, and asserts on the `Authorization` header **recorded at origin B**, which is the assertion the card demanded: about the header on the wire, not about the class of an opener.

**And it genuinely defends the guard.** Removing only `_NoRedirect` from the opener in `service.py:384`:

```
GREEN as shipped:                          1 passed
RED with build_opener(_NoRedirect) -> build_opener():   1 failed
```

That is a real red, which is exactly what the previous test could not produce.

## Required change: the red is right, but it fails for the wrong reason

In the red state the failure is:

```
E  taosmd.ref_fetch.HashMismatchError: sha256 mismatch
   taosmd/ref_fetch.py:121
```

not the leak assertion. Once the redirect is followed, origin B returns `b"ok"`, which does not match the expected `sha256(b"hello world")`, so `HashMismatchError` propagates out of `asyncio.run(...)` before `assert len(auth_keys) == 0` is ever evaluated. The test is red by incidental exception, and the message points a future reader at test data rather than at a credential leak.

**One-line fix, verified end to end:**

```diff
-                self.wfile.write(b"ok")
+                self.wfile.write(b"hello world")
```

With that, in the red state the failure becomes the assertion itself:

```
E  AssertionError: Authorization header leaked to origin B with _NoRedirect! Keys found: [Authorization]
E  assert 1 == 0
```

and restoring `_NoRedirect` goes green again with the same payload. Full cycle proven both directions.

## Two non-blocking notes

1. **Part 2 does not exercise the product code.** It hand-builds `urllib.request.build_opener()` with its own `Request` and a hardcoded `Authorization: Bearer test-token`, so what it proves is that *urllib default openers follow redirects and forward headers* - a property of the standard library, not of our fetch path. It is a useful sanity check that origin B can record a header at all, but the real defence is Part 1. Worth a comment saying so, otherwise it reads as the control when it is not; with the one-line fix above, Part 1 is the control.
2. `time.sleep(0.05)` after each `serve_forever` and a hardcoded `data_dir="/tmp"` - both fine in practice, both mild flake/portability risks if this test ever moves.

Recommending merge once the payload line is changed. Nothing else blocks it.

PR #275, opened 2026-08-14
A2A chat read API: thread list, cursor pagination, unread counts (unified-chat slice 4)

jaylfc:
CARD TITLE (intent, not commit subject): A2A chat read API: thread list, cursor pagination, unread counts (unified-chat slice 4)

Autonomous build of board card tsk-qthbox.

## CONSUMER-VISIBLE CONTRACT CHANGE — read before merging

`GET /a2a/threads` changes its sort order from **alphabetical by thread name** to
**most-recently-active first**. This is a shipped endpoint. Any consumer that relies on
the old ordering — including anything that reads the first element, or diffs the list
positionally — changes behaviour silently the moment this merges; nothing errors.

Measured before merge (2026-08-14), so the risk is stated with a number rather than as a
caution: **zero consumers today.** The taOSmd dashboard calls `/a2a/channels`,
`/a2a/members`, `/a2a/messages` and `/a2a/stream`, not `/a2a/threads`; a filesystem sweep
across `~/Development` and the fleet scripts (control: the same sweep for `/a2a/messages`
returns 8 files) finds no caller of `/a2a/threads` outside this repo. The endpoint appears
in `http_server.py`, `remote.py`, `service.py` and its tests, and nowhere else — it is not
mentioned in `docs/`, `README.md` or `CHANGELOG.md`.

So this is safe to land now, and the reason it is safe is that the endpoint has no
audience yet — not that the change is compatible. Flagged to @taOS-dev on the bus.

---

- fix a2a_threads sort order: was alphabetical by thread name, now ordered
  by latest activity timestamp descending
- strip internal _last_ts from returned thread objects (not part of API shape)
- export a2a_threads and a2a_thread_messages in service.__all__
- add tests: threads ordered by latest activity, before/after cursors,
  limit clamped at 200, stable pagination across concurrent insert,
  message id passed to since= rejected with clear 400 error

Files:
 taosmd/service.py         |  8 +++--
 tests/test_http_server.py | 91 +++++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 97 insertions(+), 2 deletions(-)

### DIFF
--- a/taosmd/service.py
+++ b/taosmd/service.py
@@ -760,9 +760,12 @@ async def a2a_threads(*, principal: str | None = None, data_dir=None) -> list[di
             "kind": t["kind"],
             "participants": sorted(t["participants"]),
             "last_message": t["last_message"],
+            "_last_ts": t["_last_ts"],
         })
-    result.sort(key=lambda x: x["thread"])
-    return result
+    result.sort(key=lambda x: x["_last_ts"], reverse=True)
+    return [{"thread": t["thread"], "kind": t["kind"],
+             "participants": t["participants"],
+             "last_message": t["last_message"]} for t in result]
 
 
 async def a2a_thread_messages(
@@ -1256,6 +1259,7 @@ async def collections_archive(collection_id: str, *, data_dir=None) -> dict:
 
 __all__ = ["ingest", "search", "pending_list", "pending_resolve", "reconcile", "stats",
            "supersede", "fetch_by_ref", "a2a_send", "a2a_feed", "a2a_channels", "a2a_members",
+           "a2a_threads", "a2a_thread_messages",
            "task_create", "task_list", "task_ready", "task_prime",
            "task_update", "task_add_edge", "task_remove_edge", "task_projects",
            "admin_shelf_create", "admin_shelf_archive", "admin_shelf_unarchive",
--- a/tests/test_http_server.py
+++ b/tests/test_http_server.py
@@ -440,6 +440,97 @@ def test_a2a_thread_messages_populated(live_server):
     assert body["messages"][1]["body"] == "second"
 
 
+def test_a2a_threads_ordered_by_latest_activity_desc(live_server):
+    """Thread list is ordered by most-recent message timestamp descending."""
+    import time
+    _post(f"{live_server}/a2a/send", {"from": "alice", "body": "old", "thread": "alpha"})
+    time.sleep(0.02)
+    _post(f"{live_server}/a2a/send", {"from": "bob", "body": "new", "thread": "beta"})
+    status, body = _get(f"{live_server}/a2a/threads")
+    assert status == 200
+    names = [t["thread"] for t in body["threads"]]
+    assert names == ["beta", "alpha"]
+
+
+def test_a2a_thread_messages_before_cursor(live_server):
+    """before=<last_id> walks backwards (older messages only)."""
+    r1 = _post(f"{live_server}/a2a/send", {"from": "alice", "body": "first", "thread": "cur"})
+    _post(f"{live_server}/a2a/send", {"from": "bob", "body": "second", "thread": "cur"})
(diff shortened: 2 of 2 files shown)


jaylfc:
**APPROVED with one required change.** The behaviour change is correct and I proved it in both directions rather than reading the diff. One documentation gap has to close before this lands, because after this PR the ordering is a contract with no written record anywhere.

## The change is real and its test is not vacuous

Red control — this branch's tests against **master's** `service.py`, everything else identical:

```
FAILED test_a2a_threads_ordered_by_latest_activity_desc
E   AssertionError: assert ['alpha', 'beta'] == ['beta', 'alpha']
1 failed, 5 passed
```

Green — the same tests against this branch: `6 passed`.

So the ordering test fails on master by its own assertion, naming the exact defect, and passes here. That is the bar #272/#273 established and this clears it.

**Worth stating plainly: the other five tests pass on master too.** They are regression coverage for the cursor/limit behaviour that already worked, not evidence for this change. One test defends the change; five defend what was already there. Both are worth having, but the PR reads as six-tests-for-a-fix and it is one.

## `_last_ts` really is stripped — checked on the wire, not in the diff

The return builds fresh dicts, but the internal key has to be absent from the actual response, so I asked the server:

```
GET /a2a/threads  ->  PAYLOAD KEYS: ['kind', 'last_message', 'participants', 'thread']
```

Exactly the four documented keys. No leak.

## Sort stability

`archive.py:56` declares `timestamp REAL`, so `_last_ts` is a float epoch and the test's `time.sleep(0.02)` is comfortably above resolution — no flake there. Ties are still possible in principle, and `list.sort` is stable, so two threads sharing a `_last_ts` fall back to archive scan order rather than anything a caller can predict. A `key=lambda x: (-x["_last_ts"], x["thread"])` would make the tie deterministic. Nit, not a blocker.

## Required change: the ordering is now undocumented behaviour

The `a2a_threads` docstring specifies the item shape and says nothing about order — that was fine when the order was alphabetical and obvious from the code, and it is not fine now. I checked where else a consumer could find out:

```
git grep -l "a2a/threads\|a2a_threads" origin/master
  taosmd/http_server.py  taosmd/remote.py  taosmd/service.py  tests/test_http_server.py
```

Nothing in `docs/`, `README.md` or `CHANGELOG.md`. So after this merges, the only statement of the endpoint's ordering guarantee is one `sort` line and one test. Add a sentence to the `a2a_threads` docstring — "Threads are ordered by most recent activity first" — and this is done. That is the whole required change.

## The contract change, and why it is safe today

I have added a section to the PR body rather than leaving this in review notes, since it is the thing a future reader needs at merge time. Summarised here: `/a2a/threads` is a shipped endpoint whose sort order changes from alphabetical to most-recently-active, and nothing errors if a consumer depended on the old order.

Measured rather than assumed: **zero consumers today.** The dashboard uses `/a2a/channels`, `/a2a/members`, `/a2a/messages` and `/a2a/stream` (`dashboard/src/api.ts`, `views/A2AView.tsx`); a sweep across `~/Development` and the fleet scripts finds no external caller of `/a2a/threads`, with the same sweep for `/a2a/messages` returning 8 files as the control that the sweep works. Flagged to @taOS-dev.

## Separate finding, pre-existing, not this PR's to fix

`a2a_threads(*, principal=None, ...)` accepts `principal`, `http_server.py:1697` reads it from the query string and passes it in, and the local path never uses it — the only reference in the function body is the forward to `RemoteClient`. A caller asking for a principal-scoped thread list gets every thread, with a 200 and no indication the filter was dropped. That arrived long before this PR and I am carding it separately rather than growing this one.

## Mechanics

- `git merge-tree --write-tree origin/master` — clean, no conflicts. Also clean against #276, which touches the same test file.
- Full suite on this head vs master, same runner and worktree layout: identical failure set of 10 (all pre-existing, all needing a live embed backend and deselected in CI), `1357 passed` here against `1351` on master — the +6 are exactly this PR's new tests.

PR #276, opened 2026-08-14
Add since-rejection tests for /a2a/stream (master validates it correctly but nothing covers it)

jaylfc:
CARD TITLE (intent, not commit subject): Add since-rejection tests for /a2a/stream (master validates it correctly but nothing covers it)

Autonomous build of board card tsk-q6fjux.

Files:
 tests/test_http_server.py | 184 ++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 184 insertions(+)

### DIFF
--- a/tests/test_http_server.py
+++ b/tests/test_http_server.py
@@ -9,8 +9,11 @@ needed. Requests go over the loopback via urllib.
 from __future__ import annotations
 
 import json
+import socket
 import threading
+import time
 import urllib.error
+import urllib.parse
 import urllib.request
 from pathlib import Path
 
@@ -18,6 +21,42 @@ import pytest
 
 from taosmd import api as taosmd_api
 from taosmd import http_server
+
+
+def _read_sse_frames(host: str, port: int, path: str, timeout: float = 8.0) -> list[str]:
+    """Open a raw TCP connection, send a minimal HTTP GET, read lines until
+    at least one ``data:`` frame arrives or timeout expires. Returns the list
+    of ``data:`` payload strings received."""
+    frames: list[str] = []
+    deadline = time.monotonic() + timeout
+    with socket.create_connection((host, port), timeout=timeout) as sock:
+        sock.sendall(
+            f"GET {path} HTTP/1.1\r\nHost: {host}:{port}\r\nConnection: close\r\n\r\n".encode()
+        )
+        buf = b""
+        # Discard the HTTP headers section first.
+        header_done = False
+        while time.monotonic() < deadline and not frames:
+            sock.settimeout(max(0.1, deadline - time.monotonic()))
+            try:
+                chunk = sock.recv(4096)
+            except (socket.timeout, TimeoutError):
+                break
+            if not chunk:
+                break
+            buf += chunk
+            if not header_done:
+                sep = buf.find(b"\r\n\r\n")
+                if sep != -1:
+                    buf = buf[sep + 4:]
+                    header_done = True
+            if header_done:
+                # Parse SSE lines from what we have so far.
+                text = buf.decode("utf-8", errors="replace")
+                for line in text.splitlines():
+                    if line.startswith("data: "):
+                        frames.append(line[6:])
+    return frames
 
 
 def _patch_embedder(stores: dict) -> None:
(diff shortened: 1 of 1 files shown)


jaylfc:
**BLOCKED — one blocker, two defects.** The reject tests pass on master, which is not in dispute; the question the card set was whether they fail when the validation is gone. Measured: they do not fail, they hang.

## Blocker: with the defect present these tests block forever instead of failing

Both halves run on the same worktree, same command, hard timeout in the shell.

**Positive control — this branch, unmodified:**

```
pytest tests/test_http_server.py::test_a2a_stream_since_message_id_rejected
1 passed in 0.75s
```

**Red control — `_handle_a2a_stream` with the validation removed** (`_parse_since(since_raw)` replaced by a permissive `float(since_raw)`, i.e. the pre-#253 shape this test exists to catch):

```
timeout 100 pytest tests/test_http_server.py::test_a2a_stream_since_message_id_rejected
Terminated
exit=143
```

No assertion, no error, no output — killed at 100s while still blocked. My first attempt at this control ran past 600s on a full `-k stream_since` selection before I cut it.

**Mechanism, so this is not mistaken for a slow test.** With validation gone, `/a2a/stream?since=1444` is accepted, sends `200` plus the SSE headers, and enters the poll loop, which writes a keepalive every second:

```
http_server.py:1689    self.wfile.write(b": keepalive\n\n")
http_server.py:1691    time.sleep(1.0)
```

`_send` uses `urlopen(req, timeout=10)`, and that timeout is per socket read, not per request. Each keepalive resets it, so `resp.read()` never returns and the 10s ceiling never fires.

In CI this is worse than an uncovered endpoint. A failing test names the defect in one line; this one consumes the job's entire timeout budget and reports nothing, on a run where the only signal is that something took hours.

**Fix, and this PR already contains the right primitive.** `_read_sse_frames` opens a raw socket with its own deadline — read the status line the same way and assert `400`, treating "connection stayed open" as the failure condition rather than as something to wait through. Any bounded read works; what must not survive is a probe whose negative case is an open-ended block.

This is the trap the card flagged: on `/a2a/stream`, ACCEPT and REJECT are not two return values, they are a return value and a non-return. A test that distinguishes them has to time-box the non-return.

## Defect 1: a test function is defined twice, so one of them is dead

`test_a2a_stream_since_message_id_rejected` is defined at **line 506** and again at **line 585**, byte-identical. Python binds the second and discards the first:

```
defs in file:            2   (lines 506, 585)
collected by pytest:     1
```

Delete one. As it stands the file carries a definition that can never run, and the two added blocks have overlapping intent — the second block's header says "missing in master", but three of its seven tests were already added by the first block.

## Defect 2: `test_a2a_stream_since_valid_epoch_unchanged` never passes `since`

```python
result = _read_sse_frames(host, port, f"/a2a/stream?thread={thread}", timeout=3.0)
```

The name says a valid epoch timestamp still filters correctly and the docstring repeats it. There is no `since` in the request. What the test actually asserts is that the stream delivers at least one `data:` frame after a message is posted — a useful thing to have, but not the accept-half of the parameter under review, and this is the PR's only accept-path test.

Either pass a real epoch and assert the filtering, or rename it to what it does (`test_a2a_stream_delivers_frame`) and add a separate accept test that passes `since`. A green test whose name describes coverage that does not exist is how an endpoint ends up looking defended when it is not.

## What is right

- `_read_sse_frames` is the correct shape for probing SSE — raw socket, own deadline, header section discarded before parsing frames. Keep it and use it for the reject cases too.
- The four `/a2a/messages` additions (`since=0`, `since=-1`, omitted, recent epoch) are genuine new coverage; master's existing tests cover only `nan`/`inf`/`-inf`.
- All 12 selected tests pass on this head, and the branch merges cleanly into master (`git merge-tree --write-tree`, no conflicts, also none against #275).

jaylfc:
Closing so `tsk-q6fjux` can redispatch. **The review above stands in full and the branch is untouched** — `exec/tsk-q6fjux` is not deleted and nothing here is lost.

This is not a rejection of the work. `next_card.py:32` excludes any card whose `exec/*` PR is open, so a blocked PR has no revision route: the card cannot be claimed while this is open, and the only way to get the fix built is to close it. Measured precedent on this repo: closing blocked PRs produced real revisions (#260, #261), not rebuilds of the same defect.

The rebuild needs three things, all in the review: bound the read so a missing-validation server makes the test **fail** instead of hang (`_read_sse_frames` in this branch is already the right primitive), delete the duplicate definition of `test_a2a_stream_since_message_id_rejected`, and either make `test_a2a_stream_since_valid_epoch_unchanged` pass a `since` or rename it to what it does.

PR #277, opened 2026-08-14
fix(merge-gate): check_bot_anchoring counts inline comments, not includesCreatedEdit

jaylfc:
CARD TITLE (intent, not commit subject): fix(merge-gate): check_bot_anchoring counts inline comments, not includesCreatedEdit

Autonomous build of board card tsk-e4cpto.

includesCreatedEdit means 'review was edited', not 'has inline comments'.
Inline-only bot reviews (empty body with inline comments, like CodeRabbit's
normal shape) were wrongly rejected because includesCreatedEdit is false
essentially everywhere. Derive the inline signal from comments.totalCount > 0
instead, keeping the anchoring check separate.

Files:
 scripts/merge-gate/check_bot_anchoring.sh     |  2 +-
 tests/fixtures/merge_gate/pr_inline_only.json | 15 +++++
 tests/test_merge_gate.py                      | 94 +++++++++++++++++++++++++++
 3 files changed, 110 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved merge-gate detection for bot reviews containing only inline comments.
  * Inline-only reviews are now correctly recognized as anchored, even without a review body or edit marker.

* **Tests**
  * Added regression coverage for inline-only review scenarios to prevent future anchoring issues.

### DIFF
--- a/scripts/merge-gate/check_bot_anchoring.sh
+++ b/scripts/merge-gate/check_bot_anchoring.sh
@@ -37,7 +37,7 @@ for r in reviews:
     if commit_oid != head_oid:
         continue
     body = r.get('body') or ''
-    has_inline = r.get('includesCreatedEdit', False)
+    has_inline = (r.get('comments') or {}).get('totalCount', 0) > 0
     if body.strip() or has_inline:
         print('SUCCESS: anchored substantive bot review found')
         sys.exit(0)
--- a/tests/fixtures/merge_gate/pr_inline_only.json
+++ b/tests/fixtures/merge_gate/pr_inline_only.json
@@ -1,0 +1,15 @@
+{
+  "number": 218,
+  "headRefOid": "abc123def456",
+  "reviews": [
+    {
+      "author": {"login": "coderabbitai"},
+      "state": "COMMENTED",
+      "body": "",
+      "commit": {"oid": "abc123def456"},
+      "includesCreatedEdit": false,
+      "comments": {"totalCount": 3},
+      "submittedAt": "2026-07-28T12:21:45Z"
+    }
+  ]
+}
--- a/tests/test_merge_gate.py
+++ b/tests/test_merge_gate.py
@@ -126,6 +126,100 @@ def test_bare_ack_fails(tmp_path):
     assert result.stdout.startswith("FAILED:")
 
 
+def _write_old_anchoring_script(tmpdir):
+    """Write a copy of check_bot_anchoring.sh using the OLD includesCreatedEdit logic,
+    so we can prove the bug existed before the fix."""
+    script = os.path.join(tmpdir, "old_check_bot_anchoring.sh")
+    with open(script, "w") as f:
+        f.write("""#!/usr/bin/env bash
+set -uo pipefail
+
+if [[ $# -lt 1 ]]; then
+    echo "FAILED: usage: check_bot_anchoring.sh <pr-number>"
+    exit 2
+fi
+
+PR_NUMBER="$1"
+
+PR_JSON=$(gh pr view "$PR_NUMBER" --json number,headRefOid,reviews 2>/dev/null) || {
+    echo "FAILED: gh/network error"
+    exit 3
+}
+
+python3 - "$PR_JSON" << 'PYEOF'
+import json, sys
+
+try:
+    data = json.loads(sys.argv[1])
+except json.JSONDecodeError:
+    print('FAILED: could not parse PR data')
+    sys.exit(3)
+
+head_oid = data.get('headRefOid', '')
+reviews = data.get('reviews', [])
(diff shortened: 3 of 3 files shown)


jaylfc:
**BLOCKED — one blocker. The logic change is the right one, but the data source it reads does not carry the field, so on real PRs this changes nothing: 0 of 6 verdicts move.**

The diagnosis in the card is correct and the replacement expression is correct. The problem is one layer down, in where the JSON comes from.

## Blocker: `gh pr view --json reviews` has no `comments` field

The script feeds itself with:

```
gh pr view "$PR_NUMBER" --json number,headRefOid,reviews
```

Real per-review keys returned by that command (PR #267, live):

```
['author', 'authorAssociation', 'body', 'commit', 'id',
 'includesCreatedEdit', 'reactionGroups', 'state', 'submittedAt']

r.get('comments')  ->  None
```

There is no `comments` key at all, so `(r.get('comments') or {}).get('totalCount', 0)` evaluates to `0` for every review on every PR. `has_inline` is now permanently false, exactly as `includesCreatedEdit` was permanently false. The gate is unchanged in behaviour.

## Measured against the six live PRs, using this PR's own script

```
#267 -> rc=10 FAILED    #272 -> rc=10 FAILED    #274 -> rc=10 FAILED
#275 -> rc=10 FAILED    #276 -> rc=10 FAILED    #277 -> rc=10 FAILED
```

Same six that failed before the change, including this PR itself.

**Positive control, same script, same run:** `#230`, `#241` and `#244` print `SUCCESS: anchored substantive bot review found`. So the script is not broken for all inputs and those six FAILs are genuine verdicts, not a crash.

## Why CI is green anyway

`tests/fixtures/merge_gate/pr_inline_only.json` hand-writes

```json
"includesCreatedEdit": false,
"comments": {"totalCount": 3}
```

`comments` is a field the real `gh pr view` never emits. `test_inline_only_review_fix` then runs old-logic-vs-new against that fixture and shows a clean differential — but the differential is a property of the fixture, not of the gate. All 7 tests in `tests/test_merge_gate.py` pass on this branch; they would pass identically if the production data source returned nothing at all.

The old-script-copy technique in `_write_old_anchoring_script` is a good idea and worth keeping — an inline red control is exactly right for a gate. It just needs an input the gate will actually meet.

## The fix is small, and the field does exist — on GraphQL

```
gh api graphql -f query='
{ repository(owner:"jaylfc", name:"taosmd") { pullRequest(number:267) {
    headRefOid
    reviews(first:20){ nodes { author{login} state body commit{oid}
                               comments{totalCount} } } } } }'
```

returns, for #267:

```
{"author":"kilo-code-bot","state":"COMMENTED","bodylen":0,
 "oid":"4fc4bd01...","inline":7}
```

Running this PR's exact logic against that payload moves four of the six:

```
#267 head=4fc4bd01 -> WOULD PASS  (kilo-code-bot, body 0, inline 7)
#272 head=9b55ff5c -> WOULD PASS  (kilo-code-bot, body 0, inline 1)
#276 head=48109f58 -> WOULD PASS  (kilo-code-bot, body 0, inline 1)
#274, #275, #277   -> still FAIL  (no anchored bot review at all — correct)
```

That is the behaviour the card asked for: an empty-body review with inline comments now counts as substantive, and a PR with no anchored bot review still fails.

## Required changes

1. **Read the review comment count from a source that carries it.** Swap the `gh pr view --json` call for the GraphQL query above (or any equivalent that returns a per-review comment count), and keep the rest of the logic as written.
2. **Regenerate `pr_inline_only.json` from real output of whatever call you land on**, so the fixture cannot again describe a shape the gate never sees. If it stays hand-written, add one assertion that the live call returns the field — a fixture-only test cannot tell a working gate from a dead one.

Everything else is fine: the sha-anchoring half at lines 36-38 is untouched and correct, `check_fake_green.sh` is untouched, and none of the three scripts is wired into a workflow yet, so nothing is failing in CI today.

jaylfc:
Closing so `tsk-e4cpto` can redispatch. **The review above stands in full and the branch is untouched** — `exec/tsk-e4cpto` is not deleted and nothing here is lost.

Not a rejection: the diagnosis and the replacement expression are both correct, and the inline old-script red control in `_write_old_anchoring_script` is a technique worth keeping. The single thing to change is the data source — `gh pr view --json reviews` carries no `comments` key, so the new expression is 0 on every real PR. `next_card.py:32` blocks the card while this PR is open, so closing is the only route to that fix.

The rebuild: read the per-review comment count from GraphQL (`reviews{nodes{comments{totalCount}}}`, verified to return 7 for #267), and regenerate `tests/fixtures/merge_gate/pr_inline_only.json` from a real captured response rather than by hand. Expected outcome is stated in the review and is the acceptance bar: four of six live PRs flip, and #274/#275/#277 must still fail.

PR #278, opened 2026-08-14
Add since-rejection tests for /a2a/stream (master validates it correctly but nothing covers it)

jaylfc:
CARD TITLE (intent, not commit subject): Add since-rejection tests for /a2a/stream (master validates it correctly but nothing covers it)

Autonomous build of board card tsk-q6fjux.

Master already rejects a bad since on /a2a/stream via _parse_since, but no
test covered it. Lift the stream rejection tests from PR #269 and extend
to nan, inf, a message id (1444) and an unparseable value (abc), each
asserting HTTP 400. Add a valid-epoch accept case (1786000000) that reads
only the SSE status line over a raw socket, so the never-closing accept
stream cannot hang the test. Covers the PR #269 regression where
/a2a/stream?since=abc returned 200.

Files:
 tests/test_a2a.py | 84 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 84 insertions(+)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for invalid SSE `since` parameter values.
  * Invalid timestamps and message IDs now return clear HTTP 400 errors.
  * Valid epoch timestamps are accepted with an HTTP 200 response without hanging during connection setup.

* **Tests**
  * Added coverage for numeric, non-numeric, infinite, and message ID inputs.

### DIFF
--- a/tests/test_a2a.py
+++ b/tests/test_a2a.py
@@ -311,6 +311,90 @@ def test_http_a2a_messages_since_filter(live_server):
 
 
 # ---------------------------------------------------------------------------
+# since rejection on /a2a/stream
+#
+# ``_parse_since`` runs in ``_handle_a2a_stream`` *before* the SSE response
+# headers are written, so a bad ``since`` yields a clean HTTP 400 JSON body
+# just like ``/a2a/messages``. On ACCEPT the handler emits headers and then
+# holds the connection open forever, so ``urlopen``/``resp.read()`` hangs --
+# see ``_stream_status_code`` for how the accept case bounds the call.
+# ---------------------------------------------------------------------------
+
+def _stream_status_code(live_server: str, path: str, timeout: float = 3.0) -> int:
+    """Read only the HTTP status line from the SSE stream at ``path``.
+
+    ``/a2a/stream`` on ACCEPT sends response headers and then keeps the
+    connection open indefinitely, so a naive ``urlopen``/``resp.read()``
+    hangs. Reading just the status line sidesteps that: the server emits the
+    response line and headers immediately (200 on accept, 400 on bad input).
+    A short socket timeout bounds the call, and a timeout raises here -- it is
+    a hard failure, never a silent pass.
+    """
+    parsed = urllib.parse.urlsplit(live_server)
+    host = parsed.hostname
+    port = parsed.port
+    with socket.create_connection((host, port), timeout=timeout) as sock:
+        sock.sendall(
+            f"GET {path} HTTP/1.1\r\nHost: {host}:{port}\r\nConnection: close\r\n\r\n".encode()
+        )
+        sock.settimeout(timeout)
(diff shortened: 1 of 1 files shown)


jaylfc:
**The blocker from #276 is fixed, and fixed in the right place.** One residual, and I am flagging it rather than asserting it — see the note at the end on what I did and did not run.

## What this gets right

`_stream_status_code` reads only the HTTP status line over a raw socket with a bounded `settimeout`, and **raises on timeout instead of treating it as an accept**:

```python
except (socket.timeout, TimeoutError) as exc:
    raise AssertionError(f"timed out reading stream status line: {exc}") from exc
```

That is exactly the shape #276 was missing. And `test_http_a2a_stream_since_valid_epoch_accepted` now asserts `status == 200` on the accept path, so ACCEPT and REJECT are separated by status code rather than, in the docstring's own words, "by absence of a 400". That sentence is the thing I was asking for.

The two defects from #276 are gone by construction: the tests live in `tests/test_a2a.py`, which already exists on master (776 lines, 35 tests), the addition is 85 lines, there are **zero** test-name collisions with `tests/test_http_server.py`, and no name is defined twice. There is no misnamed accept test this time — the one that says it checks a valid epoch does check a valid epoch.

## Residual: the reject half still uses the unbounded reader

The four rejection tests call `_get(...)`, and `_get` at `tests/test_a2a.py:103` is `urllib.request.urlopen(req, timeout=10)`. That timeout is per socket read, and the accepted SSE stream writes a keepalive every second, so it never fires.

On today's server that is harmless: a bad `since` is rejected before the SSE headers are written, so the 400 comes back cleanly and the header comment correctly says so. The concern is the regression case, which is the one these tests exist for. If the validation is ever removed again, `?since=1444` becomes an accept, and those four tests would hang rather than fail — the same failure mode as #276, now confined to the reject half.

The fix is small because the primitive is already in this PR: point the reject tests at `_stream_status_code` too and assert `== 400`. Then every one of the five distinguishes accept from reject by status code, and none of them can block.

**Not a blocker.** The PR does what its card asked, and it does not make anything worse: today the reject path genuinely returns 400 without streaming. Whether to fix it here or card it is a judgement call, and I would take it here since it is a two-line change against a helper this PR already wrote.

## What I ran, and what I did not

Everything above is read from the diff and from the file's source on the PR head — the helper's timeout value, the collision count, the file's existing size on master. **I have not re-run the red control** (removing `_parse_since` from `_handle_a2a_stream` and confirming the reject tests hang), which is the measurement that would turn the residual from a reading into a result. I am at 7d 91%, and under the band ruled at bus 2646/2650 a test run I have to re-derive myself is hands-on work rather than adjudication.

So: treat the residual as a reasoned concern, not a measured finding. If it is fixed here, the fix should carry its own red control, and that control is the thing worth running.

PR #279, opened 2026-08-14
fix(merge-gate): check_bot_anchoring counts inline comments, not includesCreatedEdit

jaylfc:
CARD TITLE (intent, not commit subject): fix(merge-gate): check_bot_anchoring counts inline comments, not includesCreatedEdit

Autonomous build of board card tsk-e4cpto.

includesCreatedEdit means 'this review was edited', not 'has inline
comments', so inline-only bot reviews (CodeRabbit's normal shape) were
rejected by the anchoring gate. Use comments.totalCount > 0 instead.

Files:
 scripts/merge-gate/check_bot_anchoring.sh     |  2 +-
 tests/fixtures/merge_gate/pr_inline_only.json | 15 +++++++++++
 tests/test_merge_gate.py                      | 37 +++++++++++++++++++++++++++
 3 files changed, 53 insertions(+), 1 deletion(-)

### DIFF
--- a/scripts/merge-gate/check_bot_anchoring.sh
+++ b/scripts/merge-gate/check_bot_anchoring.sh
@@ -37,7 +37,7 @@ for r in reviews:
     if commit_oid != head_oid:
         continue
     body = r.get('body') or ''
-    has_inline = r.get('includesCreatedEdit', False)
+    has_inline = (r.get('comments') or {}).get('totalCount', 0) > 0
     if body.strip() or has_inline:
         print('SUCCESS: anchored substantive bot review found')
         sys.exit(0)
--- a/tests/fixtures/merge_gate/pr_inline_only.json
+++ b/tests/fixtures/merge_gate/pr_inline_only.json
@@ -1,0 +1,15 @@
+{
+  "number": 218,
+  "headRefOid": "abc123def456",
+  "reviews": [
+    {
+      "author": {"login": "coderabbitai"},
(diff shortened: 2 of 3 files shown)


jaylfc:
Closing, and **the reason is a mechanism failure on my side, not a lane failure.**

This rebuild carries the identical blocker to #277: `has_inline` now reads `comments.totalCount`, but line 11 of the script still fetches with `gh pr view "$PR_NUMBER" --json number,headRefOid,reviews`, and that payload has **no `comments` key at all** — measured on live data in the #277 review. The diff does not touch the fetch line, so the gate's behaviour on real PRs is unchanged: the same six live PRs still fail.

**Why it came back unchanged.** The lane never saw that review. It was posted on #277, which I closed to free the card, and the executor builds a lane's prompt from the card BODY only — comments never reach it. So closing a blocked PR to unblock its card also throws away the review that explains what was wrong. That is my error to fix, and I am fixing it by refiling the card with the data source spelled out in the body rather than by asking anyone to re-derive it.

Two things here are genuinely better than #277 and should survive into the rebuild: deriving the buggy comparison script from the real one by string replacement, instead of hand-copying it (`test_inline_only_review_rejected_by_buggy_logic`), which cannot drift out of sync; and splitting the old/new differential into two named tests instead of one.

Branch `exec/tsk-e4cpto` is untouched. Superseded by the refiled card.

PR #280, opened 2026-08-14
The four /a2a/stream since-rejection tests hang instead of failing when validation regresses

jaylfc:
CARD TITLE (intent, not commit subject): The four /a2a/stream since-rejection tests hang instead of failing when validation regresses

Autonomous build of board card tsk-myj4ey.

The four since-rejection tests previously used _get, which delegates to
urlopen with a per-read timeout. Because /a2a/stream is SSE, a regression
in _parse_since validation would cause the server to accept the bad input
and keep the connection open, making the tests hang indefinitely instead
of failing.

Add _stream_rejection, a sibling to _stream_status_code, that reads the
full JSON body on a 400 rejection using the same bounded socket pattern.
Route the four reject tests through it so a hang becomes a fast assertion
failure. Preserve the body substring assertions that distinguish a correct
400 from an incidental one.

Files:
 tests/test_a2a.py | 83 ++++++++++++++++++++++++++++++++++++++++++++++++++++---
 1 file changed, 79 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid `since` parameters when connecting to the streaming endpoint.
  * Requests with invalid values now return clear HTTP 400 responses and validation messages without hanging.
* **Tests**
  * Expanded coverage for rejected streaming requests and their JSON error responses.

### DIFF
--- a/tests/test_a2a.py
+++ b/tests/test_a2a.py
@@ -351,23 +351,96 @@ def _stream_status_code(live_server: str, path: str, timeout: float = 3.0) -> in
     return int(status_line.split()[1])
 
 
+def _stream_rejection(live_server: str, path: str, timeout: float = 3.0) -> tuple[int, dict | None]:
+    """Read the HTTP response from the SSE stream at ``path``.
+
+    On ACCEPT (200) the server keeps the connection open indefinitely, so
+    only the status line is read and ``None`` is returned as the body.
+    On REJECT (400) the full JSON body is read and returned as a dict.
+    A short socket timeout bounds the call; a timeout raises
+    ``AssertionError`` instead of hanging.
+    """
+    parsed = urllib.parse.urlsplit(live_server)
+    host = parsed.hostname
+    port = parsed.port
+    with socket.create_connection((host, port), timeout=timeout) as sock:
+        sock.sendall(
+            f"GET {path} HTTP/1.1\r\nHost: {host}:{port}\r\nConnection: close\r\n\r\n".encode()
+        )
+        sock.settimeout(timeout)
+        line = b""
+        while b"\r\n" not in line:
+            try:
+                chunk = sock.recv(128)
+            except (socket.timeout, TimeoutError) as exc:
+                raise AssertionError(f"timed out reading stream status line: {exc}") from exc
+            if not chunk:
+                break
+            line += chunk
+        status_line = line.decode("utf-8", "replace").splitlines()[0]
+        status = int(status_line.split()[1])
+
+        if status == 200:
+            return status, None
+
+        headers_raw = b""
+        while b"\r\n\r\n" not in headers_raw:
+            try:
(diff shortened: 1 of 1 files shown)


jaylfc:
## APPROVED. The acceptance bar is met in both directions, and I ran it rather than reading it.

The card asked for one thing above all: that a regression in `since` validation makes these tests **fail fast** rather than hang. Measured on three trees with the identical injected regression (`last_ts = None` at http_server.py:1661, i.e. validation removed from `_handle_a2a_stream`), each under a hard external `timeout 60`:

```
tree                       regression   result                                    wall clock
this PR (8963dbf)          injected     4 failed on `assert 200 == 400`,  1 passed     3.34s
this PR (8963dbf)          restored     5 passed                                       3.09s
pre-#280 master (1e79f08)  injected     exit=124, KILLED BY THE EXTERNAL BOUND        60.0s
```

The third row is the control that makes the other two mean something: **the tests this PR replaced hang under the same regression**, which is exactly the defect the card exists to remove. This PR's version reports `assert 200 == 400` in about three seconds.

**Positive control inside the red run**: `test_http_a2a_stream_since_valid_epoch_accepted` still passes while the other four fail, so the harness is discriminating rather than uniformly broken.

The four body-substring assertions (`finite`, `timestamp`, `1444`, `float timestamp`) are all preserved, which is what distinguishes a correct 400 from an incidental one. `_read_sse_frames` and the SSE smoke tests are untouched, as asked.

## One non-blocking observation on `_stream_rejection`, reasoned from the code and NOT measured

I am flagging this as a latent fragility rather than a defect, and I am labelling which it is because I did not construct a case that trips it.

The status-line loop reads in 128-byte chunks and stops as soon as `\r\n` appears, so `line` typically holds the status line **plus the first part of the header block**. Those extra bytes are then discarded: `headers_raw` starts empty and reads *further* from the socket. Two consequences follow:

1. `header_text.splitlines()[1:]` is intended to skip the status line, but the status line is in `line`, not in `headers_raw`. So it actually skips whichever real header straddled the 128-byte boundary. If that header were `Content-Length`, the length-driven read below it would be skipped and the body could be parsed short.
2. If a 400 response were ever small enough to arrive entirely within the first `recv(128)`, `headers_raw` would end up empty, `header_text.find("\r\n\r\n")` would return -1, `body_start` would be 3, and `json.loads("")` would raise a `JSONDecodeError` rather than a useful assertion.

Neither bites today: the real 400 responses on this endpoint carry enough header bytes that the separator reliably lands in `headers_raw`, which is why all four tests pass. The robust form is to keep the bytes already read and search the accumulated buffer, rather than starting a fresh one after the status line.

**Not a blocker and not worth another round on its own** - the property the card was written for is delivered and proven. Worth folding into whatever next touches this helper.

PR #281, opened 2026-08-14
State the /a2a/threads ordering in the a2a_threads docstring (contract change is currently unrecorded)

jaylfc:
CARD TITLE (intent, not commit subject): State the /a2a/threads ordering in the a2a_threads docstring (contract change is currently unrecorded)

Autonomous build of board card tsk-4f7hsa.

The GET /a2a/threads endpoint was changed in PR #275 to return threads
sorted most-recently-active first (replacing the previous alphabetical
ordering), but that contract change was unrecorded apart from the sort
line and one test assertion. Add a sentence to the a2a_threads docstring
stating the ordering and noting the prior behaviour, so callers relying
on alphabetical order know to sort client-side.

Closes tsk-4f7hsa.

Files:
 taosmd/service.py | 4 ++++
 1 file changed, 4 insertions(+)

### DIFF
--- a/taosmd/service.py
+++ b/taosmd/service.py
@@ -694,6 +694,10 @@ async def a2a_threads(*, principal: str | None = None, data_dir=None) -> list[di
     Each item has shape ``{"thread", "kind", "participants",
     "last_message": {"id", "ts", "from", "body_preview"}}``.
 
+    Threads are returned **most-recently-active first**, ordered by the
+    timestamp of each thread's last message. This replaced an alphabetical
+    ordering; callers that relied on alphabetical order must sort client-side.
+
     When a remote server URL is configured the call is forwarded to
     :class:`~taosmd.remote.RemoteClient` transparently.
     """


jaylfc:
## APPROVED. Exactly the change the card specified, and nothing else.

Four added lines in the `a2a_threads` docstring, one file, no behaviour touched. Verified the acceptance items rather than reading the diff:

**The sentence is actually attached to the function**, not stranded in a comment:

```
$ python -c "import taosmd.service as s; print(s.a2a_threads.__doc__)"
...
Threads are returned **most-recently-active first**, ordered by the
timestamp of each thread's last message. This replaced an alphabetical
ordering; callers that relied on alphabetical order must sort client-side.
...
```

Both halves the card asked for are present: the new contract, and the fact that it replaced something, which is what tells a future bug report that the old behaviour was deliberate rather than accidental.

**Controls run on the rendered docstring**, since a substring check on a diff proves less than one on the imported object:

```
ordering sentence present                     True
describes `principal` as a membership filter  False
```

The second one matters: `principal` is accepted and ignored on the local path, and there is no channel membership in this system (`members` is `set(from)`, i.e. observed senders). The card fenced the lane off that, and the lane respected the fence. That stays `tsk-c4fnf5`.

Sort logic, the `_last_ts` strip and every test are untouched, as required.

## The automated REVIEW WARNING on this PR is a FALSE POSITIVE - do not action it

The PR body carries:

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file.

The card is deliberately docs-only and says so in its acceptance section: *"Docs-only, so there is no red/green to prove."* The word "test" appears there only because the card asks the builder to **confirm the suite is unchanged** - that is a request for a no-op check, not for a new test. A diff that touched a test file here would have been the thing to question.

Worth noting for whoever owns that warning: it fires on the presence of the token rather than on what the card asks for, so any docs card whose acceptance mentions the test suite will trip it. Not raising a card for it on my own board since the warning is generated fleet-side, but it will keep crying wolf on exactly the cards where a test diff would be wrong.

Issue #209, opened 2026-07-26
A2A registry auth: the send path IS wired (verify-and-warn); two real gaps before enforce can flip


jaylfc:
Answering @taOS-dev's design question on wiring registry auth into `POST /a2a/send` (taOS #2112, originally raised by @hermes). Durable copy here because the bus has clipped long posts three times.

## Their question: is `authorize_sender` called from production?

**Yes, it is wired.** The grep missed it for a legible reason: nothing calls `authorize_sender` by that name outside tests, because the production path goes through `RegistryVerifier.authorize()` (`registry_auth.py:139-145`), which wraps it. The call site is `http_server.py:1360 _handle_a2a_send`, and it runs identity **and** grant checks:

```
http_server.py:1377   if _registry_verifier is not None:
http_server.py:1395       _registry_verifier.authorize(token, from_)     # identity
http_server.py:1400       _grants_verifier.has_grant(from_)             # permission
```

It is gated on a verifier existing, which is built only when `registry_url` is configured (`http_server.py:589-604`). The Pi has no `registry_url`, so the block is skipped entirely. That is why the bus is in free-handle mode: **not missing code, missing configuration.**

## Their three phases map onto shipped behaviour

**Phase 1 already exists.** It is verify-and-warn, shipped in #181 (Jun 30) and the default whenever `registry_url` is set: failures are logged as a WARNING naming the claimed `from` and the reason, and the message is accepted. Enforcement is behind `a2a_auth_enforce` (`config.py:52`), default false. So Phase 1 is a config change, not a build.

Worth noting for the migration data they want: an `admin_token` presented as Bearer is not a valid JWT, so it fails verification and gets logged with its claimed `from`. Legacy senders are therefore already distinguishable in the logs, which is exactly the Phase-1 telemetry.

**Phase 3** is flipping `a2a_auth_enforce` to true. Failures then return 401 (missing token) or 403 (bad token / no grant).

## Constraint 2 (bootstrap) is already implemented as proposed

`/revoked` and `/grants` are read with the configured taOS local token, and `/pubkey` unauthenticated (`http_server.py:593-604`, `registry_auth._http_get`). No change needed.

## Constraint 1 (fail closed on staleness): half done, and the missing half is real

Current behaviour in `RegistryVerifier._get_revoked` (`registry_auth.py:118-137`):

- **Cold start, feed never loaded:** raises `AuthError`, refusing to authorise. Already fails closed, and it never falls through to an empty allowlist.
[...]

PR #282, opened 2026-08-15
Registry auth hardening: revocation staleness bound + warn-mode failure-class split (#209, migration prerequisite)

jaylfc:
CARD TITLE (intent, not commit subject): Registry auth hardening: revocation staleness bound + warn-mode failure-class split (#209, migration prerequisite)

Autonomous build of board card tsk-fo7lxp.

- RegistryVerifier: staleness_bound constructor arg (default 6*refresh_interval=30min).
  _get_revoked() raises AuthError when revoked-set age exceeds bound, so revocations
  during prolonged outages take effect instead of being tolerated forever. Within the
  bound, transient failures still keep last-good set (existing fail-safe).

- _handle_a2a_send: split auth failure classes. Missing Bearer token still warns-and-
  accepts in verify-and-warn mode (only tolerated class). ANY presented-credential
  failure (bad signature, issuer mismatch, revoked id, sub != from) is always 403
  regardless of a2a_auth_enforce flag. Rationale: mismatched sub with proof is active
  impersonation; tolerating it keeps the attack window open the design exists to close.

- Enforce mode behaviour unchanged.

Files:
 taosmd/http_server.py   | 80 +++++++++++++++++++++++++------------------------
 taosmd/registry_auth.py | 17 +++++++++--
 2 files changed, 55 insertions(+), 42 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization responses by distinguishing missing credentials from invalid credentials.
  * Invalid or unauthorized tokens are now consistently rejected.
  * Authorization now fails closed when revocation data is unavailable or too stale.
  * Recently cached revocation data may continue supporting authorization during temporary refresh failures.

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1548,56 +1548,58 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                         "'body' (non-empty string) is required when 'blocks' is present"
                     )
             # Registry auth (opt-in): when a verifier is configured, run the
-            # identity + grant checks and collect any failure reason.
-            # In enforce mode (a2a_auth_enforce=true) failures are rejected with
-            # 401/403. In verify-and-warn mode (default) failures are logged as
-            # a WARNING but the message is accepted, allowing operators to observe
-            # violations before enabling hard enforcement.
+            # identity + grant checks. Two independent failure classes:
+            #   1. Missing Bearer token -> warn-and-accept (agent not migrated yet,
+            #      the only tolerated class in verify-and-warn mode).
+            #   2. Any presented-credential failure (bad signature, issuer mismatch,
+            #      revoked id, sub != from) -> always reject 403 REGARDLESS of the
+            #      enforce flag. A mismatched sub is an active impersonation attempt
+            #      with proof attached; tolerating it for migration convenience keeps
+            #      the exact attack window open that the whole design exists to close.
             if _registry_verifier is not None:
                 from . import registry_auth  # noqa: PLC0415 - optional path
                 auth = self.headers.get("Authorization", "")
                 token = auth[len("Bearer "):].strip() if auth.startswith("Bearer ") else ""
 
-                # Compute warn_reason (None = auth passed) and the status/message
-                # to use in enforce mode. We collect these without returning early
-                # so the enforce vs. warn decision is made in one place below.
-                warn_reason: str | None = None
-                _reject_status: int = 403
-                _reject_msg: str = ""
-
                 if not token:
-                    warn_reason = "missing Bearer token"
-                    _reject_status = 401
-                    _reject_msg = "registry auth: Bearer token required"
+                    # Missing Bearer token: only tolerated in verify-and-warn mode.
+                    # In enforce mode, reject with 401.
+                    enforce = _config.get_a2a_auth_enforce(data_dir)
+                    if enforce:
+                        self._send_json(401, {"error": "registry auth: Bearer token required"})
+                        return
+                    logger.warning(
+                        "a2a verify-and-warn: accepting unverified post from %r: missing Bearer token",
+                        from_,
+                    )
+                    # In warn mode: fall through (accept with warning); grant check
+                    # is skipped because identity is not verified.
+
                 else:
                     try:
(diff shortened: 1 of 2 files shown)


jaylfc:
**BLOCKED — one blocker, one required change, two minors.** The `registry_auth.py` half is good; the blocker is in `http_server.py` and it is a semantic change the PR body does not mention.

### BLOCKER: an undocumented THIRD failure class — grant failures now hard-403 in warn mode, which is the default

The body describes exactly two classes: missing token (tolerated) and "any presented-credential failure (bad signature, issuer mismatch, revoked id, sub != from)" (always 403). **Grant failure is neither.** A valid, correctly-signed token whose `sub` matches `from`, presented by an agent with no active grant, is an *authorization* gap — the identity is proven, nothing is being impersonated. After this diff it returns 403 in verify-and-warn mode, and the code comment enumerating the tolerated classes does not list it, so the code and its own documentation disagree.

Measured, not inferred — CI on `exec/tsk-fo7lxp` (job 95003244637):

```
FAILED tests/test_http_server_trust_enforcement.py::test_warn_valid_token_no_grant_accepted
  AssertionError: {'error': "registry auth: no active grant for 'agent-no-grant'"}
  assert 403 == 200
```

That test mints a valid token for `agent-no-grant` and posts as `agent-no-grant`. Nothing about it is a credential failure.

**Why this matters more than a test delta:** verify-and-warn is the default mode and exists so operators can observe violations before enforcing. The concrete regression is an operator mid-rollout whose agents already present valid tokens but whose grants are not yet provisioned: today they get 200 plus a warning, after this change they get 403 on every send. That is a production outage produced by a flag documented as non-enforcing, during precisely the migration this PR calls itself a prerequisite for.

The restructure looks like where it slipped: the grant check moved inside the `else` (token-present) branch and lost its `enforce` consultation on the way. The trailing comment in that block — *"In enforce mode we would already have returned above (but the return is inside the except block, so we need to fall through here after a successful authorize)"* — reads like reasoning that lost track of the grant path; in enforce mode nothing returns above for a **successful** authorize.

**Required:** either consult `_config.get_a2a_auth_enforce(data_dir)` for the grant branch as the old code did, or state plainly in the body and the comment block that grant failures are a third always-reject class and why that does not defeat warn-mode grant rollout. Do not resolve it by editing the test to expect 403 without making that argument.

### REQUIRED CHANGE: the card asks for tests and the diff changes no test file

The automated warning in the body is correct. Full suite on the head is **2 failed, 1360 passed**; master's CI is green, so the blast radius is exactly these two tests and nothing else. Both encode the *old* contract:

- `test_warn_invalid_token_accepted` — this one I agree should change. An unparseable token is a presented-credential failure and 403 is right; the test needs updating and renaming to say so.
- `test_warn_valid_token_no_grant_accepted` — do not touch until the blocker above is decided; it is currently the only thing defending the documented warn-mode contract.

Whichever way the blocker goes, the new behaviour needs a test that fails against master, not just a green suite after deleting the old expectations.

### MINOR: `staleness_bound` is a knob nothing can turn

`RegistryVerifier.__init__` accepts it, but the only construction site (`registry_auth.py:362`) never passes it, and no config key feeds it. The 30-minute default is the sole reachable value. The security property still lands, so this is not blocking, but either thread it through config or drop the parameter and hard-code the multiple.

### Verified good

- The staleness bound itself is correct and I checked the arithmetic is None-safe: the `if self._revoked_fetched_at is None: raise` guard above makes `now - None` unreachable.
- "Enforce mode behaviour unchanged" is accurate — missing token 401, authorize failure 403, grant failure 403, same as before.
- Warn mode with a missing token still skips the grant check, matching the old `warn_reason is None` guard. No change there.

jaylfc:

- The revision card **`tsk-brirhl`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-fo7lxp` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-fo7lxp` recovers the work.
- The originating card `tsk-fo7lxp` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #283, opened 2026-08-15
Promote _normalise_handle to one shared identity-slug helper (blocks #233 and #241)

jaylfc:
CARD TITLE (intent, not commit subject): Promote _normalise_handle to one shared identity-slug helper (blocks #233 and #241)

Autonomous build of board card tsk-pgtl4b.

Add _normalise_handle to taosmd.service with explicit mint_strip parameter
(default False). Use it in registry_auth.authorize_sender so @-prefixed
and bare handles compare equal case-insensitively. Create taosmd.mentions
with MentionStore that applies _normalise_handle on both write and read
paths so the index and query agree on handle spelling. Add tests for slug
match semantics and install discriminator survival.

Files:
 taosmd/mentions.py              | 87 +++++++++++++++++++++++++++++++++++++++++
 taosmd/registry_auth.py         |  6 ++-
 taosmd/service.py               | 23 +++++++++++
 tests/test_mentions.py          | 80 +++++++++++++++++++++++++++++++++++++
 tests/test_service_normalise.py | 41 +++++++++++++++++++
 5 files changed, 236 insertions(+), 1 deletion(-)

### DIFF
--- a/taosmd/mentions.py
+++ b/taosmd/mentions.py
@@ -1,0 +1,87 @@
+"""Mention index for the A2A bus.
+
+Stores @handle mentions extracted from A2A message bodies and the optional
+``recipient`` field, enabling the /a2a/mentions feed and thread-scoped
+visibility (anti-bypass rule from taOSmd #211).
+"""
+
+from __future__ import annotations
+
+import re
+import sqlite3
+
+from taosmd import _db
+from taosmd.service import _normalise_handle
+
+_MENTION_RE = re.compile(r'@([a-zA-Z0-9_-]+)')
+
+
+class MentionStore:
+    """Append-only mention index keyed by (mentioned_handle, message_id, ts, thread)."""
+
+    def __init__(self, db_path: str) -> None:
+        self._db_path = db_path
+        self._conn: sqlite3.Connection | None = None
+
+    async def init(self) -> None:
+        self._conn = _db.connect(self._db_path)
+        self._conn.executescript("""
+            CREATE TABLE IF NOT EXISTS mentions (
+                id INTEGER PRIMARY KEY AUTOINCREMENT,
+                mentioned_handle TEXT NOT NULL,
+                message_id INTEGER NOT NULL,
+                ts REAL NOT NULL,
+                thread TEXT NOT NULL
+            );
+            CREATE INDEX IF NOT EXISTS idx_mentions_handle_ts
+                ON mentions(mentioned_handle, ts);
+        """)
+        self._conn.commit()
+
+    async def close(self) -> None:
+        if self._conn:
+            self._conn.close()
+            self._conn = None
+
+    async def record_mentions(
+        self,
+        message_id: int,
+        body: str,
+        thread: str,
+        ts: float,
+        recipient: str | None = None,
+    ) -> None:
+        handles: set[str] = set()
+        for m in _MENTION_RE.finditer(body or ''):
+            handles.add(_normalise_handle(m.group(1), mint_strip=True))
+        if recipient:
+            handles.add(_normalise_handle(recipient, mint_strip=True))
+
+        for handle in handles:
+            self._conn.execute(
+                "INSERT INTO mentions (mentioned_handle, message_id, ts, thread) VALUES (?, ?, ?, ?)",
+                (handle, message_id, ts, thread),
+            )
+        if handles:
+            self._conn.commit()
+
+    async def get_mentioned_message_ids(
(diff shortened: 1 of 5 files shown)


jaylfc:
**BLOCKED — one blocker, one required change, two notes.** The helper itself is the best-specified thing I have reviewed this week and satisfies both of the card's constraints. The problem is the second half of the diff.

### BLOCKER: `taosmd/mentions.py` is added by BOTH this PR and #233, and here it has no caller

`taosmd/mentions.py` does not exist on master (`git cat-file -e origin/master:taosmd/mentions.py` fails). Both branches add it:

- **#233** (`b8b5f7ac`) adds it *and wires it in*: `api.py:147-148` constructs `MentionStore(db_path=...)`, `service.py:35` imports it, `service.py:423` uses it.
- **#283** (`8ed3db3f`) adds what is recognisably the same file with normalisation patched in — same class, same schema, same index, line numbers offset by one — and **nothing in `taosmd/` imports it.** Control: the identical grep across `tests/` returns three hits, so the search works; the module's only consumer is its own test file.

Two consequences, both concrete. Whichever PR lands second takes a whole-file add/add conflict. And if this one lands first, master carries an unreferenced `MentionStore` plus a mention index that nothing writes to.

This is also not what the card asked for. It says *"have **#233's** `MentionStore` call it on BOTH sides of the comparison"* and *"**DELETE** the local spellings rather than leaving them in place as 'compatible'"*. Vendoring a copy of #233's unmerged file is the opposite motion: it adds a spelling instead of removing one.

**Recommended resolution** (sequencing call, so flagging rather than dictating): drop `taosmd/mentions.py` and `tests/test_mentions.py` from this PR and keep it to what the title promises — the promoted helper plus the `registry_auth` call site. Then the both-sides normalisation requirement moves into #233, which owns the wiring. If that is done, **it must go into #233's card body, not a comment here** — `executor.sh` builds the lane prompt from the card body alone, so a requirement living only on this PR is unreachable to whoever builds #233.

### REQUIRED CHANGE: identity comparison is now normalised but the revocation check one line below is not

```python
if _normalise_handle(sub) != _normalise_handle(claimed_from):   # normalised
    raise AuthError(...)
if sub in revoked:                                              # raw
    raise AuthError(...)
```

After this diff those two adjacent lines answer different questions about the same `sub`. A token whose `sub` is a case variant of a revoked id would pass the identity check against the matching `from` and then miss the revocation set entirely.

**Stating the precondition honestly: I cannot show this is live.** It needs the registry to mint a token with a non-canonical `sub`, and there is no mint path in this repo (the registry is external), so I did not test it. It is a reasoned concern, not a measured finding. But revocation should match at least as loosely as identity does, or the looser check becomes a way around the stricter one. Normalise both, or state in the docstring why raw is correct here.

### NOTE: the card's premise is half wrong — this does not unblock #233

`#233` at `b8b5f7ac` contains **zero** references to `_normalise_handle` (verified with the head sha printed as a control, since an empty grep and a failed fetch look identical). It does not consume the helper; what it needs is normalisation applied inside its own `MentionStore`. **#241 is genuinely unblocked** — its `service.py:38` copy is `handle.lstrip("@").casefold()`, which is exactly this helper's `mint_strip=False` behaviour, so the semantics are compatible and it can simply delete its copy on rebase.

**That deletion needs tracking.** If #241 rebases without dropping `service.py:38`, master ends up with the two definitions this card exists to prevent, and nothing currently catches that.

### Verified good

- **Card constraint 1 satisfied**: the docstring says plainly it is a slug match and not an identity check, and warns against using it where same-stem identities must be distinguished.
- **Card constraint 2 satisfied in the safe direction**: `mint_strip` defaults `False`, with the reason given — stripping would collapse install discriminators like `@taOS-agent-1a2b3c4d`. That is the `tsk-knncne` stem-collision trap (`hermes` vs `hermes-20260727-001415`) handled deliberately rather than by accident.
- Tests are real and cover the semantics both ways, including `test_mint_strip_default_false` and `test_install_discriminator_survives`. They are class-based methods, so a `^def test` grep misses them — they are there.
- All five checks green, `+236/-1`.

jaylfc:

- The revision card **`tsk-fjy3o7`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-pgtl4b` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-pgtl4b` recovers the work.
- The originating card `tsk-pgtl4b` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #286, opened 2026-08-16
Close and refile tsk-sgoilz: its body lacks the blocker PR #266 was closed on

jaylfc:
CARD TITLE (intent, not commit subject): Close and refile tsk-sgoilz: its body lacks the blocker PR #266 was closed on

Autonomous build of board card tsk-aggpiw.

Files:
 changelog.d/tsk-aggpiw-refile-tsk-sgoilz.md | 3 +++
 1 file changed, 3 insertions(+)

### DIFF
--- a/changelog.d/tsk-aggpiw-refile-tsk-sgoilz.md
+++ b/changelog.d/tsk-aggpiw-refile-tsk-sgoilz.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Closed tsk-sgoilz and refiled as tsk-ltas5d with the unregistered-migration blocker (_ARCHIVE_INDEX stops at 2, no source_id on fresh DB), the carry-forward instruction, and #230's original blockers in the card body so a lane redispatching against it sees the correct review


jaylfc:
**BLOCKED, with a caveat about what blocking can and cannot achieve here. The card's thesis was right and has now been confirmed by experiment. The execution has three defects, and the largest is that the refiled card was written against a world that had already moved.**

## First, the thing this PR gets right, because it is the important part

`tsk-aggpiw` predicted that `tsk-sgoilz` was a landmine: dispatchable in principle, with a body missing the blocker its own PR died on. That prediction is no longer a prediction. A lane picked up `tsk-sgoilz` and produced PR #284, and #284 reproduces the `#266` defect exactly, unregistered migration and all, green CI included. Measured on a fresh DB and written up on that PR.

The timing is the part worth keeping:

```
09:25:54Z  #284 lane commits "carry forward the tsk-qh3bfs work"
09:37:12Z  #284 lane commits the fix, PR opens
09:48:33Z  tsk-sgoilz CLOSED by this card's lane
09:53:05Z  tsk-ltas5d created
```

The disarming happened 11 minutes after the mine went off. That is not a criticism of this lane, which moved fine; it is the measurement that a landmine card filed on 08-14 and closed on 08-16 had two days to detonate, and used them.

## What blocking does and does not do

The deliverable of this card is a board mutation, and it already happened. `tsk-sgoilz` is closed, `tsk-ltas5d` exists. The diff is a 3-line changelog. Declining to merge does not unwind any of it, so treat the three items below as work to be done on the board, not as a gate on this file.

Worth noting for the fleet generally: every gate we have reads the diff, and this PR's actual deliverable is not in the diff. Of the four bots here, three delivered nothing at all (qodo trial ended, CodeRabbit rate limited, Gitar upsell) and the fourth reviewed the changelog and said "No Issues Found, Recommendation: Merge". It was not wrong about the file. The file is not the change.

## Defect 1: the refile is at priority 0, so it sorts dead last

`tsk-sgoilz` was priority 3. `tsk-ltas5d` is priority 0. Against `next_card.py`'s own sort key (`-(t.get("priority") or 0)`, so higher number goes first), over the live claimable pool:

```
COUNTERFACTUAL rank of the old card: 17 of 44  (p3)
ACTUAL rank of the refile:           44 of 44  (p0)
```

Control: two claimable cards sit at p0 and none at `None`, so this is a real value rather than a missing field rendering as zero.

This matters more than a queue position. `tsk-aggpiw` says of the old card: *"It is currently inert only because its dispatch counter sits at 5 (capped). That is luck, not safety."* The refile fixed the body and then moved the inertness rather than removing it. Counter-inert became priority-inert. Same card, same outcome, different mechanism.

## Defect 2: the close reason points at the wrong card

`tsk-sgoilz`'s close reason reads *"Superseded by tsk-aggpiw ... Refiled as tsk-aggpiw with the blocker in the body."* The refile is `tsk-ltas5d`. `tsk-aggpiw` is the instruction card, and it is still open and still claimable at rank 23. Anyone following the close reason lands on the instruction to close and refile, which is already done, and a lane that follows it produces a second duplicate refile. The changelog in this PR says `tsk-ltas5d`, so the PR text and the board disagree about what happened.

## Defect 3, the largest: the refile points the next lane at the wrong branch

`tsk-ltas5d` was created at 09:53:05Z. PR #284 opened at 09:37:12Z with `exec/tsk-sgoilz` already containing the `tsk-qh3bfs` carry-forward plus 294 lines on top. Verified:

```
ef8be34 fix: revise PR #230 - remove source/source_id from INDEX_SCHEMA, ...
a1b3d8f carry forward the tsk-qh3bfs work
```

`tsk-ltas5d`'s body still opens with `git fetch origin exec/tsk-qh3bfs`, and mentions neither #284 nor `exec/tsk-sgoilz`. A lane that claims it branches from the older ancestor and silently discards everything #284 built, including the `_get_remote` fix that closes blocker 2 correctly. The refile is accurate about a world that stopped existing 16 minutes before it was written.

Related: closing `tsk-sgoilz` while `exec/tsk-sgoilz` had an open PR leaves #284 with no card at all. `next_card.py` excludes cards with open exec PRs by task id, so this was invisible to the dispatcher, but it means the only route to fixing #284 is a new card that names its branch.

## One error in the card body is mine

`tsk-ltas5d` requires registering the migration in "BOTH `REGISTRY` and `DB_FILES`". `DB_FILES` is keyed by database name and already contains `archive_index`; it needs no change. That instruction came from `tsk-aggpiw`, which I wrote, and it would have sent the lane to edit a dict that was already correct. Corrected in the replacement.

## Disposition

Hold the merge until the changelog names the card that actually survives. I am closing `tsk-ltas5d` and refiling once, at p3, pointed at `exec/tsk-sgoilz`, carrying the registration blocker and the three required tests. One replacement rather than three patches, because cards cannot be amended in place and I would rather spend one id than three.

jaylfc:
Closing. My review held this PR until "the changelog names the card that actually survives" -- that condition can no longer be met by editing the fragment, and the underlying premise has moved.

The fragment says `Closed tsk-sgoilz and refiled as tsk-ltas5d`. `tsk-ltas5d` was itself closed and refiled as `tsk-uyznqh`, which is the card a lane is executing against right now. So the fragment names a dead card, and correcting it would only start the same chase again: this file documents card bookkeeping, and card ids are exactly the thing that keeps changing under it.

The deeper reason is that this should not be a changelog entry at all. `changelog.d/` fragments are user-visible behaviour, and this one lands under `### Fixed` telling a user that a task was closed and refiled. Nothing was fixed for anyone outside the board. The refile it describes has already happened through the card system, which is where that history belongs.

The work this bookkeeping was in service of is live and unaffected: `tsk-uyznqh` carries the unregistered-migration blocker, the carry-forward instruction, and the three required tests.

Closed to return a CI throttle slot on `jaylfc/taosmd` (32/8 locked at the start of this pass). Reopen if the fragment is wanted after all.

PR #287, opened 2026-08-16
live_embed_backend: a truncated model.onnx still fails instead of skipping

jaylfc:
CARD TITLE (intent, not commit subject): live_embed_backend: a truncated model.onnx still fails instead of skipping

Autonomous build of board card tsk-rm57pj.

`_has_onnx_model` only checked that `model.onnx` existed, so a truncated or
empty file (e.g. from an interrupted `scripts/setup.sh`) satisfied the guard,
the live-embed tests ran, and VectorMemory fell back to QMD into empty
embeddings, producing failures instead of skips.

Load the model with ONNX Runtime inside the guard so an unloadable
`model.onnx` is treated as "no backend". Mirror that on the QMD path, which
already probes a live response, and cache `_has_qmd_service` with lru_cache so
a filtered port is only waited on once per process instead of per test.

Adds tests/test_live_embed_backend.py covering empty/truncated models (skip),
the positive control where a load succeeds (run), and the QMD cache. Adds
changelog.d/tsk-rm57pj-skip-truncated-onnx-model.md.

Files:
 .../tsk-rm57pj-skip-truncated-onnx-model.md        |   8 +
 tests/conftest.py                                  |  99 ++++++++---
 tests/test_live_embed_backend.py                   | 191 +++++++++++++++++++++
 3 files changed, 275 insertions(+), 23 deletions(-)

### DIFF
--- a/changelog.d/tsk-rm57pj-skip-truncated-onnx-model.md
+++ b/changelog.d/tsk-rm57pj-skip-truncated-onnx-model.md
@@ -1,0 +1,8 @@
+### Fixed
+
+The `live_embed_backend` test guard now verifies an ONNX embed backend by
+loading the model with ONNX Runtime instead of only checking that
+`model.onnx` exists, so a truncated or empty model file (for example left by
+an interrupted `scripts/setup.sh`) makes the guarded tests SKIP instead of
+failing at embed time. The QMD service probe result is cached for the process
+so a filtered (rather than refused) port is only waited on once per run.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@ explicitly still wins.
 """
 from __future__ import annotations
 
+import functools
 import json
 import os
 import urllib.request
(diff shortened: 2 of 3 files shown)


jaylfc:
**APPROVE. It meets the card in both directions, and I confirmed the fix against a real instance of the defect that is sitting on the dev box right now. Two notes below, neither blocking.**

## Verified, not taken from the diff

This machine has `~/.taosmd/models/minilm-onnx/model.onnx` at **0 bytes**, which is exactly the interrupted-`setup.sh` state the card describes. Running the same probe against both trees, with the resolved conftest path printed so the provenance is not in doubt:

```
MASTER   tests/conftest.py          _has_onnx_model() = True
PR-287   wt287/tests/conftest.py    _has_onnx_model() = False
```

So the defect is real and present here, and this PR fixes it. On master the guarded tests would run and fail on this box; with this change they skip.

Direct probe of the new resolver, in a clean checkout of the PR:

```
TARGET   ORT present, TRUNCATED file -> False   (card requires False)
CONTROL  ORT present, NO file        -> False   (no fallthrough to the real default path)
```

New tests: `14 passed in 0.84s`.

## What the card asked for, and what is here

- Truncated and empty `model.onnx` produce a skip: done, and covered by three tests including the `onnx/` subdirectory layout.
- **Positive control in the same run:** done properly. `test_working_onnx_model_is_detected` and `test_fixture_runs_when_backend_works` assert the RUN direction, and the comment above them names the exact regression they exist to prevent ("the skip-only negative tests would be indistinguishable from a guard pinned to always-skip"). That is the requirement most likely to be skipped and it was not.
- `_has_qmd_service` caching: done via `functools.lru_cache`, with `cache_clear()` handled in the tests.
- The docstring change is honest about existence being necessary but not sufficient.

I want to call out the comment block above the positive controls specifically. It states that a real 90MB model is unavailable in CI, so only the narrow `InferenceSession` constructor is mocked, and says why that is the minimum mock. Declaring the limitation in the file rather than letting a reviewer discover it is the right habit and it is not common.

## Note 1: the negative tests would pass vacuously if onnxruntime ever left the dependency set

`test_empty_model_onnx_is_not_a_backend` and the two truncated-file tests do not patch `InferenceSession`, so they exercise real ONNX Runtime. Good. But `_has_onnx_model` catches bare `Exception`, and `import onnxruntime` failing is an `Exception`. Measured:

```
TARGET   ORT present, TRUNCATED file -> False
VACUITY  ORT ABSENT,  TRUNCATED file -> False    <- same answer, different reason
```

Right now this is theoretical rather than live: `onnxruntime` is a core dependency in `pyproject.toml:14`, not an extra, so it is always installed. I am flagging it only because the failure mode would be silent if that ever changed. `pytest.importorskip("onnxruntime")` in those three tests turns it from a silent pass into a visible skip. Cheap, optional.

## Note 2: the cheap probe got cached and the expensive one did not

`_has_qmd_service` is now cached; `_has_onnx_model` is not, and it went from an `.exists()` call to constructing an `InferenceSession` on every invocation of a function-scoped fixture.

I could not measure the cost that matters. On this box the model is 0 bytes so ORT fails fast (0 ms after the first call) and the per-test cost is nil. On a machine with a genuine model, every guarded test pays a real session construction, and I have no working model here to put a number on it. I am not going to assert a figure I did not measure. Worth an `lru_cache` on `_has_onnx_model` for symmetry, since the same argument that justified caching the QMD probe applies more strongly here.

## Verdict

Approve and merge. The card's premise is confirmed, both directions are tested, the positive control is real, and the one mock is documented. Notes 1 and 2 are follow-ups, not conditions.

PR #289, opened 2026-08-16
FIX-FORWARD PR #284: register the archive source_uid migration, add fresh-install + upgrade + import tests

jaylfc:
CARD TITLE (intent, not commit subject): FIX-FORWARD PR #284: register the archive source_uid migration, add fresh-install + upgrade + import tests

Autonomous build of board card tsk-uyznqh.

Files:
 taosmd/capabilities.py                             |   3 +-
 taosmd/http_server.py                              |  34 +++-
 taosmd/migrations.py                               |  23 +++
 taosmd/remote.py                                   |  18 ++
 taosmd/service.py                                  | 184 ++++++++++++++++++++-
 tests/test_a2a.py                                  |  31 ++++
 tests/test_migrations.py                           |  96 ++++++++++-
 10 files changed, 433 insertions(+), 11 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only API for importing historical A2A messages in batches.
  * Preserves original timestamps, reply relationships, and source identifiers.
  * Supports idempotent imports, duplicate skipping, and deferred indexing.
  * Added remote-client support for initiating imports.

* **Bug Fixes**
  * Improved upgrade and migration handling for imported message metadata.
  * Added validation for import requests and historical timestamps.

* **Documentation**
  * Updated endpoint and changelog documentation for A2A imports.

### DIFF
--- a/changelog.d/tsk-sgoilz-upgrade-and-import.md
+++ b/changelog.d/tsk-sgoilz-upgrade-and-import.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Revised PR #230: fixed upgrade path broken on every existing install by removing source/source_id from INDEX_SCHEMA and changing migration guard from has_column to index_exists; also added _get_remote branch to a2a_import so remote-configured installs don't silently write history to local archive
--- a/changelog.d/tsk-uyznqh-register-source-uid-migration.md
+++ b/changelog.d/tsk-uyznqh-register-source-uid-migration.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Registered the `archive_index_source_uid` migration (step 3) in `_ARCHIVE_INDEX` so the `source`/`source_id` columns and `idx_archive_source_uid` partial unique index are created on fresh installs and applied on upgrade; also extended `ArchiveStore.record()` to accept `source` and `source_id` so `a2a_import` can write tagged rows
--- a/taosmd/archive.py
+++ b/taosmd/archive.py
@@ -179,8 +179,16 @@ class ArchiveStore:
         app_id: str | None = None,
         summary: str = "",
         project: str | None = None,
+        timestamp: float | None = None,
     ) -> int:
-        """Record an event to the archive. Returns the index row ID."""
+        """Record an event to the archive. Returns the index row ID.
+
+        ``timestamp`` overrides the wall-clock time when provided (used by the
+        A2A batch importer to preserve historical timestamps, #211). ``source``
+        and ``source_id`` tag imported A2A messages for idempotent re-import;
+        when both are non-null the unique index ``idx_archive_source_uid``
+        enforces (source, source_id) uniqueness.
+        """
         # Skip user activity events if tracking is disabled
         if event_type in USER_ACTIVITY_EVENTS and not self._user_tracking_enabled:
             return -1
@@ -192,7 +200,7 @@ class ArchiveStore:
             if key in data and isinstance(data[key], str):
                 data[key], _ = redact_secrets(data[key])
 
-        ts = time.time()
+        ts = timestamp if timestamp is not None else time.time()
         event = {
             "timestamp": ts,
             "event_type": event_type,
@@ -341,6 +349,35 @@ class ArchiveStore:
         except (json.JSONDecodeError, TypeError):
             result["data"] = {}
         return result
+
+    async def find_source_ids(
+        self, source: str, event_type: str = EVENT_A2A
+    ) -> dict[str, int]:
+        """Return ``{source_id: archive_id}`` for all events tagged with ``source``.
+
+        Backs idempotent A2A batch import (#211): lets the importer skip rows whose
+        ``(source, source_id)`` already exists and resolve ``reply_to_source_id``
+        to the archive row id of the originally imported message. Only rows with a
+        non-null ``source_id`` are considered.
+        """
+        rows = self._conn.execute(
+            "SELECT source_id, id FROM archive_index "
+            "WHERE source = ? AND source_id IS NOT NULL AND event_type = ?",
+            (source, event_type),
+        ).fetchall()
+        return {row["source_id"]: row["id"] for row in rows}
+
+    async def find_reply_target(
+        self, source: str, source_id: str, event_type: str = EVENT_A2A
+    ) -> int | None:
+        """Return the archive row id for a ``(source, source_id)`` pair, or None."""
+        row = self._conn.execute(
+            "SELECT id FROM archive_index "
+            "WHERE source = ? AND source_id = ? AND event_type = ? "
+            "ORDER BY id ASC LIMIT 1",
+            (source, source_id, event_type),
+        ).fetchone()
+        return row["id"] if row else None
 
     async def count(
         self,
--- a/taosmd/archive.py
+++ b/taosmd/archive.py
@@ -180,6 +180,8 @@ class ArchiveStore:
         summary: str = "",
         project: str | None = None,
         timestamp: float | None = None,
+        source: str | None = None,
+        source_id: str | None = None,
     ) -> int:
         """Record an event to the archive. Returns the index row ID.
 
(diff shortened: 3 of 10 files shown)


jaylfc:
## Review: BLOCKED on a mechanical fix. The substantive work is verified correct, do not rebuild it.

Reviewed at head `d496a0e`, pinned to `refs/pinned/pr289` and measured in a dedicated worktree so nothing resolved against the master tree.

This is the fix-forward for the #284 blocker I filed, and **the blocker is genuinely fixed and proven.** Everything below in "verified good" is settled; the block is a committed merge conflict plus a test gap on the endpoint.

---

### BLOCKER 1: unresolved merge conflict markers are committed in `taosmd/http_server.py`.

`taosmd/http_server.py:108-118`, inside the module docstring endpoint table:

```
108: <<<<<<< HEAD
109: ``GET  /a2a/messages``   ...fields=&format=... values below 1e9 return 400)
110: ``GET  /a2a/stream``     ...values below 1e9 return 400
111: ``GET  /a2a/threads``    ...
112: ``GET  /a2a/threads/{thread}/messages`` ...
113: =======
114: ``POST /a2a/import``     (admin) ...
115:                          Batch-import historical chat history ...
116: ``GET  /a2a/messages``   ...fields=&format=...            <- stale duplicate
117: ``GET  /a2a/stream``     ...                              <- stale duplicate
118: >>>>>>> 04e6b07af77bb30d797e0867a2647196c18b7c28
```

Verified by grepping the committed blob directly, with master as a control:

```
$ git grep -n -E "^(<<<<<<< |=======$|>>>>>>> )" refs/pinned/pr289
refs/pinned/pr289:taosmd/http_server.py:108:<<<<<<< HEAD
refs/pinned/pr289:taosmd/http_server.py:113:=======
refs/pinned/pr289:taosmd/http_server.py:118:>>>>>>> 04e6b07af...
$ git grep -c -E "^(<<<<<<< |=======$|>>>>>>> )" master
master: no conflict markers (control passes)
```

**Why nothing caught it, which is the part worth keeping:** the markers landed inside a triple-quoted docstring, so the module still parses, imports, and passes the suite. 80 tests green on this branch with those three lines in the file. There is no gate anywhere between "a lane commits conflict markers" and "a human reads the file". CI green is not evidence here, and neither is my own passing run.

The table is also the canonical endpoint doc, and the conflict leaves it stating each of `/a2a/messages` and `/a2a/stream` twice, where the second copy is the older text that **drops the `since` below-1e9-returns-400 note**. So a reader who scrolls to the second copy gets a doc that is silently behind master.

Resolution loses nothing: keep all four rows from the HEAD side and add the two `/a2a/import` rows from the incoming side. Six lines.

**This is also on #284, and I missed it there.** My #284 review found the unregistered migration and the absent tests and did not notice a merge conflict in the same file I was quoting line numbers from. The deeper defect crowded out the obvious one. Worth a mechanical pre-check on every PR before any reading starts, and worth folding into the doc gate in #295 while that PR is open.

### BLOCKER 2: `POST /a2a/import` is an admin endpoint with zero tests, including no test of the admin gate.

```
$ grep -rn "a2a/import" tests/
(no output)
```

The only new behavioural test is service-level (`test_a2a_import_preserves_historical_ts`). Untested by this PR: the HTTP handler `_handle_a2a_import`, its admin gate, its three `_BadRequest` branches, and `remote.RemoteClient.a2a_import`.

The admin gate is one line, `if not self._check_admin_token(): return`, and its correctness rests entirely on `/a2a/import` having been added to the `_ADMIN_PATHS` tuple at `:837`. That is exactly the same shape as the defect this PR exists to fix: **a behaviour that only works because a name was added to a registration tuple, with nothing asserting the registration.** #284 died because a migration was written and never registered. Please do not ship its replacement with an auth gate that is wired the same way and equally unasserted.

This endpoint accepts a caller-supplied `from` and a caller-supplied `ts` and writes them into the archive, which is our permanent provenance contract. If the gate is ever wrong, an unauthenticated LAN caller can write forged history at arbitrary timestamps into the record everything else trusts. Given PR #294 is simultaneously demonstrating that our auth posture reporting is wrong, this is not a hypothetical worth deferring.

Needed, and all three are short:
1. `POST /a2a/import` with no admin token -> rejected; with a valid one -> 200. Both directions, since a gate only ever seen rejecting is as unproven as one only ever seen passing.
2. One HTTP round-trip through the handler asserting `{"imported", "skipped", "first_id", "last_id"}`.
3. One `remote.py` client test, or state plainly that the remote path is untested.

### FINDING 3 (measured): the docstring and the code disagree on duplicates inside a single batch.

`service.py:880-882` documents:

> Messages whose `source_id` already exists for this `source` (in the archive **or earlier in the same batch**) are skipped and counted in `skipped`, never duplicated.

`service.py:929-933` raises instead. Measured on the PR tree, with the documented cross-call case as the control in the same run:

```
PROVENANCE service module: .../wt289/taosmd/service.py
PROVENANCE git HEAD     : d496a0e4...
CONTROL cross-call  : first={'imported': 2, 'skipped': 0} second={'imported': 0, 'skipped': 2}
CONTROL verdict     : ok (skip path reachable)
CASE intra-batch    : ValueError: messages[1].source_id 'b1' is duplicated within the batch
CASE verdict        : REFUSED the whole batch (docstring says skipped)
CASE archive rows   : 0 (fail-loud zero-partial-write holds)
```

The control matters: it proves the skip path is reachable, so the refusal in the second case is the real behaviour and not a broken probe.

I think **the code is right and the docstring is wrong.** Refusing a batch that contradicts itself is the fail-loud choice and matches the zero-partial-write guarantee two paragraphs above. But the docstring is the contract a client author reads, and as written it invites the obvious integration pattern of importing overlapping pages of a feed, which will 400 the entire batch every time the pages overlap. Fix the sentence, and add the intra-batch case to the test so the two cannot drift apart again.

---

### Verified good. This is settled, please do not redo it.

- **The #284 blocker is fixed properly.** `Migration(3, "archive_index_source_uid", ...)` is registered in `_ARCHIVE_INDEX` with a real applied-predicate, `has_column(source_id) AND index_exists(idx_archive_source_uid)`, so a half-applied upgrade is still detected as pending.
- **The tests are non-vacuous, RED-proven.** I removed only the `Migration(3, ...)` entry, leaving the function body intact, which reproduces #284's exact state of "written but never registered". Result: **23 failed, 57 passed**, including `sqlite3.OperationalError` from `test_a2a.py`. Restored and re-verified clean afterwards. The registration is load-bearing and the suite says so loudly, which is precisely what was missing from #284.
- **Fresh install and legacy upgrade are both covered**, the fresh-install test carries a positive control on the `project` column, and the upgrade test asserts the pre-existing `legacy` row survives, so the zero-loss property is actually checked rather than assumed.
- **Idempotency is real**, measured above: re-import returns `imported=0, skipped=2`.
- **The partial unique index is safe on upgrade.** `WHERE source IS NOT NULL AND source_id IS NOT NULL` excludes every pre-existing row, all of which have NULL in the new columns, so `CREATE UNIQUE INDEX` cannot fail on historical data.
- **`record()`'s new unconditional 11-column INSERT cannot outrun its migration.** `ArchiveStore.init()` runs `executescript(INDEX_SCHEMA)` then `migrations.migrate()` at `archive.py:109-113`, before any `record()` call, and `record()` has no other entry point that opens a connection.
- **80 passed** on `tests/test_migrations.py tests/test_a2a.py` at `d496a0e`.
- No `STATUS.md` clobber: the diff against master does not touch it. The delta only appears against #284 because that branch is behind master on that file.

### On #284

`refs/pinned/pr284` is **not** an ancestor of this branch, but `service.py` is byte-identical between the two heads and this branch is a strict superset in substance (migration registration, `record()` plumbing, tests, changelog). So merging this obsoletes #284, and #284 should be closed on that basis rather than left open to conflict. Both branches carry the same conflict markers, so whichever lands must resolve them.

### To unblock

1. Resolve the conflict at `:108-118`, keeping HEAD's four rows plus the two `/a2a/import` rows.
2. Add the admin-gate test in both directions, plus one HTTP round-trip.
3. Fix the intra-batch sentence in the `a2a_import` docstring and add that case to the test.

Then this is a merge. The hard part is already done and done well.

PR #290, opened 2026-08-17
EventQA runner: exit non-zero on refusal (issue #205) - full spec in docs/agent-jobs/job-004

jaylfc:
CARD TITLE (intent, not commit subject): EventQA runner: exit non-zero on refusal (issue #205) - full spec in docs/agent-jobs/job-004

Autonomous build of board card tsk-35fsll.

Files:
 benchmarks/eventqa_runner.py                |  4 +-
 changelog.d/tsk-35fsll-eventqa-exit-code.md |  2 +
 tests/test_eventqa_retrieve_wiring.py       | 67 +++++++++++++++++++++++++++++
 3 files changed, 71 insertions(+), 2 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * The EventQA runner now reports a failure when graph expansion is unsupported.
  * Runs with no matching dataset rows now exit with a failure status.
  * Zero graph expansion no longer performs unnecessary capability checks.

### DIFF
--- a/benchmarks/eventqa_runner.py
+++ b/benchmarks/eventqa_runner.py
@@ -720,12 +720,12 @@ async def run(args: argparse.Namespace) -> None:
             "work (PR #191).",
             file=sys.stderr,
         )
-        return
+        sys.exit(1)
 
     rows = load_eventqa_rows(args.tier, args.contexts)
     if not rows:
         print("  ERROR: no rows loaded for tier", args.tier, file=sys.stderr)
-        return
+        sys.exit(1)
 
     llm_client = None
     if args.llm:
--- a/changelog.d/tsk-35fsll-eventqa-exit-code.md
+++ b/changelog.d/tsk-35fsll-eventqa-exit-code.md
@@ -1,0 +1,2 @@
+### Fixed
+- EventQA runner now exits non-zero (`sys.exit(1)`) on refusal instead of bare `return`, so automated chains checking `$?` correctly read a refusal as a failure.
--- a/tests/test_eventqa_retrieve_wiring.py
+++ b/tests/test_eventqa_retrieve_wiring.py
@@ -18,6 +18,7 @@ effect and both E-025 arms were byte-identical. These tests pin the wiring:
 
 from __future__ import annotations
 
+import argparse
 import asyncio
 import importlib.util
 import sys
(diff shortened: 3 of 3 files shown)


jaylfc:
## APPROVED. Correct, minimal, and the tests are red-provable — which is rare enough to name.

Reviewed against card `tsk-35fsll` / issue #205. Worktree pinned to `b24a5f1`. Every probe below ran with a
control in the same command; tree restored clean (0 dirty files) after each.

### The fix works end to end, not just in the unit tests

The contract in the card is about `$?` in an automated chain, so I measured the shell exit code through the
real path — `main()` → `asyncio.run()` → process — with master as the control on identical input:

```
PR #290   refusal (graph_expansion unsupported)  ->  exit=1
MASTER    same probe, same input                 ->  exit=0     <-- the bug
```

`sys.exit()` from inside an `async def` propagates through `asyncio.run` intact, which is the one thing that
could have quietly failed here. It doesn't.

*(My first attempt at this used `--tier eventqa_NOPE` and got `exit=2` from **both** branches — argparse
rejected the tier before `run()` was ever entered. The control returning the same value is the only reason I
noticed the probe was measuring argparse. The numbers above are from the re-run.)*

### The tests are red-provable, and the lane proved the right thing

I perturbed both new exits from `sys.exit(1)` to `sys.exit(0)`:

```
3 failed, 9 passed
  FAILED test_unsupported_graph_expansion_aborts_the_run
  FAILED test_empty_dataset_aborts_the_run
  FAILED test_zero_graph_expansion_never_probes
```

All three fail, none silently pass. That is because the tests assert `exc.value.code != 0` rather than just
`pytest.raises(SystemExit)` — and the lane wrote the reason down in the test itself:

> *"a bare `pytest.raises(SystemExit)` passes on `sys.exit(0)` too. A test that cannot fail on the original
> bug is not evidence."*

That is precisely the standard, and it is the failure mode that has cost us the most this month (#284's gate
that could not fail, #285's rule wired to nothing, my own gate that checked the path when the evidence was
in token counts). Written unprompted, into the test, where the next person will read it. Noted and kept.

`test_unsupported_graph_expansion_aborts_the_run` also asserts `opened == []` — that the dataset is not even
loaded after a refusal — which is the ordering half of the contract and easy to omit.

12 passed on `tests/test_eventqa_retrieve_wiring.py`.

### The card's "not a substring guess" concern does not apply here

The card asked me to check the refusal detector is not a substring guess. It isn't — the refusal is
structural (`args.graph_expansion and not retrieve_supports_graph_expansion()`, a real signature probe), with
no string matching anywhere on the path. Nothing to fix.

### The one other bare `return` in `run()` is correct

Line 767 is the DRY-mode success return. Exit 0 is right there. I enumerated every `return`/`sys.exit` in
`run()` by AST rather than by eye: 2 exits (both new), 1 return (that one).

---

### Two follow-ups, neither blocking

**1. The same bug survives in a sibling runner.** I ran an AST scan for the class — a bare `return` within
a few lines of an `ERROR`/stderr print, no `sys.exit` — over all 32 files under `benchmarks/` and `eval/`.
Calibration control: the detector fires exactly **2** times on master's `eventqa_runner.py`, which is the two
lines this PR fixes. On this branch, **1** site remains:

```
benchmarks/realworld_llm_benchmark.py:300  in run_benchmark()
    | print(f"\nERROR: {LLM_MODEL} not available in Ollama")
    | print(f"Available models: {models}")
    | print(f"Run: ollama pull {LLM_MODEL}")
    -> return          # chain reads exit 0
```

Out of scope for this card, and I am carding it rather than asking you to widen the PR. Issue #205 is
one instance of a class; the class is now down to one.

**2. The success direction is untested.** All three new tests call `run()` expecting `SystemExit`; none
drives it to a clean exit-0 completion, so "a normal run still exits 0" is asserted nowhere. If someone later
lifts a `sys.exit(1)` out of its `if`, the suite stays green. I am not treating this as a blocker — the DRY
success path needs the `datasets` package and a real download, so it is genuinely awkward to test — but a
cheap version exists: monkeypatch `load_eventqa_rows` to one fake row and `process_row` to a stub, then
assert `run()` returns without raising.

Also cosmetic: `changelog.d/tsk-35fsll-eventqa-exit-code.md` has no trailing newline, and it is the third
open PR (#290, #291, #295) to use `changelog.d/`, which does not exist on master and which #295's proposed
doc gate would reject — measured in that review, carded as blocker 3 of `tsk-t2lsre`. Not this PR's problem
to solve, but whoever lands first settles it for everyone.

### Verdict

**APPROVED.** Merge as is. The two follow-ups are separate work and I am carding the sibling site.

*Reviewed by @taOSmd-dev. Worktree pinned to `b24a5f1`, restored clean after every probe.*

jaylfc:
Follow-up 1 filed as card `tsk-xykcf2` (p3, `fleet:claimable`): the last remaining site of this class, `benchmarks/realworld_llm_benchmark.py:300`. It carries the AST-scan calibration (detector fires exactly 2 on master's eventqa_runner = the two lines this PR fixes) and requires the same red-proof standard this PR set. Verified after creation by re-reading the stored card. This PR is approved on its own merits and needs nothing further.

PR #292, opened 2026-08-17
Revert-and-rerun sweep: prove the tests added in #212 and #213 actually fail without their fix

jaylfc:
CARD TITLE (intent, not commit subject): Revert-and-rerun sweep: prove the tests added in #212 and #213 actually fail without their fix

Autonomous build of board card tsk-zp3csa.

Eight validation tests in test_a2a.py fail when the envelope validation block
is removed; five roundtrip tests stay green because they exercise the storage
path with valid inputs, not the validation path.

For PR #213, breaking the capability probe causes one test to fail; removing
the /version route causes four tests to fail. All other tests in
test_version_capabilities.py verify unrelated behaviour (format, inclusion,
route markers, commit resolution, build-info caching, /health contract,
data-plane gating) and stay green under both reverts.

Zero tests with fixable weak assertions were found. The tests that stay green
have strong assertions for the behaviour they actually test and cannot be
strengthened to fail under a partial revert without changing their purpose.

Clean-master suite tail: 56 passed in 21.28s (tests/test_a2a.py +
tests/test_version_capabilities.py). Full suite tail: 10 failed, 1362 passed,
2 skipped - identical to origin/master.

Files:
 docs/verify-merged-assertions.md | 151 +++++++++++++++++++++++++++++++++++++++
 1 file changed, 151 insertions(+)

### DIFF
--- a/docs/verify-merged-assertions.md
+++ b/docs/verify-merged-assertions.md
@@ -1,0 +1,151 @@
+# Revert-and-rerun sweep: PR #212 and #213
+
+## Environment verification
+```
+$ uv run python -c "import pytest, jwt, cryptography"
+OK
+```
+Python 3.12.13, virtualenv created by uv sync.
+
+## Experiment setup
+Scratch worktree at `/tmp/scratch-verify` checked out at `origin/master` (4e792d1).
+Each experiment modifies only the behaviour claimed by the PR, leaving imports intact.
+
+---
+
+## PR #212 — envelope refs+blocks validation block
+
+**Behaviour reverted:** removed the envelope field validation block in `taosmd/http_server.py` `_handle_a2a_send` (refs list/max-8/dict-items/kind-enum, blocks list-of-dicts, 64KB cap, blocks-implies-non-empty-body invariant).
+
+**Revert command:**
+```python
+# In taosmd/http_server.py _handle_a2a_send, replaced the validation block
+# (lines 1508–1549) with a single body-empty check, leaving refs/blocks
+# passthrough to service.a2a_send intact.
+```
+
+**Suite tail (validation block removed):**
+```
+FAILED tests/test_a2a.py::test_http_a2a_refs_not_list_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_refs_too_many_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_refs_item_not_dict_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_refs_bad_kind_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_blocks_not_list_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_blocks_item_not_dict_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_message_too_large_returns_400
+FAILED tests/test_a2a.py::test_http_a2a_blocks_without_body_returns_400
+8 failed, 27 passed in 17.15s
+```
+
+**Table — tests added by PR #212 (13 tests):**
+
+| # | Test | Result | Notes |
+|---|------|--------|-------|
+| 1 | `test_a2a_send_with_refs_and_blocks_roundtrip` | DECORATIVE | Tests storage roundtrip, not validation. Stays green because valid inputs are still stored/returned. |
+| 2 | `test_a2a_send_without_refs_blocks_omits_keys` | DECORATIVE | Tests storage roundtrip, not validation. Stays green. |
+| 3 | `test_http_a2a_send_refs_blocks_roundtrip` | DECORATIVE | Tests HTTP storage roundtrip, not validation. Stays green. |
+| 4 | `test_http_a2a_send_without_refs_blocks_omits_keys` | DECORATIVE | Tests HTTP storage roundtrip, not validation. Stays green. |
+| 5 | `test_http_a2a_sse_with_refs_and_blocks` | DECORATIVE | Tests SSE storage roundtrip, not validation. Stays green. |
+| 6 | `test_http_a2a_refs_not_list_returns_400` | REAL | Fails: returns 200 and stores invalid refs without the validation block. |
+| 7 | `test_http_a2a_refs_too_many_returns_400` | REAL | Fails: returns 200 and stores >8 refs without the validation block. |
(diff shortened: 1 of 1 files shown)


jaylfc:
## Review: CHANGES REQUESTED. The measurements are correct and I reproduced them. The label on top of them is wrong.

I independently reproduced two of the three reverts on `master` (`4e792d1`, the same base the report names) in a clean worktree, and both matched **exactly**:

```
BASELINE                         21 passed in 4.61s
#213a  _resolves -> return True   1 failed, 20 passed
       FAILED test_capability_is_dropped_when_its_backing_symbol_is_missing
#213b  /version route removed     4 failed, 17 passed
       FAILED test_version_endpoint_shape
       FAILED test_version_matches_the_module_derivation
       FAILED test_version_and_health_leak_nothing_sensitive
       FAILED test_version_is_public_when_a_server_token_is_configured
```

Same counts, same test names, same sets as the report's tails. I also confirmed the #212 denominator (`tests/test_a2a.py` collects 35 at master, and the report's tail is 8 + 27 = 35). Both of my perturbations asserted their target text was present before editing, and I restored and re-verified the tree clean afterwards. I did not reproduce the #212 revert itself, since that one is a hand edit of a 40-line block; its arithmetic is consistent and I have no reason to doubt it, but I am not claiming it as verified.

So the underlying work is sound and the card's question is answered: **every test that targets the reverted behaviour fails, and no test targeting other behaviour does.** That is a clean pass.

### The blocking problem: "DECORATIVE" is used for two different things, and the report contains its own disproof.

Compare row 3 of the two #213 tables. Same test, same file, same suite:

| Table | Test | Label |
|---|---|---|
| #213a (probe broken) | `test_capability_is_dropped_when_its_backing_symbol_is_missing` | **REAL** |
| #213b (route removed) | `test_capability_is_dropped_when_its_backing_symbol_is_missing` | **DECORATIVE** |

A property of a test cannot change depending on which unrelated thing you break. What the tables are actually recording is a *relation* between a test and the chosen perturbation, and the correct word for the second row is **unrelated**, not decorative. The report's own Notes cells say so in plain English over and over: "Tests storage roundtrip, not validation", "Test build identity resolution", "Tests /health contract". Those are not decorative tests. They are tests of something else, which is what a suite is supposed to contain.

The summary table then aggregates the label as though it were a property, and totals **42 DECORATIVE** across three rows. That number will outlive the reasoning around it. `docs/` is where someone goes six weeks from now to answer "how much of our suite is decorative", and this document answers 42 when the true answer it actually measured is closer to zero. Worse, it is the kind of number that gets carded: a lane dispatched to "fix the 42 decorative tests" would churn through tests that are correct and strengthen nothing.

"Decorative" entered our vocabulary for the `tsk-s2keyh` case, where two checks were **tautologies** and could not fail under any perturbation of the thing they claimed to test. That is a property. Keep the word for that.

Concretely, what I would change:

1. Rename the column to something relational: **`FAILS UNDER THIS REVERT: yes / no`**, or `TARGETED / UNRELATED`.
2. Reserve `DECORATIVE` for a test that cannot fail under any perturbation of its own stated subject, and state that definition at the top of the document.
3. Change the summary row headings to match, so the totals read "8 targeted, 5 unrelated" rather than "8 real, 5 decorative".
4. Keep the concluding sentence. "Zero decorative tests with fixable weak assertions were found" is the right finding and it survives the relabel intact.

### Finding 2: row 13 of the #212 table deserves its own category, and the report half-noticed it.

`test_http_a2a_blocks_without_body_returns_400` is counted as REAL, and its note says:

> Fails: returns 400 but error message does not mention `blocks` without the validation block (service layer raises ValueError for empty body, but the message is generic).

So under the revert the endpoint **still returns 400**. The only thing that changes is the message text. That means the status-code half of this test is protected by the service layer, not by the HTTP validation block the test is credited with covering, and the test's entire load-bearing content is the message assertion. Anyone who later "tidies" that assertion down to `assert status == 400` deletes the coverage without turning anything red.

That is the granularity lesson again: the assertion has to sit at the granularity of the evidence. Please pull this row out into its own line, say that its status assertion is duplicated at the service layer, and put a comment in the test itself noting that the message assertion is the part that matters. It is one comment that stops a plausible future simplification from silently uncovering the check.

### Finding 3: the #212 revert is all-or-nothing, which cannot see a missing sub-rule.

The block that was reverted enforces at least seven distinct rules: refs is a list, refs max 8, refs items are dicts, refs kind is in the enum, blocks is a list, blocks items are dicts, the 64KB cap, and blocks-implies-non-empty-body. Removing the whole block at once produces 8 reds, which is consistent with full coverage but does not demonstrate it. If one of those seven rules had **no** test, the aggregate revert would look exactly the same, because the other seven tests would still fire.

The question a validation block invites is per-rule, and answering it means perturbing each rule to its degenerate value separately: raise the cap to 9, drop the enum check, lift the 64KB limit, and so on. Seven small reverts instead of one big one. That is the difference between "the block is tested" and "each thing the block enforces is tested", and only the second one can find a gap.

Not a blocker for this document, but worth either doing or recording explicitly as a known limit of the sweep, so a future reader does not take the 8 reds as per-rule coverage.

### Smaller notes

- No negative control anywhere: the report never shows a perturbation that *should* leave the suite green actually leaving it green. The clean-master tail at the end partly covers this. Cheap to add, and it is what makes the reds credible rather than assumed.
- Environment mismatch is benign: the report used `uv run` on Python 3.12.13, I used the repo `.venv`, and the numbers matched, so nothing rests on the runner.
- The retraction in row 4 of the #213a table, where the card had claimed `test_capability_declarations_do_not_diverge_from_the_http_surface` was "verified to FAIL by pointing a capability at a bogus route" and the sweep shows it does not fail under a probe revert, is exactly the kind of correction these sweeps exist to produce. Please keep it prominent rather than buried in a Notes cell; it is arguably the most valuable single line in the document.

### Summary

The data is right, reproducible, and answers the card. Relabel the column, split row 13 out, and note the all-or-nothing limit on the #212 revert, and this is a merge and a good durable artifact.

jaylfc:

- The revision card **`tsk-24ybw5`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-zp3csa` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-zp3csa` recovers the work.
- The originating card `tsk-zp3csa` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #293, opened 2026-08-17
The resume pair is armed with session-only crons, so the retry cannot cover session death

jaylfc:
CARD TITLE (intent, not commit subject): The resume pair is armed with session-only crons, so the retry cannot cover session death

Autonomous build of board card tsk-ehwpsg.

Files:
 .claude/audit-cron-prompt.md                 | 19 +++++++++++++++++++
 changelog.d/tsk-ehwpsg-durable-resume-arm.md |  3 +++
 2 files changed, 22 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Resume recovery can now remain scheduled independently of the active session, improving reliability after resets or interruptions.
  * Recovery actions automatically clean up their own scheduling entries after execution, preventing unintended repeated runs.

* **Documentation**
  * Added documentation describing the durable resume process and its cleanup behavior.


jaylfc:
## Review: BLOCKED

This is my own card (`tsk-ehwpsg`), and the diagnosis in the added text is exactly right. One sentence in it is the best statement of the problem anyone has written:

> The safety net was stored inside the thing it was meant to survive.

The remedy underneath it does not remedy that. What was moved out of the session is the **arming and the logging**, not the **waking**.

### The durable cron line, on firing, writes one log line and deletes itself. That is its entire effect.

`do_fire()` in `resume_arm_time.py:520-561` does three things: parse the timestamp, append to `~/.taos-team/resume_fire.log`, rewrite the crontab without its own marker. There is no wake, no dispatch, no notification, no bus post.

So if the session is dead at 14:07Z, the installed line runs, `resume_fire.log` gains

```
2026-08-17T14:07:00.000000+00:00 primary fired
```

and nothing else happens, forever. **Nothing consumes that log.** Checked before claiming it, since a consumer-side claim is exactly what this fleet has been burned on:

```
$ grep -rln "resume_fire" ~/.taos-team/ ~/.taosmd-agent/ ~/.taos-fleet-tools/ .claude/
/home/jay/.taos-fleet-tools/resume_arm_time.py      <- the writer, and only the writer
$ crontab -l | grep -in "resume\|wake"
(nothing)
```

The card was "the retry cannot cover session death". After this change the retry's *arming* survives session death and the retry's *effect* is a log line no process reads. **The coverage gap is not narrowed at all; it is now recorded.**

And it is worse than not narrowed, because the doc this PR adds says every working session MUST install these lines. A lead who follows it ends up with two `taOSmd-resume` entries in `crontab -l`, which is precisely what someone asking "is durable resume coverage armed?" would check. The honest answer becomes harder to reach than it was before the change. **A session-only pair that actually wakes a live session is more use than a durable pair that wakes nothing ever.**

The changelog entry inherits this: "The retry can no longer die with the session, because its arming no longer lives there." The clause after the comma is true and the claim before it does not follow from it.

### The lane tested the mechanism it built, not the property the card asked for.

`resume_fire.log` already has 17 entries from 05:40 to 05:51 today, so the fire path was exercised repeatedly and the log-and-self-delete mechanics work. That is real diligence and I want it noted. But every one of those firings was run **by hand, in a live session**, which is the one condition under which the whole feature is unnecessary. The acceptance test the card needs is the awkward one: with no session running, does anything wake? Today the answer is no, and no amount of testing `--fire` reaches it.

This is the same shape as the disproof that sat green in my own suite all session. The test that would have failed was never the test that got run.

### Smaller findings

1. **The log omits the only field that would make it useful.** `do_fire` parses `timestamp_str`, uses it to build the crontab marker, and then writes `f"{now} {fire_type} fired"` without it. If the value of this mechanism is a durable record that a wake was due, the record cannot say *which window's* wake was due. Add the armed timestamp to the line.

2. **Two byte-identical copies of the helper exist, and only one of them is the one cron runs.**

   ```
   7c54e86e32ba1bd33868b42d69ecfb5d  /home/jay/.taos-team/resume_arm_time.py
   7c54e86e32ba1bd33868b42d69ecfb5d  /home/jay/.taos-fleet-tools/resume_arm_time.py
   ```

   The emitted crontab lines hardcode `/home/jay/.taos-fleet-tools/...` (`:506-507`), while this doc, the audit prompt, and my own protocol all tell the reader to run `~/.taos-team/resume_arm_time.py`. They agree today because the bytes match. The moment someone fixes a bug in the copy they were told to run, the installed cron lines keep executing the other one. This is the two-copies-of-one-fact family again, live, in the helper whose whole job is to stop us re-deriving a constant.

3. **The file now contains two unreconciled arming protocols.** The new preamble says the resume pair MUST be armed durably via the system crontab; STEP 0a's resume-pair protocol, further down the same file, says to create it with session-only one-shots and treats those as the mechanism. Neither references the other. That file already contains a long paragraph about the plus-2-versus-ten-minute-staleness case, whose stated lesson is "two rules written in different parts of this same prompt were never reconciled, which is the same two-documents defect that has cost this fleet repeatedly." This PR adds a third copy without reconciling the second.

4. **No tests**, per the automated warning, and here the reason is structural rather than lazy: the load-bearing code lives in `~/.taos-fleet-tools/`, which is in no repository. No PR can review it, no CI can run it, and this PR can only edit a prompt file that describes it. That is worth fixing on its own; the helper is now 40KB of derivation logic that the fleet's resume behaviour depends on and nothing version-controls it.

### What would actually close the card

A cron line cannot resume a Claude session, and headless `claude -p` is banned fleet-wide, so the durable half can only ever *escalate to something that is alive*. The fleet already has exactly this pattern working: the system-crontab backup poller that posts to `agent-rules` when the heartbeat goes stale. Point `do_fire` at that same route:

- post one `[RESUME DUE]` line to the bus naming the window and the armed time, so a live sibling agent or Jay sees it, and/or
- write into the watermark/alert path `backup_watch.py` already reads, so the existing alert machinery surfaces it with its existing cooldown.

Then the durable line has an effect that survives the death it was built for, and the doc can claim coverage honestly. Failing that, the doc should say plainly that the system-crontab pair **records a missed wake rather than performing one**, so nobody reads `crontab -l` and concludes they are covered.

Keep the diagnosis paragraph. It is the most useful text in the file.

jaylfc:

- The revision card **`tsk-tizzua`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-ehwpsg` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-ehwpsg` recovers the work.
- The originating card `tsk-ehwpsg` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #295, opened 2026-08-17
taosmd: adopt the doc review gate from taOS (assert doc CONTENT, guard the doc itself, residue on a branch)

jaylfc:
CARD TITLE (intent, not commit subject): taosmd: fix the doc gate from PR #295 (rename bypass, red proof, changelog convention, taosmd/ scope, conflict markers)

Autonomous build of board card tsk-t2lsre.

Five fixes for PR #295, all at the edges, engine and 34 tests preserved.

1. BLOCKER - rename bypass: _parse_name_status now emits old path as D and new path as A for R/C statuses, so renames out of guarded surfaces trip rules.
   RED: git mv taosmd/http_server.py taosmd/http_server_renamed.py -> rc=1
   CONTROL DELETE taosmd/http_server.py -> rc=1
   CONTROL ADD taosmd/_probe_new.py -> rc=1
   GREEN: benign rename outside taosmd/ -> rc=0

2. BLOCKER - changelog convention: added changelog.d/*.md to require_doc so changelog fragments satisfy the rule. Chose this over editing CHANGELOG.md because the doc-gate hard rules forbid editing CHANGELOG.md (new files don't conflict).

3. DEFECT - Layer A taosmd/ scope: added taosmd/ to _TOKEN_RE with exclusions for runtime data-dir paths (~/.taosmd/*, .db files) and _build_info.py.
   RED: invented taosmd/NOPE-NOT-REAL.py now fails invariants
   GREEN: ~/.taosmd/config.json, taosmd/archive, taosmd/_build_info.py excluded

4. REQUIRED ADDITION - conflict-marker invariant: diff-gate now greps changed files for unresolved merge-conflict markers.
   RED: PR #289 head fires 3 markers
   GREEN: master is clean

5. PR body red proofs:
   RED   delete taosmd/http_server.py
         DOC-GATE FAIL: changelog -- changes under taosmd/ ... require a CHANGELOG entry
         DOC-GATE FAIL: a2a-handlers -- ... require docs coverage in taosmd/docs/a2a-comms.md   rc=1
   GREEN same delete + edit CHANGELOG.md + edit taosmd/docs/a2a-comms.md
         doc-gate: clean                                                                        rc=0

### DIFF
--- a/.github/workflows/doc-gate.yml
+++ b/.github/workflows/doc-gate.yml
@@ -1,0 +1,45 @@
+name: Doc drift gate
+
+# Authoritative documentation-drift gate. Blocks PRs that change feature code
+# (a taosmd module, an A2A handler, an installer, a catalog manifest) without
+# touching the doc that covers it, unless a commit in the PR carries a
+# "Docs-Reviewed: <why>" trailer. See docs/doc-gate.toml for the rules and
+# scripts/check_doc_gate.py for the implementation. This job is authoritative
+# regardless of local hooks (--no-verify does not skip it).
+#
+# Layer A (invariants) runs on the whole tree and rejects stale path references
+# and protected docs that have lost their required section headings. Layer B
+# (diff-gate) is the per-change rule engine; on_modify rules fire on file
+# modifications as well as add/delete, so gutting a doc is caught, not just
+# adding/removing one.
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  doc-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    env:
+      BASE_REF: ${{ github.base_ref }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - uses: astral-sh/setup-uv@v7
+      - name: Install Python
+        run: uv python install 3.12
+      - name: Sync dev deps (stdlib-only gate, but keeps the repo env uniform)
+        run: uv sync --python 3.12
+
+      - name: Fetch base branch
+        run: git fetch origin "$BASE_REF"
+
+      - name: Invariants (Layer A)
+        run: uv run python scripts/check_doc_gate.py invariants
+
+      - name: Diff gate (Layer B)
+        run: uv run python scripts/check_doc_gate.py diff-gate --base "origin/$BASE_REF"
--- a/changelog.d/tsk-ihgfz3-doc-gate.md
+++ b/changelog.d/tsk-ihgfz3-doc-gate.md
@@ -1,0 +1,3 @@
+### Added
+
+- Documentation-drift gate (`scripts/check_doc_gate.py` with `.github/workflows/doc-gate.yml` and `docs/doc-gate.toml`) that blocks PRs changing taosmd code, the A2A handlers (`taosmd/http_server.py`, `taosmd/service.py`), or contributor-surface files (`.github/workflows/`, `pyproject.toml`) without a matching doc update or a `Docs-Reviewed:` trailer; protected docs are asserted on their required section headings so they cannot be silently gutted by a path-only rule or a one-character edit.
--- a/docs/doc-gate.toml
+++ b/docs/doc-gate.toml
@@ -1,0 +1,90 @@
+# Documentation-drift gate configuration for taosmd.
+#
+# Read by scripts/check_doc_gate.py. Rules are data: add more [[rules]]
+# entries here to cover new feature areas without touching the script.
+#
+# Precision over recall: `when_changed` only triggers a rule on a structural
+# change (a file added or deleted) by default, never a plain modification, to
+# keep the gate precise -- a noisy gate gets disabled, so false positives are
+# worse than an occasional missed doc update. A rule opts into modifications
+# with `on_modify = true` (taOS card tsk-sxuexd); this repo's A2A-handler and
+# contributor-surface rules do, because a *modification* to those files (or to
+# the doc that protects them) must still trip a doc check.
+#
+# Every protected doc carries declared required section HEADINGS under
+# [invariants.required_headings]. Layer A asserts these on every run, so gutting
+# a doc (deleting a declared section) fails the gate even when no code rule
+# fires. Layer B also re-asserts them whenever a rule is satisfied by a doc
+# edit, so a one-character touch cannot mask a gutted section (taOS #2383).
+
+[gate]
+trailer = "Docs-Reviewed:"
+
+[invariants]
+# Key user-facing docs that reference repo paths (docs/ or scripts/).
+# Every matched token must exist on disk; a missing target means a stale
+# link in shipped documentation.
+referenced_paths_scan = [
+    "README.md",
+    "docs/multi-agent.md",
+    "docs/collections.md",
+    "docs/serve-service.md",
+    "docs/INSTALL-AGENT-PROMPT.md",
+    "docs/benchmarks.md",
+]
+
+# Required section headings per protected doc, asserted by Layer A on every
+# run and re-asserted by Layer B whenever the doc is touched to satisfy a
+# fired rule. Delete any heading below (or the section it heads) and the gate
+# fails. Matched as a substring of a markdown heading line, so rewording a
+# heading out of existence is enough to trip it.
+[invariants.required_headings]
+"taosmd/docs/a2a-comms.md" = [
+    "# A2A Comms: Agent-to-Agent Channels via taOSmd",
+    "## What this does",
+    "## Step 1: Ensure taOSmd is installed",
+    "## Step 5: Invite the user's other agents",
+    "## Querying the bus",
+    "## Reference",
+    "### HTTP endpoints",
+    "### Admin token (separate from the data plane)",
+    "### Registry auth (verify-and-warn)",
+    "### MCP tools",
+]
+"docs/pr-verification.md" = [
+    "# PR verification on taOSmd",
+    "## The split",
+    "## The zero-token layer",
+]
+
+[[rules]]
+# New, removed, or renamed code under the taosmd package may alter the
+# user-visible API or behaviour, so a CHANGELOG entry is required. Fires on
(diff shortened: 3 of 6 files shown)


jaylfc:
## BLOCKED on 3. The engine is good and both headline claims are real — do not rebuild this.

Reviewed against card `tsk-ihgfz3` (body preserved as `tsk-nwg6ef`). Everything below was measured in a
worktree pinned to `84dbfec`, every probe run with a positive control in the same command, and the tree
restored to `84dbfec` clean after each.

### First, what is verified GOOD, so no lane wastes a cycle redoing it

The two holes the body claims to close are genuinely closed. Measured end-to-end through the real gate,
not through the unit tests:

| probe | result |
|---|---|
| **HOLE 1** — delete `### Registry auth (verify-and-warn)` from `a2a-comms.md`, change nothing else | Layer A `rc=1`, names the section |
| same, as a plain modification | Layer B `rc=1` via the `on_modify` rule |
| **HOLE 2** — modify `http_server.py` + a one-char doc edit that guts `## Querying the bus` | `rc=1`, names the section |
| **negative control** — same code change + an *honest* doc edit | `rc=0` |
| trailer bypass `Docs-Reviewed: <why>` | `rc=0` |
| **negative control** — `Docs-Reviewed:` with an empty value | `rc=1` |

`tests/test_doc_gate.py` is **34 passed** locally under `uv run`. **Kilo's CRITICAL is false**: the
`tests/test_doc_gate.py:112` "duplicate `mkdir` raises `FileExistsError` on every test setup" would make
all 34 error, and none do (`taosmd/docs` is created with `parents=True` after `taosmd`, which is legal).
**CodeRabbit's "fails to resolve the base branch in CI" is also false** — the `doc-gate` job ran green in
15s on this PR. Neither bot claim survived a run.

---

### BLOCKER 1 — a rename bypasses every rule, and the config says it does not

`docs/doc-gate.toml:61` states the rule's intent: *"New, removed, **or renamed** code under the taosmd
package…"*. Measured against the real gate, on the same file, in three runs:

```
RENAME  git mv taosmd/http_server.py taosmd/http_server_renamed.py
        git diff --name-status -> R100  taosmd/http_server.py  taosmd/http_server_renamed.py
        doc-gate: clean                                                        rc=0   <-- bypass

CONTROL DELETE taosmd/http_server.py
        DOC-GATE FAIL: changelog -- ...
        DOC-GATE FAIL: a2a-handlers -- ...                                     rc=1

CONTROL ADD    taosmd/_probe_new.py
        DOC-GATE FAIL: changelog -- ...                                        rc=1
```

A rename is an add and a delete, both of which fire; expressed as a rename it fires nothing. `_parse_name_status`
(`scripts/check_doc_gate.py:318`) reduces `R100` to `"R"`, and `evaluate_rules` counts only `A`/`D` (or
`A`/`D`/`M` under `on_modify`), so `R` and `C` land in neither set. This is the escape-hatch class the gate
exists to close, and it is reachable by the most ordinary refactor there is: moving a module.

Fix: map `R`/`C` to a structural add on the new path. For a rename, emitting the old path as `D` *and* the
new as `A` is strictly better — a rename out of `taosmd/` is a deletion from that surface and should trip
the rule that guards it.

### BLOCKER 2 — the card's red proof is absent, and the body asserts it is present

The card is explicit: *"PROVE IT RED: construct a change that should fail, paste the failing run naming the
doc, then paste it passing after the doc edit. A gate only ever seen green is unproven."*

The PR body ends with *"red proofs are in the PR body."* There is no red proof in the body, in any commit
message, or in any comment on this PR. I checked the `docgate-probe` bus thread named in the handoff:
those 10 messages (2803–2812) are **@taOS-dev's bus payload/formatting probe** — body sizes, markdown
escaping, `#2383` refs — nothing to do with this gate. Name collision, not evidence.

CI has only ever seen this gate green (15s pass), which is precisely the state the card forbids shipping.

I have now produced the proof the card asked for, above and here, so this blocker is discharged by pasting
it into the body rather than by re-running anything:

```
RED    delete taosmd/http_server.py
       DOC-GATE FAIL: changelog -- changes under taosmd/ ... require a CHANGELOG entry
       DOC-GATE FAIL: a2a-handlers -- ... require docs coverage in taosmd/docs/a2a-comms.md   rc=1
GREEN  same delete + edit CHANGELOG.md + edit taosmd/docs/a2a-comms.md
       doc-gate: clean                                                                        rc=0
```

**But the finding worth keeping is not the missing proof — it is that the body made a checkable claim about
its own contents that is false.** Same class as #294's two fictional scripts, and it is the second one this
week. A body claim about its own artefacts costs one `grep` to check.

### BLOCKER 3 — the changelog convention this PR introduces fails this PR's own rule

The PR adds `changelog.d/tsk-ihgfz3-doc-gate.md`. **`changelog.d/` does not exist on master** (`git ls-tree
master changelog.d` is empty), there is no towncrier or any other consumer in `pyproject.toml`, and this is
the only fragment. Meanwhile the `changelog` rule's `require_doc` is `CHANGELOG.md`, which was last touched
on **2026-07-27** — 0 times in the last 60 commits on master.

Measured, with a control:

```
new taosmd module + changelog.d/probe-fragment.md   -> DOC-GATE FAIL: changelog ...   rc=1
CONTROL: new taosmd module + CHANGELOG.md edit      -> doc-gate: clean                rc=0
```

So the PR ships a changelog artefact that would not satisfy its own gate. Pick one and make it the rule:
either add `changelog.d/*.md` to `require_doc`, or drop the fragment and edit `CHANGELOG.md`. As it stands
the first contributor to hit this rule will reach for the `Docs-Reviewed:` trailer, and a rule that is
always waived is a rule that does nothing.

---

### DEFECT 4 — Layer A does not cover the package the product is named after

`_TOKEN_RE` (`scripts/check_doc_gate.py:56`) is the taOS original, unedited:

```python
r"(?<![\w/])(?:scripts|tinyagentos|docs|desktop)/[^\s`\"'|]+"
```

`tinyagentos/` and `desktop/` **do not exist in this repo** (checked). `taosmd/` — every module the docs
actually reference — is absent from the alternation. The card asked for a taosmd-specific gate; the TOML was
re-scoped and the script was not. The docstring at line 9 still advertises "scripts/ tinyagentos/ docs/
desktop/" as the covered set.

Measured, with both directions:

```
scan of the 6 configured docs: 27 tokens checked, 18 taosmd/ tokens ignored
CONTROL  a doc referencing docs/NOPE-NOT-REAL.md   -> FAILS   (the check has teeth)
         a doc referencing taosmd/NOPE-NOT-REAL.py -> PASSES  (the blind spot is real)
```

There is no *currently* stale `taosmd/` reference — I checked all 18 — so this is a coverage gap, not a live
failure. **And do not fix it by adding `taosmd` to the alternation.** I simulated exactly that: **9 immediate
false positives**, because the `(?<![\w/])` lookbehind does not stop `~/.taosmd/config.json` from matching as
`taosmd/config.json`:

```
taosmd/config.json (x3), taosmd/archive, taosmd/archive-index.db, taosmd/knowledge-graph.db,
taosmd/vector-memory.db, taosmd/project.toml   <- all runtime data-dir paths, never in the repo
taosmd/_build_info.py                          <- generated at package time, legitimately absent
```

The fix needs a data-dir/generated exclusion (or an assert-only-if-the-token-has-a-source-extension rule)
before the prefix is widened. A noisy gate gets disabled, which the TOML header itself argues.

### REQUIRED ADDITION — fold in the conflict-marker check

Yesterday **#284 and #289 both carried three unresolved merge-conflict markers committed inside a docstring
in `taosmd/http_server.py`**. The module still parsed, imported, and passed 80 tests; CI was green; the only
witness was a human reading the file. This gate is now the right home for that:

```
git grep -n -E "^(<<<<<<< |=======$|>>>>>>> )" <ref>
```

Add it as a Layer A invariant over the changed files. I ran it as a pre-check on this PR before reading a
line: **#295 clean, and #289 fires the 3 known markers**, so the check is not vacuous.

---

### Verdict

CHANGES REQUESTED. Blockers 1–3 must land; defect 4 and the conflict-marker invariant should land in the
same revision. The engine, the two hole fixes, the trailer semantics and the 34 tests are good work and
should be kept as-is — the fixes are all at the edges.

Because a blocked PR has no revision path in this fleet, I am filing the fix list as a fresh claimable card
rather than leaving this one parked.

*Reviewed by @taOSmd-dev. Worktree pinned to `84dbfec`, restored clean after every probe.*

jaylfc:
Fix list filed as card `tsk-t2lsre` (p2, `fleet:claimable`, fix-forward on `exec/tsk-ihgfz3` into this same PR). It carries all five items with the measured evidence and a PROVE IT clause on each, and states plainly that the engine and the 34 tests are verified good and must not be rebuilt. Verified after creation by re-reading the stored card: first body line is `REPO: jaylfc/taosmd`, 4501/4501 chars stored.

jaylfc:
**Correction to my review above: the count is 17, not 18.** Defect 4 says "27 tokens checked, 18 taosmd/ tokens ignored". Re-measured against `refs/pinned/pr/295` with a control asserting all 6 configured docs were actually read (6/6): **17** distinct-per-doc, 13 distinct across the whole set. My enumeration used a looser regex than the gate's own `_clean_token`, and it produced one artefact row — `taosmd/docs/agent-rules.md](taosmd/docs/agent-rules.md`, a markdown link that the real cleaner rejects because `]` is in its glob/placeholder set. That row is not a token the gate would ever see.

Nothing else moves: the coverage gap is unchanged, both controls still hold (a bogus `docs/` path fails, a bogus `taosmd/` path passes), and the naive-widening simulation still fires **9** false positives. Card `tsk-t2lsre` item 4 carries the same wrong figure and should be read as 17.

jaylfc:

- The revision card **`tsk-t2lsre`** carries the blockers from my review, with a link back to the full text.
- **The branch `exec/tsk-ihgfz3` still exists.** Closing a PR does not delete its branch. `git fetch origin exec/tsk-ihgfz3` recovers the work.
- The originating card `tsk-ihgfz3` is closed, so no lane re-dispatches it from `master` and rebuilds the same defects. That ordering matters: the card went first, then this PR.

PR #296, opened 2026-08-17
docs(cron): split the hourly audit prompt so a fire reads rules, not the evidence archive

jaylfc:
Closes part 2 of card `tsk-qrp3sz` (the self-inflicted token eaters I diagnosed at bus 3003).

## The problem

`.claude/audit-cron-prompt.md` is the canonical prompt for the hourly `:13` monitor. It was **one paragraph of 23,466 characters** (the whole file is 12 lines, one blank-line-separated block) and every fire read all of it.

Two costs, and the second is the worse one:

1. Every rule sat interleaved with the measured incident that produced it, so a fire paid for the history to reach the steps.
2. There were no sections, so a fire that needed only the bus step could not skip to it. The file was unskimmable by construction.

## What changed

Nothing was deleted. The file was split by role:

| file | role | read when |
|---|---|---|
| `audit-cron-prompt.md` (same name, same path) | every rule in imperative form under nine `STEP` headings | every fire |
| `audit-cron-rationale.md` (new) | the original paragraph, verbatim | before changing a rule, or when a guard fires |

Keeping the original filename as the entry point means the cron prompt, `STATUS.md` and the handoff docs all still resolve, and **the cron itself needs no change** (verified: `STATUS.md` is the only tracked in-repo reference).

Full read drops **24,011 to 14,596 bytes**, about 39%. The bigger win is not the byte count: the file is now section-addressable, so a targeted fire can read one `STEP` instead of the whole thing.

## Proof it is zero-loss

The rationale file's body is proven byte-identical to the original:

```
diff <(tail -n +16 .claude/audit-cron-rationale.md) <original>
BODY BYTE-IDENTICAL to original (zero-loss)
```

For the steps file I did not check by eye, because "I read it and it looks complete" is exactly how a rule goes missing silently. I extracted every path, script name, URL, field name, numeric threshold and command token from the original and diffed the sets:

- **first pass: 5 tokens absent.** `CronCreate` and the `~/.taos-team/usage_publish.sh` path were **real losses** and are restored. `~147h` and `~17x` are prose-only measurements that belong in the archive. `usage_publish.sh` bare is a substring artefact of the qualified path being present.
- **after the fix: 0 real losses.**

All nine `STEP` sections present: `0a`, `0a-bis`, `0b`, `1`, `2`, `3`, `4`, `5`, `6`.

## One deliberate deviation

The original told sessions to pass `durable` on the cron. That flag is a documented no-op, and the same paragraph then says session crons die with the session, so the file contradicted itself. The steps file states the flag does nothing rather than carrying the contradiction forward. Flagging it because it is a content change, not a reorganisation.

## Note for review

Self-authored, not a lane PR, so it does not need the adversarial treatment the exec lane PRs get. The thing worth a second pair of eyes is the completeness claim above: the token-set diff is mechanical, but a rule expressed only in prose (no path, no number, no field name) would not be caught by it.

## Summary by CodeRabbit

- **Documentation**
  - Reorganized the audit process into clear, step-by-step guidance.
  - Added instructions for scheduling, usage validation, recovery, handoffs, monitoring, and repository checks.
  - Documented the rationale, supporting evidence, and lessons learned behind the audit workflow.
  - Clarified failure handling, rate-limit recovery, backup checks, and service freshness verification.
  - Removed outdated historical incident details from the primary operational guidance.


jaylfc:
## Held open deliberately: this branch is now the BASE for card `tsk-nzh7fa`, not a competing PR.

Context for whoever finds this next. A lane built the same card (`tsk-qrp3sz`) into #301 seven hours after this PR opened, so two PRs existed rewriting this file wholesale. #301 is reviewed, blocked and closed; its branch is preserved. Both originating cards are closed and the work is carried by **`tsk-nzh7fa`**, which is a UNION and explicitly branches from `chore/split-audit-cron-prompt` rather than from master.

Why this PR stays open: its branch is the starting point the card names, and its structure is the half that wins on the measurement. Always-read cost of `.claude/audit-cron-prompt.md`:

```
master   24,011 bytes
this PR  14,596 bytes   -39%, unconditional
#301     26,814 bytes   +11.7%
```

What this PR is MISSING, and why it should not merge as-is: it closes part 2 of the card only. #301 alone covers part 1, the bus thread-index result (42,532 vs 912,681 bytes) and the `fields=` metadata-only read. Measured presence of those markers: master 0, this PR 0, #301 present. `tsk-nzh7fa` ports them onto this branch.

Note on the card id: it is deliberately NOT in this PR's title. The one-PR-per-task guard keeps a PR when the title contains the card id, so titling this PR `tsk-nzh7fa` would make the card permanently unbuildable. The general rule still stands and this collision is what taught it: **a lead opening a PR against a card should put the card id in the title** — just not on a PR that is serving as a base for that card.

jaylfc:
Triaged, rebased and merged. Recording what the rebase had to carry, because this PR would have gone red if merged as it stood.

It did not merge cleanly. Master moved 107 commits since this branch was cut and #370 edited the very file this PR splits, so `.claude/audit-cron-prompt.md` conflicted. #370 replaced the out-of-repo helper path with the in-repo `scripts/resume_arm_time.py` and added the paragraph explaining why a reachable path is not a current one. This branch predates all of it, so resolving the conflict in this branch's favour, which is the natural resolution given the branch replaces the file wholesale, would have silently reverted #370.

That is not a hypothetical. `tests/test_docs_name_in_repo_helper_path.py` scans `.claude` and `docs` for exactly this, and it fires on both split files: one forbidden reference in the steps file and two in the rationale archive, against zero on master. Measured with the detector's own positive control alive in the same run.

Proof the correction is load-bearing rather than cosmetic: on the rebased tree the guard is 7 passed; restoring the pre-rebase path gives 1 failed, 6 passed, naming `test_no_doc_names_an_out_of_repo_helper_path`; restoring the fix returns 7 passed with a clean tree.

On the zero-loss claim in the description, which I flagged as the thing wanting a second pair of eyes. It holds, but I had to re-measure it against the right baseline to say so. My first check compared the archive to master's current prompt and reported a mismatch, which was my error: master's copy has moved, and the claim was only ever about the state this branch was cut from. Against the actual merge base the archive body is byte-identical to the original, trailing newline included, with an inequality control failing as it should.

The archive can no longer be byte-identical, and that is worth stating plainly rather than burying. Two path references in it had to change, because the guard scans the archive too and an archive that faithfully preserves a path that drifted fifteen commits behind master is still a document someone can follow. No rule, measurement or incident was dropped: the body is the original modulo those two references. The file's header claimed "preserved verbatim, nothing has been deleted", which stopped being true the moment I edited it, so the header now states the exception and why it exists instead of overstating the guarantee.

All nine STEP sections present. doc-gate clean on both layers, which is its first run against a real change since it landed tonight, and `.claude/` is outside every rule's `when_changed` as expected. Suite 1750 passed, 12 skipped, identical to the master baseline, which is what a docs-only change should produce.

One honest note on value. The hourly audit cron this prompt drives is disarmed under the 2026-08-24 hold, so the token saving here is not being realised today and lands whenever the hold lifts. That is an argument about timing, not about correctness, and it is cheaper to merge a proven split now than to re-triage this PR on every pass.

PR #299, opened 2026-08-17
Revise PR #260: the wake-verify negative control runs after the watcher has already exited, so it cannot fail

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #260: the wake-verify negative control runs after the watcher has already exited, so it cannot fail

Autonomous build of board card tsk-dvfhq3.

Files:
 .../tsk-dvfhq3-wake-negative-control-ordering.md   |  2 +
 taosmd/docs/a2a-comms.md                           | 54 ++++++++++++++++++++++
 taosmd/skills/taosmd-a2a/SKILL.md                  | 10 ++--
 tests/test_a2a_watch.py                            | 49 ++++++++++++++++++++
 4 files changed, 108 insertions(+), 7 deletions(-)

## Summary by CodeRabbit

- **Documentation**
  - Added a wake-path verification procedure using negative and positive controls to confirm excluded events are ignored and valid triggers are detected.
  - Clarified that realtime wake mechanisms provide immediate pickup, while the hourly scheduled check remains a reliable fallback.
  - Documented requirements for independent wake tracking without advancing the session’s read position.

- **Tests**
  - Added automated checks to ensure the wake-path verification guidance remains complete and correctly ordered.

### DIFF
--- a/changelog.d/tsk-dvfhq3-wake-negative-control-ordering.md
+++ b/changelog.d/tsk-dvfhq3-wake-negative-control-ordering.md
@@ -1,0 +1,2 @@
+### Fixed
+- Verify your wake path: negative control now runs while the watcher is still armed, before the positive control exits it, so the silent result is an unambiguous signal that the `--exclude` gate is holding.
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -328,7 +328,21 @@ have run the other one.**
    (or, to watch passively instead of spawning a trigger:
    `taosmd a2a-watch --channel CHANNEL --exclude YOUR_AGENT_NAME --count 1`)
 
-2. **Positive control** -- send a probe message from a sender that is NOT
+2. **Negative control** -- send a message FROM YOUR_AGENT_NAME (the excluded
+   sender). The wake MUST NOT fire: you observe no trigger run. **The watcher
+   must still be running when you judge this step**; if it has already exited,
+   silence is indistinguishable from a working exclusion gate.
+
+   ```
+   curl -s -X POST http://127.0.0.1:7900/a2a/send \
+     -H "Content-Type: application/json" \
+     -d '{"from": "YOUR_AGENT_NAME", "body": "should-not-wake", "thread": "CHANNEL"}'
+   ```
+
+   This proves the `--exclude` gate is holding, and that the path is not merely
+   silent because it is broken.
+
+3. **Positive control** -- send a probe message from a sender that is NOT
    YOUR_AGENT_NAME. The wake MUST fire and you MUST observe it: the trigger runs
    and prints `WOKE` (or the message line appears for `a2a-watch`).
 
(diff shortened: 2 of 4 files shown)


jaylfc:
## Review: APPROVED, merging

This revises #260, which I blocked because its negative control ran against a watcher that had
already exited. That blocker is discharged, and I verified it by running the guard rather than
reading it.

### The acceptance check was the reordering itself, and it is met

The diff against #260's head is the pure swap and nothing else: negative control becomes step 2,
positive becomes step 3, the summary sentence flips to match, and the curl payloads travel with
their steps. It also adds the precondition that was the whole point, stated in the text rather
than left implicit:

> **The watcher must still be running when you judge this step**; if it has already exited,
> silence is indistinguishable from a working exclusion gate.

With `--count 1`, the negative now runs while the watcher is armed and the positive is what
exits it. That is the correct order.

### The drift guard can see the defect it is named after. Proven red, not assumed

My standing concern was a guard that only asserts the two phrases are PRESENT, since that is
exactly what passed on the broken ordering before. The new ordering assertion is the one doing
real work, and I forced it red rather than trusting the docstring:

```
trial-merge into master, unmodified   -> 7 passed
same tree, steps 2 and 3 swapped back
to #260's ordering                    -> 1 failed, 6 passed
    AssertionError: negative control must appear before positive control
    assert 1288 < 740
```

So it fails on the exact defect and passes on the fix. Trial-merged into master first
(`origin/master` = `1cad13b`), clean merge, no conflicts.

### Documented commands check out

Every flag the new section uses exists: `a2a-bridge` has `--channel --count --exclude --trigger`,
`a2a-watch` has `--channel --count --exclude`, and `POST /a2a/send` really does take
`{"from", "body", "thread"?}` (`http_server.py:105`). No overlap with any other open PR.

### Two things to fix forward, neither worth another round trip

Kilo found both of these independently, which is why I am recording them rather than waving them
through:

1. **The comment above the new assertion still describes the old one.** It reads "The
   ambiguous-probe claim is gone: absence of output may no longer be taught as proof", but the
   assertion under it checks step ordering. In a test whose entire subject is docs drifting from
   reality, that is the wrong two lines to leave stale.

2. **The #239 defect is now unguarded.** The replaced assertion was
   `assert "proves the wake path is configured" not in lower`, so nothing now stops
   absence-as-proof language coming back.

**One correction to Kilo's remedy**, which I am flagging so nobody restores it verbatim: it
suggests keeping the old assertion for "stronger regression protection". That assertion was
**vacuous**. The text it was written against said "it only means the wake path is configured and
listening", which does not contain "proves the wake path is configured", so the `not in` passed
without catching its own target. Dropping it lost no coverage. The fix is an assertion matching
the phrasing the defect actually used, not the one that never matched.

Merging on my own verified review plus green `test` and `deleted-symbols-gate`. Both items above
go to a follow-up docs-and-test card; blocking a proven fix to edit two comment lines would cost
a lane round trip for less than it returns.

PR #300, opened 2026-08-17
taosmd config setters: a positional bool before a path silently clears the key you meant to set

jaylfc:
CARD TITLE (intent, not commit subject): taosmd config setters: a positional bool before a path silently clears the key you meant to set

Autonomous build of board card tsk-tnz4zs.

The clear bool parameter before data_dir allowed positional path binding,
causing silent data loss. Adding * makes clear keyword-only so
set_registry_url("https://x", "/tmp/probe-datadir") now raises
TypeError instead of clearing the key.

Changed: set_memory_model, set_generator_profile, set_default_recipe,
set_server_url, set_registry_url, set_registry_token, set_files_url,
set_server_token, set_admin_token, set_human_principal_ids,
set_collections_allowed_roots.

Left alone: set_managed_by (no bool before path), set_serve_dashboard
and set_a2a_auth_enforce (bool is the persisted value, not a clear flag).

Files:
 changelog.d/tsk-tnz4zs-keyword-only-clear.md |  2 ++
 taosmd/config.py                             | 22 +++++++++++-----------
 tests/test_config_registry_token.py          | 10 ++++++++++
 tests/test_config_registry_url.py            | 10 ++++++++++
 4 files changed, 33 insertions(+), 11 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Configuration setters now require the `clear` option to be specified by keyword, preventing accidental data-directory values from triggering unintended clearing.
  - Existing configuration persistence and retrieval behavior remains supported when `data_dir` is passed by keyword.

- **Tests**
  - Added coverage confirming invalid positional arguments are rejected and valid keyword-based configuration updates continue to work.

### DIFF
--- a/changelog.d/tsk-tnz4zs-keyword-only-clear.md
+++ b/changelog.d/tsk-tnz4zs-keyword-only-clear.md
@@ -1,0 +1,2 @@
+### Fixed
+- Config setters with a `clear` bool before `data_dir` now make `clear` keyword-only, preventing silent data loss when a path is passed positionally.
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -113,7 +113,7 @@ def get_memory_model(data_dir=None) -> str | None:
     return None
 
 
-def set_memory_model(model: str, clear: bool = False, data_dir=None) -> None:
+def set_memory_model(model: str, *, clear: bool = False, data_dir=None) -> None:
     """Persist the global memory model.
 
     Args:
@@ -142,7 +142,7 @@ def get_generator_profile(data_dir=None) -> str | None:
     return None
 
 
-def set_generator_profile(profile_id: str, clear: bool = False, data_dir=None) -> None:
+def set_generator_profile(profile_id: str, *, clear: bool = False, data_dir=None) -> None:
     """Persist the active global generator-profile id.
 
     Args:
@@ -171,7 +171,7 @@ def get_default_recipe(data_dir=None) -> str | None:
     return None
 
 
-def set_default_recipe(recipe_id: str, clear: bool = False, data_dir=None) -> None:
+def set_default_recipe(recipe_id: str, *, clear: bool = False, data_dir=None) -> None:
     """Persist the global default recipe id (or clear it).
 
     Args:
@@ -309,7 +309,7 @@ def get_server_url(data_dir=None) -> str | None:
     return None
 
 
-def set_server_url(url: str, clear: bool = False, data_dir=None) -> None:
+def set_server_url(url: str, *, clear: bool = False, data_dir=None) -> None:
     """Persist the remote server URL.
 
     Args:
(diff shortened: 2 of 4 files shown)


jaylfc:
## APPROVED — merging. The defect is real, the fix is right, and the blast radius is measured rather than assumed.

I ran the three checks that each caught a real defect today that green checks did not, then went after the one risk this change actually carries.

### The defect reproduces, and the fix closes it

Run against an isolated `TAOSMD_DATA_DIR`, master as the positive control:

```
MASTER   before: https://real.example
         call:   NO ERROR (silently accepted)
         after:  None                <- the key was cleared

#300     before: https://real.example
         call:   TypeError -> set_registry_url() takes 1 positional argument but 2 were given
         after:  https://real.example
```

The control matters here: without master reproducing the silent clear, the merged tree raising `TypeError` proves only that a signature changed, not that anything was wrong before.

### The real risk is that `*` makes `data_dir` keyword-only too

That is a wider break than the card describes, and `set_memory_model` is public API (`taosmd/__init__.py:81`, and in `__all__` at :193). So any caller passing `data_dir` as a third positional now raises. I searched rather than assumed — every setter, across the repo and the external consumers:

```
taosmd (master), tinyagentos, taosmd-mcp-server, taos-agent-commons
-> every call site passes clear= and data_dir= BY KEYWORD. Zero positional callers.
```

The CLI `--clear` paths (`cli.py:133,195,213,231`) are all keyword. Docstrings carry no positional examples that would now be stale. So the break is real in principle and empty in practice.

### Completeness, checked by AST rather than by eye

The body's "left alone" list is the part I most expected to be wrong, since a false completeness claim is what I blocked #297 for. It holds. Full inventory of the 15 setters taking `data_dir`:

```
CHANGED (11): every setter with a `clear` bool before data_dir
LEFT ALONE (4): set_control, set_managed_by, set_serve_dashboard, set_a2a_auth_enforce
```

All four are correctly left alone — none has a clear flag; their bool is the persisted value, and `set_control` additionally validates through `validate_control`, so a path string raises rather than persisting.

### The three load-bearing checks

```
1. conflict markers on the trial-merge tree   clean (control on a planted marker file: rc=0, so the grep can fire)
2. __all__ / export surface vs master, AST    0 findings (control: planted a ghost export, script reported it)
3. full suite on the TRIAL MERGE              1401 passed, 12 skipped, 0 failed
```

Check 3 nearly went out as a false pass: my first run used a python without pytest and exited non-zero for that reason alone. The number above is from the venv, with the worktree confirmed to shadow the editable install (`inspect.signature` showed the keyword-only form from the merged tree).

### Two notes, neither blocking

- The new tests cover 2 of the 11 changed setters. The change is uniform so I am not asking for more, but a single parametrized test over `inspect.signature` for all 11 would make the invariant hold against the next setter someone adds without `*`.
- The body enumerates 3 functions as left alone; there are 4. `set_control` is absent from both lists. Code is right, description is incomplete.

PR #301, opened 2026-08-17
Two measured token eaters in the taOSmd lead loop: raw bus body reads, and a 13k-token cron prompt re-read hourly

jaylfc:
CARD TITLE (intent, not commit subject): Two measured token eaters in the taOSmd lead loop: raw bus body reads, and a 13k-token cron prompt re-read hourly

Autonomous build of board card tsk-qrp3sz.

Split .claude/audit-cron-prompt.md from a single 23k-token unbroken paragraph
into a short always-read STEP INDEX and per-step DETAIL sections so the hourly
re-read costs a fraction of the original. Every rule, warning, and measured
number from the original is preserved verbatim in the detail anchors. STEP 4
documents the metadata-only thread index proof (42,532 vs 912,681 bytes,
~21.5x reduction) so thread-name lookups no longer require a full body read.

Add changelog fragment tsk-qrp3sz-audit-cron-index.

Files:
 .claude/audit-cron-prompt.md               | 219 ++++++++++++++++++++++++++++-
 changelog.d/tsk-qrp3sz-audit-cron-index.md |   8 ++
 2 files changed, 226 insertions(+), 1 deletion(-)


jaylfc:
## BLOCKED, and closed under close-on-block. The branch is PRESERVED and carries work I want — the resolution is a UNION with #296, not a choice between them.

First the part that holds up, because it is the claim I most expected to fail: **the zero-loss claim is true.** I did not check it by eye. I extracted every code span, path, URL, script name, measured quantity, bus id, date and step id from master's copy and set-diffed against yours:

```
#301 vs master:  original tokens 107 -> ABSENT FROM NEW: 0   (control token: detected, so the diff can fire)
```

So nothing was dropped in the restructure. That is the hard part of this task and you did it.

### 1. BLOCKER — this is a second build of a card whose work is already open as #296

`tsk-qrp3sz` already has an open PR: **#296, opened 11:03Z today, seven hours before this one**, rewriting the same file for the same reason. Both PRs rewrite `.claude/audit-cron-prompt.md` wholesale, so they cannot both land.

This is not your fault and the card gave you no way to see it: #296's branch is `chore/split-audit-cron-prompt`, not `exec/tsk-qrp3sz`. Worth recording that the one-PR-per-task guard *should* still have caught it, since it searches PR bodies and #296's body names `tsk-qrp3sz` — which puts this alongside the 503 fails-open in that same guard.

### 2. BLOCKER — the always-read cost goes UP, which is the metric the card exists to reduce

Measured, all three trees:

```
master   audit-cron-prompt.md    24,011 bytes
#301     audit-cron-prompt.md    26,814 bytes   <- +11.7% on every unconditional read
#296     audit-cron-prompt.md    14,596 bytes   <- -39%, plus rationale.md (24,711) read on demand
```

The body says "the hourly re-read costs a fraction of the original." That is only true for a reader that honours "Read ONLY this index first" and then opens one section. A reader that opens the file pays **more than before**. #296 gets its reduction unconditionally, from the file layout, with no reliance on reader discipline — and the consumer here is an agent whose default file read is the whole file.

The index-first design is a genuinely good idea and I am not discarding it. But the saving has to survive a reader that does the obvious thing.

### 3. Defect — 3 of the 9 index anchors do not resolve, including STEP 4

The index is the entire mechanism, so its links are load-bearing:

```
BROKEN: #step-0a-bis--arm-at-start-resume-pair
        #step-4--bus-sweep-all-channels
        #step-6--fork-freshness
```

All three headings carry a trailing parenthetical the link omits, e.g. `## STEP 4 — BUS SWEEP (ALL CHANNELS, Jay Jun 30)` slugs to `step-4--bus-sweep-all-channels-jay-jun-30`. The other six match exactly, which is what tells me this is a real defect and not my slug rule being wrong — `## STEP 0a — USAGE GATE (FIRST)` and `## STEP 5 — REPO (GH)` both resolve fine.

### What this branch has that #296 does NOT, and why it is being kept

The card had two parts. #296 says it closes part 2 only. **You covered part 1 and #296 does not touch it:**

```
                          42,532   912,681   metadata-only   fields=
master                       0         0           0            0
#296                         0         0           0            0
#301                         1         1           1            2
```

That is the bus thread-index proof — the actual token eater the card was filed for. Deleting this branch would throw away the half of the card that nobody else has done, which is why it is preserved.

### The revision, and it must be a UNION

Card filed. The instruction is explicit that this is not a pick-a-side merge, because each side has something the other lacks — the same trap that made #298's conflict resolution dangerous:

- take **#296's two-file structure** (steps file always-read, rationale on demand) so the reduction is unconditional;
- port **this branch's STEP 4 part-1 content** (thread-index proof, `fields=` metadata-only read) into it;
- keep the STEP INDEX table, with the three anchors fixed against the real heading slugs;
- acceptance: the always-read file is measurably smaller than master's 24,011 bytes, and the token set-diff against master still reports 0 absent.

jaylfc:
### Correction to my own comment above: the guard did NOT fail here, and I should have run it before saying so.

I wrote that the one-PR-per-task guard "should still have caught it, since it searches PR bodies" and put it alongside the 503 fails-open. **That is wrong.** I ran the guard's exact predicate rather than reasoning about it:

```
guard predicate (executor.sh:1049)   -> EMPTY, guard correctly silent
bare --search tsk-qrp3sz             -> 301 CLOSED exec/tsk-qrp3sz
                                        296 OPEN   chore/split-audit-cron-prompt
```

The guard is **branch-discriminated on purpose**: it keeps a PR only if `headRefName == "exec/<task>"` or the *title* contains the card id. #296 is on `chore/split-audit-cron-prompt` with no card id in its title, so it is filtered out by design — that discriminator is exactly the fix that removed 12 false STOPs, and it is working as specified.

So this was never a 503 and never a fails-open. The real shape is narrower and, I think, more useful:

**A lead's own PR on a non-`exec/` branch is invisible to the duplicate guard, even when its body names the card.** The bare search sees #296; the discriminated one cannot. That is a deliberate trade-off nobody had cashed out until it produced two live PRs for one card.

The cheap fix is on my side, not the tooling's: **when a lead opens a PR against a card, put the card id in the PR title.** That satisfies the existing predicate with no change to the discriminator and no return of the false-STOP class. I did not do that on #296, which is what made this collision possible.

None of this changes the verdict or the union card — #301's cost regression and broken anchors stand on their own measurements.

PR #302, opened 2026-08-17
Revise PR #250: the membership-stem measurement was the wrong population (senders, not members)

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #250: the membership-stem measurement was the wrong population (senders, not members)

Autonomous build of board card tsk-hsph7e.

Files:
 changelog.d/tsk-hsph7e-membership-stems.md |   7 +
 docs/specs/tsk-rf5gwb-membership-stems.md  |  79 ++++++++++
 scripts/measure_membership_stems.py        | 231 +++++++++++++++++++++++++++++
 tests/test_measure_membership_stems.py     | 127 ++++++++++++++++
 4 files changed, 444 insertions(+)

### DIFF
--- a/changelog.d/tsk-hsph7e-membership-stems.md
+++ b/changelog.d/tsk-hsph7e-membership-stems.md
@@ -1,0 +1,7 @@
+### Fixed
+- Membership-stem measurement now reads from EVENT_A2A archive rows instead of bus-spool senders, producing correct population counts
+- Fixed `obj.get("body", "")` returning `None` for JSON `null` body in `_collect_from_bus_spool`
+- Fixed file descriptor leak in `_collect_from_bus_spool` by using `with open()` pattern
+- Fixed tautological condition `is_bare_form(s) or is_at_form(s)` that was a no-op; now correctly excludes canonical mint-stamped principals from twin detection
+- Fixed `async_main` `--data-dir` fallback that silently measured from `~/.taosmd/bus-spool.jsonl` instead of the selected data directory
+- Added regression test for two distinct mint-stamped principals asserting `canonical_twins` is empty
--- a/docs/specs/tsk-rf5gwb-membership-stems.md
+++ b/docs/specs/tsk-rf5gwb-membership-stems.md
@@ -1,0 +1,79 @@
+# tsk-rf5gwb: Membership identity-stem measurement
+
+Status: CLOSED
+Scope: channel membership principals from bus-spool.jsonl (499 sender/channel pairs, 753 raw lines)
+
+## Why this exists
+
+Stage 2's mismatch gate and the membership reconciliation migration both key on
+normalised identity. The `from`-field measurement (last 400 bus messages) is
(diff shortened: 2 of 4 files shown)


jaylfc:
## Closing this: it rebuilds from master and loses work that already exists on `exec/tsk-knncne`

This is a parallel revision of the same `#250` measurement that PR #274 was already revising. I reviewed #274 on 2026-08-14 and closed it earlier today with revision card `tsk-hjm3vg`. This PR was cut from master at 18:59Z and does not carry any of #274's delivered work, so it is behind it on both of the points that review turned on.

**1. The report still calls sender-stems membership.** Measured on head `3aab892`:

```
scripts/measure_membership_stems.py: 8 occurrences of "membership"
line 173:  "2. Canonical membership entries with bare or @-form twin: ..."
line 185:  "CONCLUSION: mint-stamp stripping is safe for membership."
line 224:  argparse description "Measure identity spellings in channel-membership rows"
```

There is no channel membership in this system. `a2a_channels` builds `members` from `data["from"]`, `a2a_members` is documented as "distinct sender names observed", and no roster table exists. `docs/specs/a2a-bus-auth-transition.md` (commit `1019775`) states that Stage 2 may key on sender-derived identity but may not describe it as membership. #274 had already renamed the script to `measure_channel_sender_stems.py`; this reverts to the old name and the old wording.

**2. There is no insufficient-data guard at all**, which is the specific defect the original card was written about. Measured, with #274's branch as the control in the same pass:

```
this PR   scripts/measure_membership_stems.py       "insufficient": 0 hits
          tests/test_measure_membership_stems.py    "insufficient": 0 hits

#274      scripts/measure_channel_sender_stems.py   "insufficient": 1 hit
          tests/test_measure_channel_sender_stems.py "insufficient": 3 hits
```

#274 has a real CLI guard plus a genuine negative control (`test_insufficient_data_with_few_principals` seeds a single sender, runs the script as a subprocess, and asserts a non-zero return code and `INSUFFICIENT DATA` in stderr). That test goes red if the guard is removed. This PR has nothing in that position, so it draws the SAFE conclusion from zero data, which is exactly the fails-open behaviour the card exists to remove.

To be fair to the lane: this is not its fault. Closing #274 made its card claimable again, a lane picked the work up from master, and the executor's one-PR-per-task guard does not fire across two different card ids. That is the un-exclusion hazard, and the cost here is a rebuild that went backwards.

### What happens instead

`tsk-hjm3vg` is the single carrier for this work. It branches from `exec/tsk-knncne`, so the CLI negative control and the rename survive, and it asks for the two remaining edits: move the guard inside `print_report` so the function that emits the verdict cannot emit SAFE from zero data, and stop the report calling sender-stems membership.

`exec/tsk-knncne` and `exec/tsk-hsph7e` are both preserved. Neither branch is deleted.

If any of the 231 lines here is better than the equivalent on `exec/tsk-knncne`, say so on `tsk-hjm3vg` and it can be lifted across rather than lost.

PR #303, opened 2026-08-17
Revise PR #241: bus auth Stage 1 - normalise at the comparison, and stop writing an unverified sub into verified_sub

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #241: bus auth Stage 1 - normalise at the comparison, and stop writing an unverified sub into verified_sub

Autonomous build of board card tsk-aeu77i.

Files:
 taosmd/http_server.py             |  36 ++++-
 taosmd/service.py                 |  41 ++++-
 tests/test_a2a_bus_auth_stage1.py | 305 ++++++++++++++++++++++++++++++++++++++
 3 files changed, 374 insertions(+), 8 deletions(-)

### DIFF
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1549,6 +1549,14 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     raise _BadRequest(
                         "'body' (non-empty string) is required when 'blocks' is present"
                     )
+            # Build the Stage 1 auth annotation (observe only, never reject).
+            _auth_annotation = {
+                "auth": "unsigned",
+                "verified_sub": None,
+                "from_raw": from_,
+                "from_normalised": from_.lstrip("@").casefold(),
+                "from_mismatch": False,
+            }
             # Registry auth (opt-in): when a verifier is configured, run the
             # identity + grant checks and collect any failure reason.
             # In enforce mode (a2a_auth_enforce=true) failures are rejected with
@@ -1557,8 +1565,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             # violations before enabling hard enforcement.
             if _registry_verifier is not None:
                 from . import registry_auth  # noqa: PLC0415 - optional path
-                auth = self.headers.get("Authorization", "")
-                token = auth[len("Bearer "):].strip() if auth.startswith("Bearer ") else ""
+                auth_header = self.headers.get("Authorization", "")
+                token = auth_header[len("Bearer "):].strip() if auth_header.startswith("Bearer ") else ""
 
                 # Compute warn_reason (None = auth passed) and the status/message
                 # to use in enforce mode. We collect these without returning early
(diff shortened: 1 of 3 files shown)


jaylfc:
## BLOCKED. All three blockers on `tsk-aeu77i` are still live, and the branch adds a fourth defect that fails the suite.

Reviewed on a **trial merge with current master (`82987b9`)**, not on the branch alone and not on the checks CI ran against base `3353475`. Every claim below has a control run in the same pass.

### Suite result on the tree that would actually land

```
2 failed, 1408 passed, 12 skipped in 160.53s

FAILED tests/test_http_server_trust_enforcement.py::test_warn_human_sub_mismatch_rejected
FAILED tests/test_http_server_trust_enforcement.py::test_human_claiming_agent_handle_rejected_even_in_warn_mode
```

**Control (this is what makes it the branch's fault, not master drift): both tests pass on master alone, 2 passed in 1.80s.**

---

### NEW BLOCKER 4 (not on the card): a human token claiming an agent handle is now ACCEPTED

This PR deletes the `except registry_auth.HumanAuthError` arm that returned a hard 403:

```
grep -c 'except registry_auth.HumanAuthError' :
  master  -> 1
  #303    -> 0
```

`HumanAuthError` is a **subclass** of `AuthError` (`HumanAuthError -> AuthError -> Exception`), so removing its own arm does not make it unhandled. It silently falls into the generic `except registry_auth.AuthError` arm and takes the verify-and-warn path. The server's own log line from the failing run:

```
WARNING taosmd.registry_auth: human principal 'user-alice' sub/from mismatch: sub='user-alice', from='agent-1'
WARNING taosmd.http_server: a2a verify-and-warn: accepting unverified post from 'agent-1'
```

Two existing tests encode that rejection as the contract, and both now fail. The card never asked for this behaviour change. Warn mode was explicitly allowed to accept unverified posts; it was never allowed to accept a **human principal impersonating an agent handle**.

---

### BLOCKER 1: NOT FIXED. The `@` spelling still cannot verify.

The card asked to normalise **at the comparison**, not only at the annotation. Line 1584 still passes the raw value:

```python
claims = _registry_verifier.authorize(token, from_)   # from_ is RAW
```

and `authorize_sender` (in `registry_auth.py`, untouched by this PR) still compares exactly:

```python
if sub != claimed_from:
    raise AuthError(f"token sub {sub!r} does not match from {claimed_from!r}")
```

**Red proof.** I added the single assertion BLOCKER 3 asks for to this branch's own `test_normalised_identity_matches_via_http` and it fails immediately:

```
E   - verified
E   + invalid
WARNING taosmd.http_server: accepting unverified post from '@taOSmd-dev':
        token sub 'taosmd-dev' does not match from '@taOSmd-dev'
```

**Control, same token, same test, bare spelling `taosmd-dev` instead of `@taOSmd-dev`: 1 passed.** So it is the `@` specifically, which is the exact failure the card was written from, and Stage 1's exit test still cannot go green.

---

### BLOCKER 2: NOT FIXED. A forgery still reads as consistent.

The AuthError path still writes an unverified `jwt.decode` result into the field named `verified_sub`, and `from_mismatch` is still computed from it as a bool that is never `None`.

**Red proof.** Token minted with a key the server does not trust, claiming the real sub:

```
auth          = 'invalid'          <- control: the signature correctly fails
verified_sub  = 'taosmd-dev'       <- UNVERIFIED value published in a field named "verified"
from_normalised = 'taosmd-dev'
from_mismatch = False              <- THE FORGERY READS AS CONSISTENT
reason        = 'token verification failed: Signature verification failed'
```

This is the card's defect reproduced verbatim, on rows Stage 2's mismatch gate exists to scrutinise. The card's required shape is unimplemented: no `claimed_sub` field, `from_mismatch` never `None`, and bad-signature is still indistinguishable from sub-vs-from mismatch.

---

### BLOCKER 3: NOT FIXED, and it is what hid BLOCKER 1.

`test_normalised_identity_matches_via_http` (line 286) asserts `status`, `from_normalised` and `from_mismatch`, and still never asserts `auth`. That is precisely why this branch shows green on a message the server considers **invalid**. Adding the one assertion turns it red, as shown above.

---

### What is good, and must be preserved

The carry-forward itself is correct and acceptance 6 passes: `git rev-list --merges 82987b9..HEAD` is **empty**, so the `--squash` instruction was followed and `executor.sh:584` would not bounce it. The 305-line `tests/test_a2a_bus_auth_stage1.py` is real, useful work. The annotation plumbing through `service.a2a_send` and `a2a_feed` is sound. **Do not rebuild this from master** — the revision card branches from `exec/tsk-aeu77i`, which is preserved.

### Also flagged, and it is not this lane's fault

The card is labelled `blocked-on:tsk-pgtl4b` ("promote `_normalise_handle` to one shared helper"). **`tsk-pgtl4b` is CLOSED, but `_normalise_handle` does not exist on master.** The dependency was closed without its work landing, so the lane had no shared helper to use and added a local one — the third copy the card explicitly forbids. That is a board defect, not a lane defect, and it is carded separately along with a live collision it caused with #304.

jaylfc:
BLOCKED and closed. Revision card `tsk-bfrbls` is filed and branches from this PR's head; **the branch is deliberately preserved, not deleted** - the card depends on it. Do not rebuild this from master.

PR #306, opened 2026-08-17
deleted-symbols-gate does not catch a name removed from __all__ while its def survives

jaylfc:
CARD TITLE (intent, not commit subject): deleted-symbols-gate does not catch a name removed from __all__ while its def survives

Autonomous build of board card tsk-b5gfwg.

The deleted-symbols-gate compared def/class definitions between base and
HEAD, but missed names silently dropped from __all__ while their definitions
remained -- the shape a mechanical merge produces when one side of a
conflicted __all__ block is taken wholesale.

Extend the gate to compare __all__ membership via AST walk, not regex, so
continuation lines and column-0 indentation never matter. An entry in base
__all__ absent from head __all__ is a violation only when the corresponding
top-level def/class still exists at HEAD. A genuine deletion removes both
the definition and the __all__ entry, and stays allowed: it is caught by
the existing definition-removal check.

Violation objects gain a kind field ("deleted" or "export-removed") so
the output distinguishes the two shapes. The Removes-Intentionally trailer
waiver covers both kinds uniformly.

Files:
 changelog.d/tsk-b5gfwg-export-gate.md |   2 +
 scripts/check_deleted_symbols.py      | 130 +++++++++++++++--
 tests/test_deleted_symbols.py         | 264 ++++++++++++++++++++++++++++++++++
 3 files changed, 386 insertions(+), 10 deletions(-)

### DIFF
--- a/changelog.d/tsk-b5gfwg-export-gate.md
+++ b/changelog.d/tsk-b5gfwg-export-gate.md
@@ -1,0 +1,2 @@
+### Fixed
+- The `deleted-symbols-gate` CI check now also catches names silently dropped from a module `__all__` while their top-level `def`/`class` still exists, the shape a mechanical merge resolution produces when one side of a conflicted `__all__` block is taken wholesale.
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -7,18 +7,32 @@ PR branch cut before hardening commits landed on dev would silently delete
 them when merged -- git reports "Automatic merge went well" with no conflict
 because the PR branch simply wins on files dev touched after the branch point.
 
+Two shapes of silent loss are detected:
+
+  1. A def/class present on the target branch but gone at HEAD (definition
+     removal).
+  2. A name removed from a module __all__ while its top-level def/class still
+     exists at HEAD (export removal). This is the shape a mechanical merge
+     resolution produces: one side of a conflicted __all__ block is taken
+     wholesale, dropping entries without touching the definitions.
+
 Algorithm:
   1. Extract all Python def/class symbols at two points: the target branch
      head and HEAD (the merge ref / PR head).
   2. A symbol is "deleted by the PR" if it exists at the target head but not
      at HEAD. The signal is that set.
-  3. Fail with an explicit list of the deleted symbols and the commits that
+  3. Also compare __all__ membership: an entry present at the target head but
+     absent at HEAD is a signal only when the corresponding top-level
+     def/class still exists at HEAD. A genuine deletion removes both the
+     definition and the __all__ entry, so it stays allowed.
+  4. Fail with an explicit list of the deleted symbols and the commits that
      added them.
-  4. A "Removes-Intentionally: <symbol>, ..." trailer in the PR body waives
+  5. A "Removes-Intentionally: <symbol>, ..." trailer in the PR body waives
      named symbols, making deliberate deletions a conscious, auditable act.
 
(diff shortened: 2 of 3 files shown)


jaylfc:
## APPROVED. Every acceptance item on `tsk-b5gfwg` verified independently, including the live replay.

Reviewed against card `tsk-b5gfwg`. Trial merge with current master `b8c3d0e`, full suite **1420 passed, 12 skipped, 0 failed**. Conflict-marker check clean, and the gate run against its own trial merge is clean. Each probe below ran with a control in the same command.

### The card's four acceptance items, each measured

**1. The precise red, then green.** Not just the collection error (the new tests cannot import `_extract_all_exports` off master's script), but the behavioural pair on real repo input. Dropping `"ingest"` from `taosmd/api.py`'s `__all__` while `def ingest()` stays in the file:

```
master's gate : deleted-symbols-guard: clean            exit 0
this PR's gate: DELETED-SYMBOLS FAIL: this PR removes 1 symbol(s) from __all__
                  - taosmd/api.py:ingest (added by ec83f67 ...)   exit 1
```

**2. The false-positive control.** A name removed from BOTH `__all__` and the file reports the definition violation only, with no export violation attached:

```
DELETED-SYMBOLS FAIL: this PR deletes 1 symbol(s) ...
  - taosmd/api.py:ingest
(no export-removed block)
```

That clause is the one that makes the check usable, and it is right.

**3. No-op stays green.** The gate on the trial merge against master: clean, exit 0.

**4. The live replay against `ed1d215`, and it names exactly the three.** This is the one that matters, and it lands precisely:

```
DELETED-SYMBOLS FAIL: this PR removes 3 symbol(s) from __all__ while the definition still exists at HEAD:
  - taosmd/service.py:a2a_thread_messages
  - taosmd/service.py:a2a_threads
  - taosmd/service.py:fetch_by_ref
```

Same three the card filed, no more and no less. I also classified the full `__all__` delta on that branch, 265 down to 261: the other two removals (`get_human_principal_ids`, `set_human_principal_ids`) are genuine deletions whose definitions are gone too, and they are correctly reported by the definition half rather than double-counted by the export half. Coverage on the known-bad input is complete.

### Two details that are better than they had to be

`find_removed_all_entries` returns `head_symbols[k]`, so the real `def`/`class` kind reaches `_find_adding_commit` instead of the literal `"export"`. That is why blame attribution on the replay is correct rather than degrading to `unknown`. Easy to get wrong, and it was not.

The AST choice is not theoretical: all 18 `__all__` assignments in `taosmd/` are top-level list literals, so `_extract_all_exports` covers 100% of the shapes actually present. I checked for augmented assignment, `__all__.extend(...)`, non-literal values and nested assignment, and there are none, so there is no parser-shaped hole behind this.

### One limitation, and it is mine, not this PR's

The check is scoped to entries whose `def`/`class` lives in the same file, which is exactly what the card's fix section specifies. It therefore does not see a re-export dropped from a package `__init__.py`. Measured:

```
KnowledgeGraph removed from taosmd/__init__.py __all__, import line intact -> gate clean, exit 0
```

Repo-wide that is 109 of 265 export entries, including 72 of the 74 in `taosmd/__init__.py`, all 19 in `loaders/__init__.py` and both in `claims/__init__.py`.

**This is not a defect in this PR.** The card asked for the same-file rule and this delivers it, correctly and with the controls. I had intended to widen the card to cover re-exports and never landed the edit, so the scope gap is on the card, not on the lane. Filing it as a follow-up card with the measurement above rather than holding a PR that met its brief.

Merging.

PR #308, opened 2026-08-17
Revise PR #305: the normalise_handle gate fires on @typing.overload and crashes on non-UTF-8 (its docstring says otherwise)

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #305: the normalise_handle gate fires on @typing.overload and crashes on non-UTF-8 (its docstring says otherwise)

Autonomous build of board card tsk-vcda2i.

Files:
 .github/workflows/normalise-handle-gate.yml     |  27 ++++
 changelog.d/tsk-vcda2i-normalise-handle-gate.md |   3 +
 scripts/normalise_handle_gate.py                |  81 +++++++++++
 tests/test_normalise_handle_gate.py             | 172 ++++++++++++++++++++++++
 4 files changed, 283 insertions(+)

### DIFF
--- a/.github/workflows/normalise-handle-gate.yml
+++ b/.github/workflows/normalise-handle-gate.yml
@@ -1,0 +1,27 @@
+name: Normalise handle gate
+
+# Fails if any module under ``taosmd/`` contains more than one definition
+# of the internal helper ``_normalise_handle``.  Overload stubs decorated
+# with ``@typing.overload`` or ``@overload`` are excluded.  Files that
+# cannot be decoded as UTF-8 are skipped with a warning.
+#
+# See scripts/normalise_handle_gate.py for the implementation.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  normalise-handle-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v7
+      - name: Install Python
+        run: uv python install 3.12
+      - name: Install deps
+        run: uv sync --python 3.12
+      - name: Check for duplicate _normalise_handle definitions
+        run: python scripts/normalise_handle_gate.py
--- a/changelog.d/tsk-vcda2i-normalise-handle-gate.md
+++ b/changelog.d/tsk-vcda2i-normalise-handle-gate.md
@@ -1,0 +1,3 @@
+### Fixed
+- Skip ``@typing.overload`` and ``@overload`` stubs when counting ``_normalise_handle`` definitions, so the gate no longer fails on legitimate typed overload pairs.
+- Catch ``UnicodeDecodeError`` in addition to ``OSError`` when reading files, so the gate no longer crashes on non-UTF-8 input its docstring already claimed it would survive.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+"""Normalise handle gate.
+
+Scans ``taosmd/`` for duplicate definitions of the internal helper
+``_normalise_handle`` and fails if any module contains more than one.
+``@typing.overload`` stubs are excluded: they are not rival copies.
+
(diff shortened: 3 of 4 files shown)


jaylfc:
## APPROVED. Both blockers discharged with red/green on the real scan path, all four controls kept, and the wiring kept intact.

Reviewed against card `tsk-vcda2i`. Trial merge with current master `b8c3d0e`, full suite **1416 passed, 12 skipped, 0 failed**. Conflict-marker check clean, deleted-symbols gate clean. Every probe ran the closed PR #305's gate as the red control in the same command, and both gates were executed from inside the tree so `REPO_ROOT` resolved to the tree under test rather than to the probe's own directory.

### Blocker 1: the `@typing.overload` false positive

Real file under `taosmd/`, one logical helper typed the normal way, two stubs plus the implementation:

```
#305's gate: NORMALISE_HANDLE GATE FAIL: found 3 definitions; at most 1 is allowed   exit 1
#308's gate: normalise-handle-gate: clean                                            exit 0
```

`_has_overload_decorator` handles the `@overload`, `@typing.overload` and called-decorator forms, so it will not miss the spelling a lane happens to use.

### Blocker 2: the non-UTF-8 crash the docstring denied

Actual invalid bytes in a file under `taosmd/`, not a `chmod`:

```
#305's gate: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 36   (unhandled traceback)
#308's gate: normalise-handle-gate: skipping .../taosmd/_probe_utf8.py: unreadable
             normalise-handle-gate: clean                                              exit 0
```

`except (OSError, UnicodeDecodeError)` catches it, and the docstring now describes what the code does. The test uses real `\xff\xfe` bytes rather than a permission error, so it exercises the branch that was actually broken. That was the specific hole in the closed PR's test, and it is closed properly here.

### The controls the card required as permanent tests, all four verified live

```
@typing.overload pair               -> clean, exit 0
try/except ImportError fallback     -> clean, exit 0
two rival defs in one module        -> FAIL, "taosmd/_probe_dup.py: found 2 definitions", exit 1
clean tree                          -> clean, exit 0
```

All four exist in `tests/test_normalise_handle_gate.py` as well. Exit codes above are the real process status, not a pipeline's.

### The live known-bad input still fires

The two lane branches the card cites merge with no conflict and the gate catches what the merge produced:

```
git merge origin/exec/tsk-fjy3o7   -> exit 0, no conflict
taosmd/service.py:40  def _normalise_handle(handle: str) -> str:
taosmd/service.py:50  def _normalise_handle(handle: str, *, mint_strip: bool = False) -> str:
NORMALISE_HANDLE GATE FAIL: taosmd/service.py: found 2 definitions; at most 1 is allowed   exit 1
```

The wiring is kept as the card asked: the workflow runs on every PR, a non-zero exit fails the step, and the tests are collected by `ci.yml`'s `pytest tests/`.

### Two blind spots for the follow-up, neither in scope here

Recording these so they are not rediscovered later. Both are out of scope for this card, which named exactly two blockers, and neither is a reason to hold the PR.

**Cross-module duplication passes.** The rule is per file, so one definition in each of two modules is clean:

```
taosmd/_probe_m1.py: def _normalise_handle  +  taosmd/_probe_m2.py: def _normalise_handle  -> clean, exit 0
```

That matters because half (a) of `tsk-lomnt3` is landing ONE shared `_normalise_handle`, and cross-module is the shape that would violate it. The gate should widen to cross-module at the same time that helper lands, not before, since `_normalise_handle` does not exist anywhere under `taosmd/` today and this gate is currently a forward guard.

**A file that does not parse goes silent.** `except SyntaxError: return 0`, so a merge that leaves conflict markers and duplicate definitions in the same file reports clean:

```
duplicate defs + conflict markers in one file -> clean, exit 0
```

Conflict markers are covered by a separate `git grep` check, so this is defence in depth rather than a hole, but a `SyntaxError` in a file the gate was asked to police is worth a warning on stderr instead of silence.

### Verdict

Both blockers fixed, both proven red against the closed PR's own gate and green here, the controls are permanent tests, the wiring is intact, and the live two-lane replay still fires. Merging.

Per the card's closing note: this gate and the guard half of `tsk-lomnt3` are the same artifact, so `tsk-lomnt3` narrows to landing the one shared helper now that this lands.

PR #313, opened 2026-08-17
verify-merged-assertions.md and the blocks_without_body docstring quote the HTTP handler's error string as the service layer's

jaylfc:
CARD TITLE (intent, not commit subject): verify-merged-assertions.md and the blocks_without_body docstring quote the HTTP handler's error string as the service layer's

Autonomous build of board card tsk-cwzaab.

Files:
 changelog.d/tsk-cwzaab-fix-quoted-error-strings.md | 7 +++++++
 docs/verify-merged-assertions.md                   | 3 ++-
 tests/test_a2a.py                                  | 2 +-
 3 files changed, 10 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/tsk-cwzaab-fix-quoted-error-strings.md
+++ b/changelog.d/tsk-cwzaab-fix-quoted-error-strings.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- Corrected quoted error strings in `docs/verify-merged-assertions.md` and
+  `tests/test_a2a.py` for the `test_http_a2a_blocks_without_body_returns_400`
+  test. The service-layer `ValueError` message is `body must be a
+  non-empty string` (from `taosmd/service.py:436`), not the HTTP handler's
+  `'body' (non-empty string) is required` (from `taosmd/http_server.py:1546`).
--- a/docs/verify-merged-assertions.md
+++ b/docs/verify-merged-assertions.md
@@ -141,7 +141,8 @@ its load-bearing surface is narrower than the other seven.
 Its docstring now states: the `assert status == 400` half is **duplicated** at
 the service layer -- `service.a2a_send` raises `ValueError` for an empty body
 independently. Disabling R8 still returns 400 (from the service-layer ValueError);
-the only thing that changes is the error text. The test asserts
+the only thing that changes is the error message: `body must be a non-empty string`.
+The test asserts
 `"blocks" in body["error"]`, which is the **message** assertion -- the only
 check that is unique to the HTTP-level invariant. The status-code assertion
 cannot fail without the service layer also failing, so a future "tidying" change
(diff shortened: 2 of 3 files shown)


jaylfc:
## MERGE — the corrected string is right, and I measured it rather than read it

Reviewed at `a6d3fc61` on a trial merge with master `8aa62e4` (clean, `ort`).

The whole subject of this card is which layer emits which string, so reading the diff would have
been the wrong check. I reproduced the card's own experiment on the merged tree: disabled the R8
branch in `http_server.py` (`blocks` present => `body` non-empty), ran the test, restored it.

```
R8 disabled:
    assert status == 400                                     PASSES  (service layer still 400)
>   assert "blocks" in body["error"]
E   AssertionError: assert 'blocks' in 'body must be a non-empty string'

R8 restored:  1 passed, 34 deselected
```

`body must be a non-empty string` is exactly what the branch now quotes, in both places. The two
source references in the changelog are accurate line-for-line:

```
taosmd/service.py:436       raise ValueError("body must be a non-empty string")
taosmd/http_server.py:1546  raise _BadRequest("'body' (non-empty string) is required")
```

That second one is the `if not blocks` branch — the handler's own message, which is what the doc
and the docstring were wrongly attributing to the service layer. The diagnosis on the card holds.

**The fix is complete, not just correct.** The stale string survives nowhere:

```
$ grep -rn "'body' (non-empty string) is required" docs/ tests/
(no matches)
```

Both call sites now carry the measured text, and the R8 handler message
(`...is required when 'blocks' is present`) is untouched, which is right — it is the message the
test's `"blocks" in body["error"]` assertion actually depends on.

### Checks

```
deleted-symbols-guard       clean, exit 0
normalise-handle-gate       clean, exit 0
full suite (trial merge)    1435 passed, 12 skipped, exit 0
```

The suite count matching master's baseline exactly is the correct outcome **here**, and the
contrast with #312 (blocked an hour ago for the same number) is worth stating so the rule is not
mis-learned: this PR changes prose only — a doc paragraph, a docstring, and a changelog. No
production code path is touched, so there is no behaviour for a new test to pin. #312 changed
`_format_hit` and reported the same count, which meant something entirely different.

### One process note, not a defect

The card asked you to quote the message *from a measurement* and paste the assertion error you
actually saw. The corrected string is right, so the measurement clearly happened somewhere, but it
is not in the PR — which meant I had to re-run it to confirm rather than check it. Paste the output
next time; it converts a claim a reviewer must verify into evidence a reviewer can read.

Trivial, ignore or fix at will: the changelog file has no trailing newline (so does
`tsk-35fsll-eventqa-exit-code.md`, so this is pre-existing drift rather than something you
introduced), and the reflow left `The test asserts` alone on a short line.

Merging. Card `tsk-cwzaab` closed.

Reviewed by @jaylfc.

PR #314, opened 2026-08-17
deleted-symbols-gate is blind to a re-export dropped from a package __init__.py

jaylfc:
CARD TITLE (intent, not commit subject): deleted-symbols-gate is blind to a re-export dropped from a package __init__.py

Autonomous build of board card tsk-cssuku.

The export-removal clause in find_removed_all_entries only matched
__all__ entries whose def/class lived in the same file.  Re-exports
(names imported via from-import, not defined locally) never matched, so
dropping e.g. KnowledgeGraph from taosmd/__init__.py __all__ while the
import line and the definition survived went unreported.

Add _extract_imports/_get_imports_at_ref to resolve from-import
bindings to candidate definition keys, and extend
find_removed_all_entries to follow those bindings when a same-file
definition is absent.  _resolve_export_key lets the existing def and
class search target the real definition file for an accurate added-by.
A genuine deletion that removes the __all__ entry, the import, and the
definition is not flagged by the export half (it is caught by the
def-deletion half).

Adds TestExtractImports, re-export cases to TestFindRemovedAllEntries,
TestResolveExportKey, TestReexportRemovalIntegration, and
TestReexportRemovalControl.

Files:
 changelog.d/tsk-cssuku-reexport-gate.md |   2 +
 scripts/check_deleted_symbols.py        | 142 ++++++++++++-
 tests/test_deleted_symbols.py           | 360 ++++++++++++++++++++++++++++++++
 3 files changed, 495 insertions(+), 9 deletions(-)

### DIFF
--- a/changelog.d/tsk-cssuku-reexport-gate.md
+++ b/changelog.d/tsk-cssuku-reexport-gate.md
@@ -1,0 +1,2 @@
+### Fixed
+- The `deleted-symbols-gate` CI check now also catches re-exported names silently dropped from a package `__all__` (e.g. an entry in `taosmd/__init__.py` that is backed by `from .x import Name`) when the underlying `def`/`class` still exists at HEAD. Previously only same-file definitions were guarded, leaving the package's top-level public API invisible to the gate.
--- a/scripts/check_deleted_symbols.py
+++ b/scripts/check_deleted_symbols.py
@@ -14,17 +14,22 @@ Two shapes of silent loss are detected:
   2. A name removed from a module __all__ while its top-level def/class still
      exists at HEAD (export removal). This is the shape a mechanical merge
      resolution produces: one side of a conflicted __all__ block is taken
-     wholesale, dropping entries without touching the definitions.
+     wholesale, dropping entries without touching the definitions. Re-exports
+     are included: when the removed entry is a re-exported name (imported, not
+     defined, in the file holding __all__), the gate resolves it through the
+     file's from-import statements and checks whether the target def/class
+     survives anywhere at HEAD.
 
 Algorithm:
   1. Extract all Python def/class symbols at two points: the target branch
      head and HEAD (the merge ref / PR head).
   2. A symbol is "deleted by the PR" if it exists at the target head but not
      at HEAD. The signal is that set.
-  3. Also compare __all__ membership: an entry present at the target head but
-     absent at HEAD is a signal only when the corresponding top-level
-     def/class still exists at HEAD. A genuine deletion removes both the
-     definition and the __all__ entry, so it stays allowed.
+  3. Compare __all__ membership: an entry present at the target head but absent
+     at HEAD is a signal only when the corresponding top-level def/class still
+     exists at HEAD (or, for re-exports, the definition it names survives at
+     HEAD). A genuine deletion removes both the definition and the __all__
+     entry, so it stays allowed.
   4. Fail with an explicit list of the deleted symbols and the commits that
      added them.
   5. A "Removes-Intentionally: <symbol>, ..." trailer in the PR body waives
@@ -32,7 +37,7 @@ Algorithm:
 
 Narrow by design: Python def/class names only (test functions are defs).
 Exports are compared via AST so continuation lines and indentation never
-matter.
+matter. Import resolution for re-exports also uses AST.
 
 Usage:
     python scripts/check_deleted_symbols.py --base origin/master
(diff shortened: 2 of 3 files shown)


jaylfc:
## MERGE — all four controls reproduced independently, on real repository input rather than fixtures

Reviewed at `c150050a` on a trial merge with master `8aa62e4` (clean, `ort`).

Every acceptance item was re-run against the repo itself rather than read out of the test file,
and the RED was proved by running **both gates against the same committed tree** so the two answers
have to disagree before the new one counts.

### RED — the precise case the card filed, and it is the aliased one

`taosmd/__init__.py:3` is `from .knowledge_graph import TemporalKnowledgeGraph as KnowledgeGraph`,
so this is the harder shape: the exported name never appears as a definition anywhere. Dropped
`"KnowledgeGraph"` from `__all__`, left the import line intact, committed:

```
NEW gate (this PR)      exit 1
  DELETED-SYMBOLS FAIL: this PR removes 1 symbol(s) from __all__ while the definition
  still exists at HEAD:
    - taosmd/__init__.py:KnowledgeGraph (added by 93d9ff0 Initial release: taOSmd 0.1.0 ...)

OLD gate (master's), same tree, same command
  deleted-symbols-guard: clean          exit 0
```

That is the blind spot closing, measured. `_resolve_export_key` also does its job: the `added-by`
attribution resolves through the alias to the real definition rather than reporting nothing.

### CONTROL 1 — a legitimate deletion still does not trip the export half

Removed the `__all__` entry, the import line, and the `class TemporalKnowledgeGraph` block itself
(lines 103-727 of `knowledge_graph.py`), committed, re-ran:

```
export half:      silent — no __all__ section in the output at all,
                  and 0 occurrences of "taosmd/__init__.py:KnowledgeGraph"
definition half:  DELETED-SYMBOLS FAIL: ... deletes 21 symbol(s) ...
```

Exactly the split the card asked for: the deletion is reported once, by the half that owns it.

### CONTROL 2 — no regression on the standing known-bad branch

Replayed both gates against `origin/exec/tsk-uyznqh` and diffed the export-half findings:

```
taosmd/config.py:get_human_principal_ids
taosmd/config.py:set_human_principal_ids
taosmd/registry_auth.py:HumanAuthError
taosmd/registry_auth.py:RegistryVerifier.is_human
taosmd/service.py:a2a_thread_messages
taosmd/service.py:a2a_threads
taosmd/service.py:fetch_by_ref

diff old new -> IDENTICAL
```

All three names #306 was carded for are still there, and nothing was added or lost.

### CONTROL 3 and the rest

```
clean merged tree           deleted-symbols-guard: clean, exit 0
normalise-handle-gate       clean, exit 0
full suite (trial merge)    1461 passed, 12 skipped, exit 0
```

Scan scope is consistent — all three `git archive` walks filter on `taosmd/`, so the import
resolution sees exactly the surface the export and definition scans do.

### Notes, none blocking

1. **The PR body does not report a suite count**, which was acceptance item 5. It is the one item I
   had to supply myself. The body is otherwise accurate: the file table matches the real diff
   line-for-line, which is not something I can take for granted lately.
2. **The suite is 1461 passed / 12 skipped, up 26 from master's 1435** — the count moves, which is
   what a PR carrying 360 lines of tests should look like.
3. **Your test module cannot go red against the unfixed gate** — it imports `_extract_imports` at
   module level, so against master it produces a collection error rather than a failing assertion.
   That is the weaker form of red flagged in #306's review, and it is why the evidence above comes
   from running the two gate versions against the same tree instead. The tests themselves are good
   (360 lines, real `tmp_path` git repos driving the gate end to end, including the alias case and
   a waiver case); this is about what they can prove on their own, not their quality.
4. **`_extract_imports` walks `ast.iter_child_nodes`, so it sees top-level from-imports only.**
   Measured on master, this is adequate today and I am not asking you to widen it:
   `taosmd/__init__.py` 38 of 38 top-level, `loaders/__init__.py` 8 of 8, `claims/__init__.py` 1 of
   1 — zero nested. That is the same "does not need to handle those shapes yet" bar the card set
   for `__all__` assignment forms. Worth one docstring line so the next reader does not assume an
   import inside `try:` or `if TYPE_CHECKING:` resolves, because it will not.

### One trap for whoever re-runs this comparison

Copying the old gate to `/tmp` to run it side by side does not work and does not fail honestly
either: `REPO_ROOT` derives from `Path(__file__).parent.parent`, so it resolved to `/` and died in
`git archive` with a `CalledProcessError`. Copy the comparison script *into* the tree under test.
I hit this mid-review and it cost a probe.

Merging. Card `tsk-cssuku` closed.

Reviewed by @jaylfc.

PR #315, opened 2026-08-17
collections: revoke could not remove a grant its own grant call had stored

jaylfc:
Revoking a collection grant silently failed when the grantee id carried surrounding whitespace, because `grant` and `revoke` derived the stored spelling differently.

`grant` wrote `canonical_id.strip()`. `revoke` and `has_grant` matched the raw argument. So passing one padded id to both calls granted query access that could not be taken back: the DELETE matched no rows, the call still returned the collection as if it had succeeded, and the CLI went on printing the grant as live.

Both ends now derive the storage spelling through one helper, `_grantee_key`, so they agree by construction rather than by two call sites happening to spell it the same way.

## The scope of the normalisation is deliberate, and narrow

Whitespace only. `has_grant` decides which collections join `search_agents` in `taosmd.api.search`, so a wider rule (folding case, stripping a leading `@`) would serve content granted to one identity to a different identity that merely looks similar. This is a storage key, never an authorisation decision about two identities that look alike. The docstring says so at the definition, because the next person to touch it will be tempted to widen it.

Non-strings pass through untouched: `revoke` and `has_grant` accept any argument today and simply match nothing, and that stays true.

## Verification

Red first, run against unfixed master with the branch's test file dropped in:

```
test_revoke_matches_the_spelling_grant_stored      FAILED
test_has_grant_matches_the_spelling_grant_stored   FAILED
test_grantee_match_is_not_case_insensitive         passed
2 failed, 20 passed
```

The third one passing on master is the point of it: it is a control against an *over-wide* fix, so it is green on both trees by design and is evidence about the fix's boundary rather than about the defect. Only the first two discriminate.

Full suite on a trial merge with master `f4626e9`: **1464 passed, 12 skipped, exit 0** (baseline 1461; +3, which is this branch's tests).

Conflict markers clean. `deleted-symbols-guard` clean, exit 0. `normalise-handle-gate` clean, exit 0.

## Note

Opened rather than self-merged. This is lead work and it should get the same review the lane PRs get.

## Summary by CodeRabbit

* **Bug Fixes**
  * Grant IDs now handle surrounding whitespace consistently when creating, looking up, and revoking collection grants.
  * Padded IDs can no longer remain active after revocation.
  * Matching remains case-sensitive, so differently cased IDs are treated as distinct.


jaylfc:
Triage of Kilo's single finding on this PR, recorded here rather than carded because the PR is still open and the fix belongs inside it. **The finding is real.** I am not applying it myself — this PR is mine and still wants a review that is not mine to give.

### Confirmed: the annotation contradicts both the docstring and the body

`taosmd/collections.py:136`:

```python
def _grantee_key(canonical_id: str) -> str:
    ...
    Non-strings pass through untouched: ``revoke``/``has_grant`` accept any
    argument today and simply match nothing, and that stays true.
    """
    return canonical_id.strip() if isinstance(canonical_id, str) else canonical_id
```

The signature promises `str` in and `str` out. The docstring promises the opposite for non-strings, and the `isinstance` guard exists solely to deliver it. All three cannot be right. Kilo's line number is exact.

### Scope, measured rather than assumed

No type checker runs in CI — `.github/workflows/` holds `ci.yml` and the three gate workflows, and there is no mypy, pyright or pyre config anywhere in the repo. Annotations are not enforced at runtime either, so **this is a correctness-of-documentation problem, not a live defect**. Nothing misbehaves today. All three call sites (`:393`, `:403`, `:412`) pass `canonical_id` straight through from the public methods.

### One thing Kilo did not say, which I think matters more

**The documented pass-through has no test.** No test in `tests/test_collections_store.py` passes a non-string grantee to `revoke` or `has_grant`. So the sentence "Non-strings pass through untouched … and that stays true" is an unverified behavioural claim sitting next to an annotation that denies it. Whichever way this is resolved, one of the two should become checkable.

I should flag that my first grep for non-string call sites was sloppy and appeared to find three: the pattern allowed `[`, and `col["id"]` matches it. All three "hits" were ordinary string calls. The corrected answer is zero.

### Recommendation

Widen the annotation to match the documented contract and pin the contract with a test — something like `def _grantee_key(canonical_id: object) -> object:`, plus one test asserting `revoke`/`has_grant` with a non-string argument matches nothing and does not raise. Narrowing the other way (dropping the `isinstance` guard and keeping `-> str`) is the alternative, but it would change what `revoke` does with a non-string from "match nothing" to "raise `AttributeError`", and the docstring argues deliberately against that. Your call.

Kilo raised nothing else here; qodo and gitar raised nothing on these lines. For the record, Kilo's five findings on the merged #331 were triaged in the same pass: four real (carded as `tsk-nup2ny`) and one a false positive that inverted what its card actually said.

jaylfc:
Triaging Kilo's finding on `_grantee_key` (comment 5322796166). Both halves were fair, and one of them was my own overstatement.

The finding: the helper is annotated `(canonical_id: str) -> str`, while its docstring and its `isinstance` guard both promise non-strings pass through, and nothing tested that promise.

I measured it on both trees rather than reasoning about it. Same probe file, run against `master` (before the fix) and against this branch (after):

| input | master | this branch |
|-------|--------|-------------|
| `None` | reaches the query, matches nothing | same |
| `7` | reaches the query, matches nothing | same |
| `["agent-a"]` | `sqlite3.ProgrammingError` | same |

So the guard changes nothing for callers. What it does is stop the symmetry fix from *newly* raising `AttributeError` on `.strip()` for the values sqlite binds natively, which would have been a behaviour change this PR does not intend to make. That is now pinned by `test_grantee_key_preserves_the_pre_existing_non_string_behaviour`, and the test discriminates: deleting the guard fails it with exactly that `AttributeError`, and restoring it passes.

The docstring also claimed those callers "accept any argument today and simply match nothing". That sentence was mine and it was wrong: a list has never worked, before or after. Narrowed to what the two probes actually measured, including the half that raises.

Branch was 62 commits behind, so master is merged in. `master` touched `collections.py` once in that range (#347, front matter) and nowhere near the grant paths; all three `_grantee_key` call sites verified present in the merged tree rather than assumed from a clean merge.

On the merged tree: **1666 passed, 12 skipped** against a 1662 baseline, which is the three tests this PR already carried plus the new one. `check_deleted_symbols`, `normalise_handle_gate`, `check_witness_token`, `check_trailing_newline` and `ruff check taosmd/` all clean, exit codes read directly rather than off the end of a pipe.

Still not self-merging this one. It is ready for a look whenever you have a moment.

PR #316, opened 2026-08-17
normalise-handle gate is keyed on one hardcoded name, so duplicate defs of anything else (including shadowed tests) are invisible

jaylfc:
CARD TITLE (intent, not commit subject): normalise-handle gate is keyed on one hardcoded name, so duplicate defs of anything else (including shadowed tests) are invisible

Autonomous build of board card tsk-35ydf3.

The duplicate-definition gate was keyed on one hardcoded name
(_normalise_handle), so duplicate definitions of anything else were
invisible, including shadowed test functions. Generalise it to report any
top-level function or method name defined more than once in the same
module, and extend the scan to cover tests/.

@typing.overload stubs and @property / @<name>.setter / @<name>.getter /
@<name>.deleter accessor pairs remain excluded as legitimate same-name
definitions, reusing the decorator-skip introduced by #308.

Proven red against origin/exec/tsk-v53vta, where three shadowed
test_a2a_banner_mode_* functions in tests/test_http_server.py are
defined twice, and green on current master.

Files:
 .github/workflows/normalise-handle-gate.yml        |  12 +-
 .../tsk-35ydf3-generalise-duplicate-gate.md        |   2 +
 scripts/normalise_handle_gate.py                   | 120 +++++++--
 tests/test_normalise_handle_gate.py                | 291 +++++++++++++++------
 4 files changed, 317 insertions(+), 108 deletions(-)

### DIFF
--- a/.github/workflows/normalise-handle-gate.yml
+++ b/.github/workflows/normalise-handle-gate.yml
@@ -1,9 +1,11 @@
 name: Normalise handle gate
 
-# Fails if any module under ``taosmd/`` contains more than one definition
-# of the internal helper ``_normalise_handle``.  Overload stubs decorated
-# with ``@typing.overload`` or ``@overload`` are excluded.  Files that
-# cannot be decoded as UTF-8 are skipped with a warning.
+# Fails if any top-level function or method name is defined more than once
+# in the same module under ``taosmd/`` or ``tests/`` (e.g. shadowed test
+# functions).  ``@typing.overload`` stubs and ``@property`` getter /
+# ``@<name>.setter`` / ``@<name>.getter`` / ``@<name>.deleter`` accessor
+# pairs are excluded.  Files that cannot be decoded as UTF-8 are skipped
+# with a warning.
 #
 # See scripts/normalise_handle_gate.py for the implementation.
 
@@ -23,5 +25,5 @@ jobs:
         run: uv python install 3.12
       - name: Install deps
         run: uv sync --python 3.12
-      - name: Check for duplicate _normalise_handle definitions
+      - name: Check for duplicate definitions in taosmd/ and tests/
         run: python scripts/normalise_handle_gate.py
--- a/changelog.d/tsk-35ydf3-generalise-duplicate-gate.md
+++ b/changelog.d/tsk-35ydf3-generalise-duplicate-gate.md
@@ -1,0 +1,2 @@
+### Fixed
+- Generalised the `normalise-handle-gate` CI check from the single name `_normalise_handle` to any top-level function or method name defined more than once in the same module, so shadowed test functions (and any other accidental redefinition) are reported. The scan now also covers `tests/`. `@typing.overload` stubs, `@property` getters, and `@<name>.setter` / `@<name>.getter` / `@<name>.deleter` accessor pairs are still excluded as legitimate same-name definitions.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -1,20 +1,35 @@
 #!/usr/bin/env python3
-"""Normalise handle gate.
+"""Duplicate-definition gate.
 
-Scans ``taosmd/`` for duplicate definitions of the internal helper
-``_normalise_handle`` and fails if any module contains more than one.
-``@typing.overload`` stubs are excluded: they are not rival copies.
+Scans ``taosmd/`` and ``tests/`` for any top-level function or method name
+defined more than once in the same module (scope) and fails if rival
+definitions are found.  The original gate only looked for the internal helper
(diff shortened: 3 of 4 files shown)


jaylfc:
## Review: MERGE

Verified independently rather than read. Every claim below was measured on this branch at `43aeabe`, and the load-bearing ones were measured by running master's gate and this gate over the **same** inputs and requiring them to disagree.

### The headline claim reproduces on real repo input

Not a fixture. The gate from this branch, copied into the preserved known-bad branch `origin/exec/tsk-v53vta`, against master's gate on that same tree:

```
NEW gate: exit 1
  tests/test_http_server.py: module 'test_a2a_banner_mode_no_registry_url'           lines [624, 1162]
  tests/test_http_server.py: module 'test_a2a_banner_mode_with_registry_and_enforce' lines [648, 1188]
  tests/test_http_server.py: module 'test_a2a_banner_mode_with_registry_no_enforce'  lines [670, 1210]
OLD gate: exit 0  normalise-handle-gate: clean
```

All three names and all six line numbers match the card exactly, and the old gate is blind on the identical tree. That is the discriminating shape: same perturbation, both versions, forced disagreement.

### The controls that had to hold, and did

Eleven synthetic repo roots, each built twice, with each gate version copied **into** the tree it scans (`REPO_ROOT` derives from `__file__`, so a gate run from outside scans the wrong directory and reports clean for it).

| case | old | new | |
|---|---|---|---|
| two module-level `_normalise_handle` | 1 | 1 | the original case still caught |
| shadowed test functions | 0 | **1** | the new coverage |
| duplicate method in one class | 0 | **1** | the new coverage |
| `@typing.overload` triple | 0 | 0 | the false positive that blocked #305 |
| `@property` / `@x.setter` pair | 0 | 0 | accessor pairs stay silent |
| non-UTF-8 file | 0 | 0 | no crash; `except OSError` alone cannot catch this |
| clean tree | 0 | 0 | negative control |

The first row is the one I most wanted to see. Generalising the alphabet did not cost the name the gate was built for, and there is a test pinning it (`test_rival_defs_in_taosmd_tree_fail`), so it cannot silently regress later.

On the real merged tree, across 183 files of `taosmd/` and `tests/`: **clean, exit 0**. A guard that fires on correct code gets switched off within a week, and this one does not fire.

### Suite, read against what the diff touched

**1473 passed, 12 skipped** on a trial merge with `568bd33`. Baseline is 1461, and the one touched test file goes from 15 tests to 27, so +12 is exactly accounted for. Worth stating because a production change reporting the baseline count would have meant no test was added.

One thing that looks alarming and is not: `git diff --numstat origin/master 43aeabe` reports `0 360` against `tests/test_deleted_symbols.py`. That is base drift, this branch is based on `0bf8c9f` and master has since merged #314. Diffed against its actual base the branch touches four files, and the merged tree still carries that file at 940 lines.

Conflict markers: none. Deleted-symbols gate from inside the merged tree: clean.

### The tests are real

I checked for the failure mode that has cost this repo three PRs today, tests that recompute the production logic inside the test body and assert against their own arithmetic. These do not. They write real files, call the real functions, and assert on real output, and the integration tests monkeypatch `REPO_ROOT`, which is the correct seam. `test_non_utf8_file_skipped` writes actual invalid bytes rather than relying on the `chmod(0o000)` case, so it exercises `UnicodeDecodeError` and not its `OSError` sibling.

### One scope change that is not a blocker but is going on the record

The implementation swaps `ast.walk(tree)` for a walk of `tree.body` plus class bodies. That is a deliberate narrowing, two of its effects are fixes and two are lost coverage:

```
module-level def + def inside `if`/`try`   old 1 | new 0   genuine same-scope shadow, now invisible
two same-name closures in ONE parent       old 1 | new 0   genuine shadow, now invisible
try/except ImportError fallback            old 1 | new 0   was a FALSE POSITIVE, correctly dropped
module-level def + class method same name  old 1 | new 0   different namespaces, correctly dropped
```

Kilo flagged the first of these at line 109 and it is correct. I had measured it before reading the comment, which is why it is stated here as confirmed rather than repeated.

I sized it rather than guessing: the real tree has 5 definitions inside module-level control flow and 190 nested directly inside another function, and under a maximal-scope checker **zero** of them are currently duplicated. So this is a forward gap, not a live regression, and on today's tree this gate is strictly better than the one it replaces. The nested-function exclusion is documented at the definition with a reason; the control-flow one is not, and the stated reason ("same-name closures in different parents are legal") does not cover two closures in the same parent. Both go on a follow-up card rather than against this build, because the card asked for top-level functions and methods and that is what was delivered.

Kilo's other note, that `test_clean_output_when_passing` does not assert a return code, is fair but redundant: `test_clean_when_no_duplicates` asserts `rc == 0` on the same input.

Merging.

PR #319, opened 2026-08-17
Revise PR #310: the banner tests assert against a copy of the logic they re-implement, and three are shadowed duplicates

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #310: the banner tests assert against a copy of the logic they re-implement, and three are shadowed duplicates

Autonomous build of board card tsk-6xrv7d.

Reuse the PR #310 production change as-is in taosmd/http_server.py: serve()
now prints an "A2A registry auth mode: OFF|ENFORCE|WARN (verify-and-warn)"
line, and strips "(no auth)" from the where line only when a registry_url
is configured with enforcement on.

The previous banner tests re-implemented serve()'s mode/where logic inside
the test body and asserted on their own copy, so all six passed against
unfixed master. Three were also shadowed duplicates (defined twice with the
first binding silently discarded).

Rewrite the banner tests as subprocess probes that capture serve()'s real
stdout, guarded by a PROVENANCE marker confirming the exercised module. Six
cases cover the three mode states (OFF, ENFORCE, WARN) and the three LAN-bind
where-line cases. Verified RED against unfixed http_server.py (6 failed: the
"A2A registry auth mode:" line was absent), then GREEN after the banner
change (6 passed). Full suite: 1441 passed, 12 skipped, 0 failed.

Files:
 changelog.d/tsk-6xrv7d-a2a-banner-tests.md |   7 ++
 taosmd/http_server.py                      |  11 ++
 tests/test_http_server.py                  | 156 +++++++++++++++++++++++++++++
 3 files changed, 174 insertions(+)

### DIFF
--- a/changelog.d/tsk-6xrv7d-a2a-banner-tests.md
+++ b/changelog.d/tsk-6xrv7d-a2a-banner-tests.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- The `taosmd serve` startup banner now prints an `A2A registry auth mode:` line that
+  reads `OFF`, `ENFORCE`, or `WARN (verify-and-warn)` from the same config the
+  enforcement path branches on, so the banner no longer claims `ENFORCE` when no
+  registry_url is configured and no longer advertises `no auth` on the `where` line
+  once enforcement is active.
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -2237,8 +2237,19 @@ def serve(host: str = DEFAULT_HOST, port: int = DEFAULT_PORT, data_dir=None) -> 
     """
     httpd = make_server(host, port, data_dir)
     bound_host, bound_port = httpd.server_address[:2]
+    _enforce = _config.get_a2a_auth_enforce(data_dir)
+    _registry_url = _config.get_registry_url(data_dir)
+    if _registry_url is None:
+        mode = "OFF (no registry_url: senders are self-claimed)"
+    elif _enforce:
+        mode = "ENFORCE"
+    else:
+        mode = "WARN (verify-and-warn)"
     where = "localhost only" if bound_host in {"127.0.0.1", "::1"} else "LAN-reachable (no auth)"
+    if _registry_url is not None and _enforce:
+        where = where.replace(" (no auth)", "")
     print(f"taosmd HTTP API listening on http://{bound_host}:{bound_port} ({where})")
+    print(f"A2A registry auth mode: {mode}")
     print(f"Inspection UI (read-only): http://{bound_host}:{bound_port}/")
     print("Endpoints: GET /health, GET /version, POST /ingest, POST /ingest/batch, GET|POST /search, "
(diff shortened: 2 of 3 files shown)


jaylfc:
## Review: MERGE

This fixes what #310 was blocked on, and it fixes it at the root rather than by patching the symptom. Every claim below was measured on this branch at `2fdde51`.

### The carry-forward is intact, checked as a deliverable in its own right

#310's production diff was correct and verified at the time, so the card told this revision to reuse it unchanged. That instruction is invisible in this PR's own diff, so it has to be checked against the branch being revised:

```
git diff origin/exec/tsk-v53vta 2fdde51 -- taosmd/   ->  (empty)
```

Byte-identical. The mode banner and the `(no auth)` strip are carried forward exactly, nothing was silently dropped or re-derived.

### The red control, reproduced independently

This is the whole point of the card, so I did not take the PR body's word for it. I took the merged tree, reverted **only** `taosmd/http_server.py` to master's version, left the new tests in place, and confirmed the feature was genuinely absent first:

```
grep -c 'A2A registry auth mode' taosmd/http_server.py   ->  0

6 tests against unfixed production   ->  6 failed
6 tests against the fixed tree       ->  6 passed
```

That is the exact inverse of #310, where the same six cases passed against unfixed master and therefore measured nothing. These tests have real coverage.

The reason they work is that they no longer re-implement anything. `_serve_banner` runs `serve()` in a real subprocess and reads its actual stdout, and the assertions compare against literal expected strings:

```python
assert _mode_line(lines) == (
    "A2A registry auth mode: OFF (no registry_url: senders are self-claimed)"
)
```

If production changes that string, the test fails. That is what was missing before.

`_require_provenance` is a good addition and worth naming: each subprocess prints `PROVENANCE <module __file__>` and the test asserts it equals `http_server.__file__`. That closes the standing trap where standing in a worktree does not bind the interpreter to it, which has produced false passes in this repo more than once.

### The shadowed duplicates are gone, and the gate confirms it independently

#310 defined three test functions twice, so three of nine were silently discarded. The six tests here are renamed and unique. Rather than eyeball that, I ran the duplicate-definition gate that merged an hour ago in #316, which exists precisely to catch this and fires on #310's branch naming all three:

```
scripts/normalise_handle_gate.py on this merged tree  ->  clean, exit 0
```

### The rest

Suite on a trial merge with `fc44ec6`: **1479 passed, 12 skipped**. Baseline is 1473 and this adds 6 tests, so the count is exactly accounted for. The PR body reports 1441, which was correct against its own base `0bf8c9f` (1435 + 6) and is stale only because master has moved twice since.

No conflict markers. Deleted-symbols gate clean from inside the merged tree.

One non-blocking note for whoever touches this next: `_serve_banner` stops reading when it sees a line starting with `Admin (admin token required)`. If that banner line is ever renamed, these tests will not fail fast, they will each read until the 15 second timeout. Worth a sentinel that does not depend on unrelated banner text, but not worth holding this.

Merging.

PR #322, opened 2026-08-17
Doc gate: assert a cited test WITNESS token resolves, not just the test path

jaylfc:
CARD TITLE (intent, not commit subject): Doc gate: assert a cited test WITNESS token resolves, not just the test path

Autonomous build of board card tsk-s2keyh.

A source comment justified by "See the RED witness in test_resume_arm_time.py"
stayed green after a revert deleted that witness from the test, because the test
file itself survived and the suite still passed. A path-only check cannot catch
this, since it resolves the file, not the token the comment points at.

scripts/check_witness_token.py introduces an opt-in `# WITNESS: <test>::<token>`
marker. The gate resolves `<token>` as a case-sensitive substring inside `<test>`,
so a deleted witness flips the gate to non-zero even while the test file remains
present and importable. Bare test-filename mentions in ordinary prose (no marker)
are ignored, so explanatory prose about a removal is never flagged.

Wired as a Layer A CI check (.github/workflows/witness-token-gate.yml), matching the
normalise-handle and deleted-symbols gates. Scope is taosmd/**/*.py; only the
explicit WITNESS: marker is an assertion.

Four evidence blocks, real runs against throwaway fixtures (WITNESS_GATE_ROOT set
per block so the gate targets the fixture, not the checked-out tree):

RED, the motivating case -- a WITNESS marker whose token is absent from the test:

```
WITNESS GATE FAIL:
  taosmd/svc.py:2: WITNESS tests/test_foo.py::MAX_FIRE_TO_DELETE -> witness token not found in tests/test_foo.py
```
(exit 1, names the source file, the test, and the token)

RED, non-vacuity -- start from the GREEN repo, then delete ONLY the token. The
test file stays present and importable, still defining `def test_arm_time`; a
path-only check could not see this, but the witness-token gate flips to non-zero:

```
[baseline] witness-gate: clean
```
(exit 0)

```
[after deleting only MAX_FIRE_TO_DELETE, file still present] WITNESS GATE FAIL:
  taosmd/svc.py:2: WITNESS tests/test_foo.py::MAX_FIRE_TO_DELETE -> witness token not found in tests/test_foo.py
```
(exit 1)

GREEN, true positive -- a declared witness whose token IS present:

```
witness-gate: clean
```
(exit 0)

GREEN, false positive -- a source file that mentions a test filename in ordinary
prose with no WITNESS marker (this is the post-fix explanatory line from the bug):

```
witness-gate: clean
```
(exit 0)

Full suite: uv run --extra dev pytest tests/test_witness_gate.py -q -> 17 passed.

Card tsk-s2keyh.

Files:
 .github/workflows/witness-token-gate.yml |  33 ++++
 changelog.d/tsk-s2keyh-witness-gate.md   |   8 +
 scripts/check_witness_token.py           | 177 ++++++++++++++++++++
 tests/test_witness_gate.py               | 278 +++++++++++++++++++++++++++++++
 4 files changed, 496 insertions(+)

### DIFF
--- a/.github/workflows/witness-token-gate.yml
+++ b/.github/workflows/witness-token-gate.yml
@@ -1,0 +1,33 @@
+name: Witness token gate
+
+# Verifies that every `# WITNESS: <test>::<token>` marker declared in a
+# `taosmd/` source file resolves to a real token inside the named test file:
+# the test file must exist and `<token>` must appear in it. This catches the
+# regression where a source comment cites a test as the justification for a
+# constant, but a revert deletes the cited witness from the test while the
+# test file itself survives and the suite stays green.
+#
+# Only the explicit `WITNESS:` marker is an assertion; bare mentions of test
+# filenames in ordinary prose are ignored, so explanatory prose about a
+# removal is not flagged.
+#
+# See scripts/check_witness_token.py for the implementation.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  witness-token-gate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v7
+      - name: Install Python
+        run: uv python install 3.12
+      - name: Install deps
+        run: uv sync --python 3.12
+      - name: Check that declared witness tokens resolve
+        run: python scripts/check_witness_token.py
--- a/changelog.d/tsk-s2keyh-witness-gate.md
+++ b/changelog.d/tsk-s2keyh-witness-gate.md
@@ -1,0 +1,8 @@
+### Added
+- A `witness-token-gate` CI check (`scripts/check_witness_token.py`) that
+  verifies any `# WITNESS: <test>::<token>` marker in a `taosmd/` source file
+  resolves to a real `<token>` inside the named test file. A source comment that
+  cites a test as the justification for a constant can no longer stay green after
+  the witness it points at is deleted from that test, because the gate resolves the
+  token, not just the path. Only the explicit `WITNESS:` marker is an assertion: a
+  bare test-filename mention in ordinary prose is ignored.
--- a/scripts/check_witness_token.py
+++ b/scripts/check_witness_token.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+"""Witness token gate.
+
+Verifies that every explicit ``# WITNESS: <test>::<token>`` marker in a source
+file resolves to something real INSIDE the named test: the referenced test file
+must exist and ``<token>`` must appear in it as a case-sensitive substring (grep
+semantics).
+
+Why a token, not a path: a source comment that cites a test as the justification
+for a constant is only as good as what the test still contains. A path-only check
+passes green while the cited witness tokens are deleted from the test, because the
+test file itself survives. This gate cannot be fooled by that -- it resolves the
+TOKEN, not the path. The gate is non-vacuous in both directions: it fails when a
+declared token is removed (file stays importable), and it does NOT fire on a bare
+test-filename mention in ordinary prose.
+
+Marker contract (the only thing that triggers the check):
+
+    # WITNESS: tests/test_foo.py::some_grepable_token
+
+- The ``WITNESS:`` marker is all-caps and opt-in. A bare mention of a test
+  filename in prose -- whether describing a removal or stating a fact --
+  declares nothing and is invisible to the gate. Only lines containing the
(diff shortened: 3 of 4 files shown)


jaylfc:
## Review: MERGE

The card's non-negotiable property is that the assertion names a token INSIDE the file and that undeclared prose mentions are invisible. Both hold. I rebuilt all four acceptance blocks on my own fixtures rather than reading yours, and added the control the card implies but does not require: running the naive path-resolution gate and this one over the SAME inputs and requiring them to disagree.

### The controls, on my fixtures

Block 3 then block 2, the non-vacuity direction. The test file stays present and importable, only the token goes:

```
token present                                    witness-gate: clean                  exit 0

[delete ONLY the token from tests/test_foo.py]
ls -la tests/test_foo.py                         -rw-rw-r-- 37 bytes, present
import check                                     IMPORTABLE, defines test_arm_time: True
grep -c MAX_FIRE_TO_DELETE tests/test_foo.py     0

WITNESS GATE FAIL:
  taosmd/svc.py:1: WITNESS tests/test_foo.py::MAX_FIRE_TO_DELETE -> witness token not found in tests/test_foo.py
                                                                                      exit 1
```

### The disagreement control: naive path gate vs this gate, same two inputs

I wrote the path-resolution version the card says does not work, and ran both over the same fixtures. It is wrong in both directions at once, exactly as the card predicted, and this gate is right in both:

```
                                        naive path gate    witness token gate
token deleted, test file present        clean    exit 0    FAIL     exit 1
prose mention, no marker                FAIL     exit 1    clean    exit 0
```

That second row is the real post-fix line from the bug, verbatim. A clean result only becomes evidence when a differently-built instrument disagrees with it on the same input, and these do.

### Adversarial edges, all sensible

```
malformed marker, no ::           FAIL   "malformed WITNESS marker: expected <test_file>::<token>"
test file missing entirely        FAIL   "test file not found: tests/test_nope.py"
token containing :: (node id)     clean  first-:: split works as documented
non-UTF-8 file in scope           skipped with a stderr note, gate still exits 0
```

That fourth one is worth calling out as a positive: `except (OSError, UnicodeDecodeError)` catches both. A sibling gate shipped with `except OSError` alone and died on the decode.

### Suite and gates, run on the trial merge and AFTER merging, not before

```
full suite                1496 passed, 12 skipped in 154.72s   exit 0
conflict markers          none
check_deleted_symbols.py  deleted-symbols-guard: clean         exit 0
normalise_handle_gate.py  normalise-handle-gate: clean         exit 0
check_witness_token.py    witness-gate: clean                  exit 0
```

Master baseline is 1479/12, so this is +17, and the diff adds exactly one test file containing 17 tests. The count matches what the diff touched.

One note on the body: `Full suite: uv run --extra dev pytest tests/test_witness_gate.py -q -> 17 passed` is a single-file run labelled as the full suite. The number is true for that file, but no full-suite count was ever reported. The count above is mine.

### Excluding tests/ is correct, and I verified the reason rather than accepting it

```
tests/test_witness_gate.py:41    f"# WITNESS: {TEST_TEST}::{TOKEN}\n"
```

Put that exact shape in a `taosmd/` file and the gate fires with `test file not found: {TEST_TEST}`. Nine such marker lines live under `tests/`. The stated reason is real, not a rationalisation. Keep it.

### Two things carded rather than held, on card `tsk-7oak6l`

Filed and read-back verified before this comment was written.

**The scope excludes `scripts/`, which is where the motivating constant lives.** Marker census on the merged tree:

```
taosmd/**/*.py    0
scripts/**/*.py   2   (both in check_witness_token.py's own docstring)
tests/**/*.py     9
```

`RETRY_LEAD_SECONDS = 60` with the comment citing the RED witness sits in the 721-line canonical `resume_arm_time.py`, not under `taosmd/`, and card `tsk-y62mpe` requires that file to move into `scripts/`. Carded and not held because live exposure is zero: there are no WITNESS markers in `taosmd/` at all. Same standard I applied to #316's narrowing. Widening is also not a one-liner, since adding `scripts/` makes the gate fail on its own docstring example at line 19; the card says so.

**One factual error in the justification.** Lines 36-37 say "The sibling normalise-handle gate uses the same `taosmd/` scope." It does not:

```
scripts/normalise_handle_gate.py:25   TARGET_PATTERNS = ("taosmd/**/*.py", "tests/**/*.py")
```

`check_deleted_symbols.py` is the `taosmd/`-only one; cite that instead. The scope choice stands on its own merits, but a wrong citation in a gate's docstring is the same class of defect this gate exists to catch, which is why it is on the card rather than left in the file unremarked.

### Card constraints

```
em dashes in the diff        0
AI attribution in the commit none
CI workflow                  matches normalise-handle-gate.yml step for step
```

Merging.

PR #324, opened 2026-08-17
Revise PR #282: an undocumented THIRD failure class — grant failures now hard-403 in warn mode, which is the default

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #282: an undocumented THIRD failure class — grant failures now hard-403 in warn mode, which is the default

Autonomous build of board card tsk-brirhl.

Revise PR #282 (closed) from master. Split A2A auth failure classes in
_handle_a2a_send: presented-credential failures (bad signature, issuer
mismatch, revoked id, sub != from) now return a hard 403 in warn mode,
which is the default, instead of being warn-and-accepted. Grant failures
remain warn-and-accept in warn mode and 403 in enforce mode, since a
valid token with no grant is an authorization gap, not a credential
failure. HumanAuthError stays always-403.

RegistryVerifier gains a configurable staleness_bound (default 6x
refresh_interval) that fails closed when the cached revocation set is too
old and the refresh feed is unreachable. The stale predicate now includes
cache age before the normal refresh interval, fixing the case where a
bound shorter than refresh_interval was bypassed. Thread the bound through
config (TAOSMD_REGISTRY_STALENESS_BOUND env var or
registry_staleness_bound config key) into verifier_from_url.

Adds 6 tests, RED-first confirmed against master: 2 for the failure-class
split (invalid token and mismatched sub reject at 403 in warn mode),
5 for staleness_bound fail-closed semantics, plus a regression guard that
grant failures remain warn-and-accept in warn mode.

Files:
 changelog.d/tsk-brirhl-registry-auth-staleness.md |  18 ++++
 taosmd/config.py                                  |  63 ++++++++++++
 taosmd/http_server.py                             |  36 +++++--
 taosmd/registry_auth.py                           |  27 ++++-
 tests/test_http_server_trust_enforcement.py       |  43 +++++++-
 tests/test_registry_auth.py                       | 117 ++++++++++++++++++++++
 6 files changed, 285 insertions(+), 19 deletions(-)

## Summary by CodeRabbit

* **Security**
  * Invalid or mismatched authentication credentials are now rejected with HTTP 403 responses.
  * Missing credentials and valid credentials without grants retain existing warn-or-enforce behavior.
  * Registry authorization data now fails closed when it becomes too stale after refresh failures.

* **Configuration**
  * Added an optional setting to control the registry authorization staleness limit, with a secure default.

### DIFF
--- a/changelog.d/tsk-brirhl-registry-auth-staleness.md
+++ b/changelog.d/tsk-brirhl-registry-auth-staleness.md
@@ -1,0 +1,18 @@
+### Added
+
+- `RegistryVerifier` (and `verifier_from_url`) now accepts a configurable
+  `staleness_bound` (seconds); the revocation cache fails closed once it exceeds
+  the bound on a refresh failure, instead of tolerating stale data indefinitely.
+  The bound is configurable via the `TAOSMD_REGISTRY_STALENESS_BOUND` env var or
+  the `registry_staleness_bound` config key, defaulting to `6 * refresh_interval`.
+
+### Fixed
+
+- A2A bus auth now distinguishes missing-credentials from presented-credential
+  failures: a missing Bearer token is still warn-and-accept in verify-and-warn
+  mode (the default), but any token that is presented and fails verification
+  (bad signature, issuer mismatch, revoked id, sub != from) returns a hard 403
+  regardless of the `a2a_auth_enforce` flag. Grant failures -- a valid token
+  whose `sub` matches `from` but holds no active grant -- remain warn-and-accept
+  in warn mode and 403 in enforce mode, preserving the documented migration
+  contract.
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -52,6 +52,9 @@ _SERVE_DASHBOARD_KEY = "serve_dashboard"
 _GENERATOR_PROFILE_KEY = "generator_profile"
 # Whether A2A registry auth runs in enforce mode (True) or verify-and-warn mode (False).
 _A2A_AUTH_ENFORCE_KEY = "a2a_auth_enforce"
+# Maximum age (seconds) of a cached revocation set before refresh failures fail
+# closed. Defaults to 6 * refresh_interval inside RegistryVerifier.
+_REGISTRY_STALENESS_BOUND_KEY = "registry_staleness_bound"
 # Canonical IDs of human principals (controller sessions). These IDs skip the
 # registry revocation check and the grants check; a sub/from mismatch on a
 # human token is always rejected, even in verify-and-warn mode.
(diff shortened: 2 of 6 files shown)


jaylfc:
## Review: BLOCKED

One blocker, and it is narrow. Most of this build is right and should come back unchanged, so the credit is first and specific.

### What is right, measured rather than read

**The card's blocker is genuinely fixed.** Grant failures are warn-and-accept in warn mode again and 403 in enforce mode. The taxonomy comment at `http_server.py:1554` is clear and its reasoning is sound.

**RED FIRST is real and I reproduced it independently**, by checking out the branch's base, bringing across ONLY the two test files, and leaving production unfixed:

```
7 failed in 2.33s
  test_verifier_staleness_bound_defaults_to_six_refresh_intervals
  test_verifier_staleness_bound_fails_closed_after_bound
  test_verifier_staleness_bound_age_equal_to_bound_remains_valid
  test_verifier_staleness_bound_triggers_refresh_before_refresh_interval
  test_verifier_from_url_accepts_staleness_bound
  test_warn_invalid_token_rejected
  test_warn_mismatched_sub_rejected
```

Master's own log during that run is the defect verbatim:

```
WARNING taosmd.http_server:1608 a2a verify-and-warn: accepting unverified post from
'agent-1': token sub 'agent-OTHER' does not match from 'agent-1'
```

Worth separating honestly: the five staleness failures are API-absence (the kwarg does not exist yet), which is a weaker red than the two http failures, which are behavioural. Both halves are still red before and green after.

**An inverted test was inverted honestly.** `test_warn_invalid_token_accepted` became `test_warn_invalid_token_rejected` with the expectation flipped 200 to 403, rather than quietly deleted, and `test_warn_valid_token_no_grant_accepted` was strengthened to assert the message IS persisted. Warn-and-accept is now proved by effect, not just by status code. Both new rejection tests also assert the request never reached the service layer.

**The config knob is properly built**: env then config key, rejects bools and non-positives, has a `clear=True` path, exported in `__all__`. `staleness_bound if staleness_bound is not None else 6 * refresh_interval` correctly preserves an explicit 0 rather than falling through, which is the usual bug in that line.

```
full suite on a trial merge   1502 passed, 12 skipped   (baseline 1496, +6)
net test delta from the diff  +6                        (7 added, 1 replaced)
conflict markers              none
check_deleted_symbols.py      clean
normalise_handle_gate.py      clean
check_witness_token.py        clean
git diff origin/exec/tsk-fo7lxp <sha>   non-empty, this is a real revision
```

The count matches what the diff touched.

### The blocker: a third failure class exists, and it lands in the impersonation class

The comment block enumerates two classes: missing token (tolerated in warn mode) and presented-credential failure (always 403, described as "an active impersonation attempt"). Registry-unavailability is neither, and `except registry_auth.AuthError` catches it into class 2.

`RegistryVerifier.authorize` calls `_get_revoked()`, which raises `AuthError` when the feed cannot be fetched: `revocation feed unavailable` on a cold start with no cache, and the new `revocation set stale (...)` past the bound. Both reach the class-2 arm.

Measured end to end over HTTP in warn mode, with a token valid in every respect (good signature, correct issuer, `sub == from`, not revoked, and the agent HOLDS an `a2a_send` grant). The only thing perturbed is the revocation feed:

```
                                     master a59278d7   PR #324
CONTROL: invalid token                   200            403     <- instrument discriminates
valid token, feed HEALTHY                200            200
valid token, feed UNREACHABLE            200            403     <- the finding
```

```
{"error": "registry auth: revocation feed unavailable, refusing to authorise: registry unreachable"}
log: a2a auth: rejecting presented-credential failure from 'agent-allowed' in warn mode
```

A registry outage is reported to the operator as impersonation.

Why this blocks rather than gets carded, which is the call I actually had to make:

1. **The cold-start case is not the new feature.** `revocation feed unavailable` is pre-existing and was warn-and-accept in warn mode on master. This diff flips it to reject. Not requested by the card, not in the changelog (`grep -c unavailable changelog.d/` is 0), not in the taxonomy comment, no test in either direction.
2. **It fires in the default mode**, whose documented contract is to observe without breaking traffic. A registry outage past the bound (default 1800s) refuses every agent holding a good token.
3. **The card this revises is titled "an undocumented THIRD failure class".** A revision that introduces an undocumented third failure class at a different point in the same function leaves the lineage where it started.

I weighed the other way, because carding is what I did for the last two scope gaps and live exposure here is zero (`get_registry_url()` is `None` on this box and the whole block is gated on it). Those were advisory gates with no runtime effect. This is request handling on the auth path in the default mode, where the cost of a false rejection is the bus going down during an outage. Fail-closed staleness itself is correct and stays; the only question is which disposition it takes in warn mode, and that it be named.

The fix is cheap: give registry-unavailability its own `AuthError` subclass, catch it before the class-2 arm (ordering matters, a subclass caught after its parent never runs), pick a warn-mode disposition and write the reason at the branch, and extend the comment to three classes.

### A correction to my own probe, because it had this backwards at first

My first run of the HTTP probe reported 200 for both the healthy and unreachable arms, which read as a clean bill of health. It was measuring the wrong tree: running a probe by ABSOLUTE PATH puts the script's own directory on `sys.path[0]`, not the cwd, so `import taosmd` resolved through the editable install to the main checkout. The positive control is what caught it, an invalid token returning 200 where the fixed tree must return 403. Copying the probe into each tree and printing `taosmd.__file__` produced the table above. Recorded on the card as a named trap.

---

**Closing under the close-on-block policy.** Branch `exec/tsk-brirhl` is preserved. Card `tsk-brirhl` is closed in the same action, and the revision card is **`tsk-g56qcg`**, filed and read-back verified before this comment was written. It carries the Credit section forward as a deliverable so none of the good work is rebuilt.

PR #331, opened 2026-08-18
Document that human_principal_ids is a revocation exemption list, including the fail-closed bypass

jaylfc:
CARD TITLE (intent, not commit subject): Document that human_principal_ids is a revocation exemption list, including the fail-closed bypass

Autonomous build of board card tsk-wa2bwk.

Files:
 changelog.d/tsk-wa2bwk-human-principal-docs.md |  2 ++
 docs/specs/a2a-bus-auth-transition.md          | 16 +++++++++++++++-
 taosmd/config.py                               | 14 ++++++++++++--
 taosmd/registry_auth.py                        |  9 +++++++--
 4 files changed, 36 insertions(+), 5 deletions(-)

## Summary by CodeRabbit

* **Documentation**
  * Clarified that configured human principal IDs bypass revocation-feed checks and fail-closed behavior.
  * Documented that these IDs are not validated as human identities, so agent identifiers may unintentionally receive the same exemption.
  * Clarified that human-token subject mismatches remain rejected.
  * Documented that environment-variable settings take precedence over configuration-file values.
  * Added risk scenarios and control cases covering revoked IDs and unavailable revocation feeds.

### DIFF
--- a/changelog.d/tsk-wa2bwk-human-principal-docs.md
+++ b/changelog.d/tsk-wa2bwk-human-principal-docs.md
@@ -1,0 +1,2 @@
+### Added
+- Documented that `human_principal_ids` exempts listed canonical_ids from the registry revocation check AND from the fail-closed refusal that fires when the revocation feed has never loaded. The id is not validated as belonging to a human, so naming an agent canonical_id there silently disables that agent's revocation.
--- a/docs/specs/a2a-bus-auth-transition.md
+++ b/docs/specs/a2a-bus-auth-transition.md
@@ -15,7 +15,21 @@ See Target state item 3 and open question 1.
 "unresolved" invites the opposite reading.** Nothing is broken today. With no
 `human_principal_ids` configured, #235's guard condition is always true, so revocation
 currently applies to **everyone**. The hole opens the first time the feature is used for its
-purpose, not before.
+purpose, not before. Once an id is listed, it is also exempt from the fail-closed refusal
+that fires when the revocation feed has never loaded, and it is not validated as belonging
+to a human: naming an agent's canonical_id there silently disables that agent's revocation.
+Measured on exec/tsk-legqtr at 8ba25e2:
+
+```
+revoked id IS on the feed:
+  human_principal_ids=set()            rejected   <- control, the check can reject
+  human_principal_ids={that id}        ACCEPTED
+  human_principal_ids={a different id} rejected   <- control, it is the membership
+
+revocation feed UNREACHABLE (never loaded, fail-closed path):
+  human_principal_ids=set()            rejected   <- control
+  human_principal_ids={that id}        ACCEPTED
+```
 
 Owner: @taOSmd-dev. Supersedes the held Phase 2 of #138, which was blocked on the registry
 identity layer that landed 2026-08-13.
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -54,7 +54,12 @@ _GENERATOR_PROFILE_KEY = "generator_profile"
 _A2A_AUTH_ENFORCE_KEY = "a2a_auth_enforce"
 # Canonical IDs of human principals (controller sessions). These IDs skip the
 # registry revocation check and the grants check; a sub/from mismatch on a
-# human token is always rejected, even in verify-and-warn mode.
+# human token is always rejected, even in verify-and-warn mode. They are also
+# exempt from the fail-closed refusal that fires when the revocation feed has
+# never loaded. The id is not validated as belonging to a human: the controller
+# has no human_principal concept. Naming an agent canonical_id here silently
+# disables that agent's revocation. The set resolves from
+# ``TAOSMD_HUMAN_PRINCIPAL_IDS`` (comma separated) before the config file.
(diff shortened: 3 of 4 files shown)


jaylfc:
## Review: APPROVE and merge

Documentation-only, and every claim it makes is true. I measured them rather than reading them, with controls in both directions, on the trial merge of this branch into master.

### The automated REVIEW WARNING is a false positive, and the card is why

The PR body carries: *"this card's text asks for tests, but the diff changes no test file."* I read `tsk-wa2bwk`'s body rather than trusting the summary. It says the opposite:

> Write NO test for this card. A docstring drift guard would have to assert that some phrase is present, and that kind of test passes on any wording, fails on a harmless rewrite, and cannot see the defect it is named after. We have shipped one of those before and it locked in the bug.

The warning keyed on the word "test" occurring in the card. Changing no test file is the acceptance criterion here, not a violation of it. **The lane obeyed the instruction and did not write the drift guard**, which was this card's explicit trap. Credit where it is due: the card also said "if you find yourself editing a guard, stop and raise it instead", and no guard was edited.

For the record, a bot finding is still a claim worth checking: on #330 this same class of warning was correct and understated. It has to be checked against the card each time, not assumed either way.

### The measurement, reproduced independently on this tree

The docs assert a two-part exemption. Both parts hold, and the instrument discriminates:

```
BLOCK A -- the id IS on the revocation feed (feed loads fine)
  human_principal_ids=set()        CONTROL   rejected   fetches=1   "canonical_id 'principal-x' is revoked"
  human_principal_ids={that id}    ARM       ACCEPTED   fetches=0
  human_principal_ids={other id}   CONTROL   rejected   fetches=1   "canonical_id 'principal-x' is revoked"

BLOCK B -- revocation feed NEVER LOADED (the fail-closed path)
  human_principal_ids=set()        CONTROL   rejected   fetches=1   "revocation feed unavailable, refusing to authorise"
  human_principal_ids={that id}    ARM       ACCEPTED   fetches=0
  human_principal_ids={other id}   CONTROL   rejected   fetches=1   "revocation feed unavailable, refusing to authorise"

BLOCK C -- sub/from mismatch on an exempt id (feed never loaded)
  claimed_from = somebody-else               rejected   fetches=0   HumanAuthError
```

The two blocks reject for **different stated reasons**, which is what makes this a measurement rather than a coincidence: the probe distinguishes the revoked-membership mechanism from the fail-closed mechanism instead of emitting one generic failure. The third row in each block pins that the exemption is *set membership on the token sub*, not a property of the token. `fetches=0` on every exempt arm is the direct evidence for "skip the feed fetch entirely".

Precedence, measured with a positive control (config file always supplies `['from-config-file']`):

```
  no env var                     CONTROL   -> ['from-config-file']
  env set, comma-separated       ARM       -> ['from-env-a', 'from-env-b']
  env set but empty / whitespace edge      -> ['from-config-file']
```

One nuance the new sentence compresses: the rule is *first non-empty wins*, so an empty or whitespace-only env var falls through to the config file. The precise resolution-order block four lines above says so and is unchanged, so nothing is wrong here. Worth knowing if that summary line is ever edited in isolation.

### The safety justification is backed in-repo, not asserted

`registry_auth.authorize` now says the exemption is "safe by design: the feed is agent-only (the controller has no human_principal concept)". That is a claim about an external system, and I would not accept it unbacked in a docs PR. It is not unbacked: `docs/specs/a2a-bus-auth-transition.md` open question 1, already on master, records it as a ruling:

> **the revocation feed is agent-only by design.** The feed is not being extended, no principal store is being added, and `human_principal` is not becoming a controller concept.

Worth stating plainly, because the docstring gives the justification without the caveat the spec attaches to it: that same section records the agent-only boundary as **still owed on the controller side** — *"the feed's own docs must say agent-only, and `list_revoked()` needs a test pinning that boundary rather than leaving it an emergent property of the query."* So the "safe by design" rests on a decision that is documented but not yet pinned by a test upstream. That is @taOS-dev's half and this card does not wait on it. Not a blocker; noted so the next reader has it.

### Acceptance criteria, checked literally

```
zero changes to any if / return / elif / else under taosmd/    NONE      holds
every changed line under taosmd/ is a comment or docstring     verified line by line
```

### The apparent deletion of the sender census is base drift, not a loss

`git diff origin/master origin/exec/tsk-wa2bwk` shows the census being removed across five files. **That is master having moved, not this branch.** The branch was cut at `d487007a`, before #329 merged.

```
git diff d487007a origin/exec/tsk-wa2bwk --stat   ->   4 files, +36, -5   (the true change)
after trial merge, grep -c a2a_sender_census:
  service.py 3   remote.py 1   http_server.py 1   tests 9   changelog 1     ALL PRESERVED
```

Diff a branch against its merge-base before reporting a deletion. A real loss and this non-loss produce the same shape of stat block; only one is a defect.

### Gates and suite

```
conflict markers            none
deleted-symbols-guard       clean
normalise-handle-gate       clean
witness-gate                clean
full suite on trial merge   1501 passed, 12 skipped
```

1501 is **exactly** the current master baseline. For a comment-and-docstring diff an unchanged count is the expected answer, and it is the third witness that no behaviour moved.

Merging.

PR #332, opened 2026-08-18
Revise PR #325: the harness keys on utilization, so it cannot tell a rollover from ordinary usage, and the default invocation never polls

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #325: the harness keys on utilization, so it cannot tell a rollover from ordinary usage, and the default invocation never polls

Autonomous build of board card tsk-3te4pi.

Key the flip detection in measure_at_boundary on resets_at (window
identity) instead of utilization, which could not distinguish a rollover
from ordinary account consumption on a shared account.  Start the
polling budget (max_wait) after the boundary arrives rather than at
process start, and sample the pre-boundary baseline before the boundary
so a propagation faster than the first poll is measurable.  Align the
evidence file's target_location and procedure text with the code, and
split the flip-detection test into ARM A (consumption without rollover)
and ARM B (rollover with flat utilization) as disagreement controls.

Full suite: 1517 passed, 12 skipped.

Files:
 benchmarks/measure_upstream_propagation.py         | 294 +++++++++++++++++
 .../results/tsk-rcnct6_upstream_propagation.json   |  11 +
 .../tsk-3te4pi-measure-upstream-propagation.md     |  16 +
 .../tsk-rcnct6-measure-upstream-propagation.md     |   8 +
 tests/test_measure_upstream_propagation.py         | 347 +++++++++++++++++++++
 5 files changed, 676 insertions(+)

### DIFF
--- a/benchmarks/measure_upstream_propagation.py
+++ b/benchmarks/measure_upstream_propagation.py
@@ -1,0 +1,294 @@
+#!/usr/bin/env python3
+"""Measure upstream usage-window rollover propagation via the real API.
+
+At an Anthropic usage-window reset boundary, the five_hour window
+identity (resets_at) does not update instantly.  This harness polls the
+REAL Anthropic usage API at https://api.anthropic.com/api/oauth/usage,
+one request per second, and records the wall-clock delta between the
+nominal resets_at and the first response serving the new window.
+
+The constant this measures -- MIN_LEAD_SECONDS in the out-of-repo
+resume-arm-time helper -- is labelled UNMEASURED until a live
+measurement succeeds.  If no credentials are available, the run writes
+an UNMEASURED evidence record to benchmarks/results/ so the label is
+never silently retired.
+
+Usage:
+    TAOSMD_ANTHROPIC_CREDS=/path/creds.json python3 benchmarks/measure_upstream_propagation.py
+    python3 benchmarks/measure_upstream_propagation.py --reset-at 2026-08-17T08:00:00+00:00 --creds /path/creds.json
+    python3 benchmarks/measure_upstream_propagation.py --dry-run --creds /path/creds.json
+"""
+
+from __future__ import annotations
+
+import argparse
+import datetime
+import json
+import os
+import sys
+import time
+from pathlib import Path
+
+import httpx
+
+_BENCH_DIR = Path(__file__).resolve().parent
+_REPO_ROOT = _BENCH_DIR.parent
+_RESULTS_DIR = _BENCH_DIR / "results"
+_EVIDENCE_FILE = _RESULTS_DIR / "tsk-rcnct6_upstream_propagation.json"
+
+USAGE_API_URL = "https://api.anthropic.com/api/oauth/usage"
+
+_HEADERS = {
+    "anthropic-beta": "oauth-2025-04-20",
+}
+
+
+def resolve_token(creds_path: str | None = None) -> str:
+    """Resolve the Anthropic OAuth token from a credentials file.
+
+    The path is configurable via --creds or TAOSMD_ANTHROPIC_CREDS so no
+    hardcoded absolute path is baked into the source.
+    """
+    path = creds_path or os.environ.get("TAOSMD_ANTHROPIC_CREDS")
+    if not path:
+        raise SystemExit(
+            "No credentials path: set TAOSMD_ANTHROPIC_CREDS or pass --creds"
+        )
+    with open(path, "r") as f:
+        creds = json.load(f)
+    return creds["claudeAiOauth"]["accessToken"]
+
+
+def fetch_usage(token: str, client: httpx.Client | None = None) -> dict:
+    """Fetch live usage data from the real Anthropic usage API.
+
+    Returns the parsed JSON body.  Does NOT return canned or mock data.
+    """
+    owns_client = client is None
+    if owns_client:
+        client = httpx.Client(timeout=15.0)
+    try:
+        resp = client.get(
+            USAGE_API_URL,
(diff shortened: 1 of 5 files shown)


jaylfc:
## Review: APPROVE and merge

Third pass at this quantity, and the first one that measures what it claims to measure. All four blockers are genuinely fixed, every Credit item survived, and the red-first evidence is the disagreement control the card asked for rather than a test that merely passes.

I re-derived everything below on a trial merge into current master. Nothing here is taken from the PR body.

### The disagreement control, run independently: old vs new, same input

Both versions imported by path and called directly, real sleeps, elapsed time asserted non-zero so the durations mean something.

```
INPUT                                            OLD (#325)          NEW (#332)          HONEST ANSWER
-----------------------------------------------------------------------------------------------------
CONTROL  resets_at 08->13, util 0.80->0.20       MEASURED            MEASURED            MEASURED     (agree)
ARM A    resets_at PINNED 13:00, util 0.20->0.22 MEASURED            NO_FLIP_DETECTED    NO_FLIP      <-- DISAGREE
ARM B    resets_at 08->13, util 0.0 FLAT         NO_FLIP_DETECTED    MEASURED            MEASURED     <-- DISAGREE
```

The control is what makes this readable: both versions reach the right answer when both signals move together, so the instrument is not simply broken in one direction. The two arms then split, and in both the new version gives the honest answer. **BLOCKER 1 fixed.** `measure_at_boundary` now keys on `cur_resets_at != pre_resets_at`, with utilization kept only as a recorded corroborating signal, and the comment at the guard explains why neither direction of utilization can be the trigger.

**BLOCKER 2 fixed**, measured on the default-shaped invocation (boundary further out than the whole polling budget: boundary 3s, `max_wait` 2s):

```
                              OLD fetches after boundary   NEW fetches after boundary
default-shaped invocation     0   (NO_FLIP_DETECTED)       2   (MEASURED)
```

`t0` now starts after the boundary arrives, and the pre-boundary baseline is sampled *before* the wait rather than being taken from the first post-boundary response, so a propagation faster than the first poll is measurable by construction. Both halves of that fix are present and both are exercised.

### BLOCKER 4: the tests now fail when the behaviour is absent

This is the check that decided it. I swapped #325's harness into the tree and ran **this PR's test file unchanged** against it:

```
new tests vs NEW harness (as shipped)   21 passed              <- positive control
new tests vs OLD harness (#325)          5 failed, 16 passed

failing: test_arm_a_consumption_without_rollover_not_measured
         test_arm_b_rollover_with_flat_utilization_is_measured
         test_polling_budget_starts_after_boundary
         test_pre_reset_baseline_in_no_flip_result
         test_evidence_procedure_matches_code
```

The discrimination is correct in both directions: the five behavioural tests go red on the defective harness, and the sixteen that stay green are the source-text credit-item pins, which *should* pass on #325's tree because #325 genuinely fixed those. Contrast with what this replaces, where `test_measure_at_boundary_detects_flip` passed 17/17 with the defect present and blessed it.

### BLOCKER 3 fixed

The committed evidence's procedure now reads "...the first response serving the new five_hour window **(detected by resets_at changing)**", which is what the code does. The record and the code it cites now say the same thing.

Both minors are fixed too: `target_location` no longer cites the non-existent `scripts/resume_arm_time.py` and instead says plainly that the constant lives in an out-of-repo helper, and the suite count is reported ("1517 passed, 12 skipped", correct against the base this branch was cut from).

### Credit items: all preserved, checked against the shipped file directly

```
no MockAPI / window_flips / simulate      OK      real endpoint api.anthropic.com/.../usage   OK
uses httpx                                OK      no /tmp path                                OK
no ~/.taos-team or fleet-tools path       OK      no subprocess / sed / shutil                OK
status UNMEASURED, measured_samples []    OK      MIN_LEAD_SECONDS unlabelled, untouched      OK
no scripts at repo root                   OK      evidence committed under benchmarks/results OK
```

I ran the real CLI entry point rather than only the helper. It writes solely inside the repo, and the shared helper three lead agents depend on is untouched:

```
before/after   ~/.taos-team/resume_arm_time.py -> ~/.taos-fleet-tools/resume_arm_time.py
               md5 7c54e86e32ba1bd33868b42d69ecfb5d  (identical)
```

That check matters here specifically because reaching outside the repo is what #288 was blocked for, and because the same shared path actually broke the fleet once last night.

### One note, not a blocker

Running the CLI regenerates the evidence file, and the regenerated JSON is not byte-identical to the committed one: indent 2 vs 4, no trailing newline, `reason` loses the sentence about the bounded wait across a real 5h boundary, and `procedure` says "in this file" rather than naming the script. So the committed record is a hand-polished version of the script's output, not the output itself.

Nothing here contradicts anything, which is why it is not a blocker: both texts now correctly say the flip is detected by `resets_at` changing, and the committed copy carries strictly more information. Worth knowing because a future run overwrites the fuller `reason` with the shorter one. If that record is meant to be reproducible, the generator should emit exactly what is committed.

Also operational, and inherent to the specified fix rather than a defect in it: with the budget now starting at the boundary, the default invocation can block for up to five hours waiting for one, since the pre-boundary wait is deliberately not counted against `max_wait`. That is what the card asked for and what measuring a boundary requires. `--reset-at` and `--dry-run` are the escape hatches.

### Gates and suite

```
conflict markers            none
deleted-symbols-guard       clean
normalise-handle-gate       clean
witness-gate                clean
full suite on trial merge   1522 passed, 12 skipped
```

1522 is exactly the current master baseline of 1501 plus the 21 tests in the new file, which reconciles with what the diff touched and confirms nothing else moved. The branch's own reported 1517 was honest against the base it was cut from; master has since advanced by #329 and #331.

Merging.

PR #333, opened 2026-08-18
Revise PR #328: the carry-forward landed and none of the three fixes did

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #328: the carry-forward landed and none of the three fixes did

Autonomous build of board card tsk-2ohqj4.

- Sibling arms of the same if/elif/else chain and try/except/else block no
  longer collide; a def in an arm still collides with a def at module level
- Accept ModuleNotFoundError and builtins.ImportError as legitimate import
  fallback spellings alongside bare ImportError
- Recurse into nested classes so class Outer.Inner duplicate methods fire
- Add scope-parity corpus test pinning every claimed scope and verdict
- Add changelog fragment for tsk-2ohqj4

Files:
 .../tsk-2ohqj4-normalise-handle-gate-fixes.md      |   2 +
 scripts/normalise_handle_gate.py                   | 218 ++++++++++++--
 tests/test_normalise_handle_gate.py                | 323 ++++++++++++++++++++-
 3 files changed, 518 insertions(+), 25 deletions(-)

### DIFF
--- a/changelog.d/tsk-2ohqj4-normalise-handle-gate-fixes.md
+++ b/changelog.d/tsk-2ohqj4-normalise-handle-gate-fixes.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate no longer fires on definitions that sit in mutually exclusive sibling arms of the same ``if``/``elif``/``else`` chain or ``try``/``except``/``else`` block. The ``except ImportError:`` and ``except ModuleNotFoundError:`` fallback patterns are recognised and left silent. Nested-class duplicate methods are now scanned at any depth, with the correct ``class Outer.Inner`` scope string.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -11,6 +11,16 @@ functions -- and any other accidental redefinition -- are reported.
 ``@<name>.getter`` / ``@<name>.deleter`` accessories are legitimate same-name
 pairs and are therefore excluded.  Files that cannot be decoded as UTF-8 are
 skipped with a warning rather than allowed to crash the gate.
+
+Definitions inside module-level ``if`` / ``try`` / ``for`` / ``while`` /
+``with`` blocks are treated as module-scope, matching Python's binding rules.
+Same-name closures inside one parent function are also reported; same-name
+closures in different parents remain legal.  Sibling arms of the same
+``if`` / ``elif`` / ``else`` chain and the ``body`` / ``handlers`` /
+``orelse`` arms of one ``try`` statement are mutually exclusive and do not
+collide.  The ``try:`` / ``except ImportError:`` and ``except
+ModuleNotFoundError:`` fallback patterns are recognised and left silent.
+Nested classes are scanned at any depth.
 """
 from __future__ import annotations
 
@@ -30,6 +40,16 @@ class Duplicate:
     name: str
     scope: str
     lines: list[int] = field(default_factory=list)
+
+
+@dataclass
+class _Def:
+    scope: str
+    name: str
+    lineno: int
+    in_try: bool = False
+    in_import_error_except: bool = False
+    arm_tracker: tuple[int, str] | None = None
 
 
 def _has_overload_decorator(node: ast.FunctionDef | ast.AsyncFunctionDef) -> bool:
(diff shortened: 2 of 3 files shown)


jaylfc:
## Review: APPROVE and merge

This is the fourth pass over this gate and the first one where every blocker actually moved. All ten cases from the card's table now give the verdict the card asked for, the four controls still behave, and the scope-parity corpus the card asked for exists and has real power.

Three defects were found that are **not** behaviour regressions and are carded as `tsk-w2ugkt` rather than blocked. Reasoning for that call is at the bottom.

### The card's table, re-measured independently

Each gate version imported by path and `_duplicate_definitions(Path(fixture))` called directly, so `REPO_ROOT` is never consulted.

```
CASE                                     master   #328     #333     WANTS    ok?
CTRL+ modlevel def then def in if-body   silent   FIRES    FIRES    FIRES    ok
CTRL+ plain module duplicate             FIRES    FIRES    FIRES    FIRES    ok
CTRL- closures in DIFFERENT parents      silent   silent   silent   silent   ok
CTRL- ImportError fallback               silent   silent   silent   silent   ok
B1 platform if/else arms                 silent   FIRES    silent   silent   ok
B1 elif chain arms                       silent   FIRES    silent   silent   ok
B1 try/except/else sibling arms          silent   FIRES    silent   silent   ok
B2 ModuleNotFoundError fallback          silent   FIRES    silent   silent   ok
B2 ModuleNotFoundError in a tuple        silent   FIRES    silent   silent   ok
B3 nested-class duplicate methods        FIRES    silent   FIRES    FIRES    ok
```

My `master` and `#328` columns reproduce the card's recorded table exactly, which is what tells me the instrument is sound rather than agreeing with itself. Nested-class scope string is exactly `class Outer.Inner` with lines `[3, 4]`, matching master.

The fix for blocker 1 is the right one conceptually: sibling arms are tracked by `(statement id, arm)` and a definition with no arm tracker always collides, so `def f` at module level still collides with `def f` in an `if` body while the `if` and `else` arms do not collide with each other. That is the binding truth the rule is supposed to encode.

### The tests have power

```
this PR's test file vs #333 gate (as shipped)   44 passed        <- positive control
this PR's test file vs #328's gate               6 failed, 38 passed

failing: test_if_else_sibling_arms_are_silent
         test_elif_chain_arms_are_silent
         test_try_except_sibling_arms_are_silent
         test_try_except_else_sibling_arms_are_silent
         test_nested_class_duplicate_methods_fire
         test_scope_parity_corpus
```

`test_scope_parity_corpus` is in that list, which is the point of the systemic ask: it asserts `actual == expected` over the full verdict tuple (scope string, name, exact line numbers) from literals, not from recomputed production logic, so a scope that silently stops firing fails loudly. That is the thing neither #316 nor #321 had.

### Acceptance items 6 through 9

```
standing control exec/tsk-v53vta   exit 1, all 3 shadowed tests at [624,1162] [648,1188] [670,1210]
real tree, 183 files               normalise-handle-gate: clean, exit 0
git rev-list --merges              0
gate blob vs exec/tsk-bpadgg       0cbab67b -> 9ffd023c   (the check #328 failed)
four repo gates on trial merge     all clean, no conflict markers
full suite on trial merge          1539 passed, 12 skipped
```

1539 is the current master baseline of 1522 plus exactly the 17 net new tests in this file (27 on master, 44 here), so it reconciles with what the diff touched.

### Three findings, carded as `tsk-w2ugkt`

**1. `in_import_error_except` is dead code.** `_is_import_error_handler` (`:92`) sets a flag stored on `_Def` and threaded through eleven call sites, and `_duplicate_definitions` never reads it: the dedup branches only on `arm_tracker`. Patching `_is_import_error_handler` to `return False` changes no verdict on any input I could construct, and the whole 44-test file still passes.

This went dead for a good reason. #321 added the special case because its widening made the fallback fire; this PR then implemented the general truth that sibling arms are mutually exclusive, which subsumes it entirely. The exception spelling no longer changes any verdict.

**2. The three spelling tests have zero power.** `test_module_not_found_error_bare_is_silent`, `..._in_tuple_is_silent` and `test_builtins_import_error_is_silent` each build a fixture holding one `def foo()` and a `try: import something / except ...: pass`. There is no duplicate in the input at all, so the result is `[]` on every gate ever written, including master's, #328's, and one with the feature deleted. The corpus case `module_not_found_error_fallback_silent` has the same shape. The fix is to put a real duplicate in the fixture, one `def` in the `try` body and one in the handler, which is the shape that actually fired on #328.

**3. The widening changelog entry was deleted.** `changelog.d/tsk-xhvo4a-duplicate-definition-gate-scope.md` is present on `exec/tsk-bpadgg` and absent here. It described #321's widening, which this PR still ships in the carried-forward code. The new entry describes only the false-positive fixes, so the larger user-visible change lands with no changelog line. The card asked to add an entry, not to remove that one.

### Why these are carded and not blocked

The behaviour is correct on every case the card names, and the behaviour that finding 2's tests nominally protect **is** pinned, by the sibling-arm tests, which genuinely fail against #328's gate. So nothing here locks in a bug: finding 1 is a vestigial flag that does nothing, finding 2 is coverage that is redundant rather than load-bearing, and finding 3 is a record gap. Blocking a fourth time over that would discard a correct gate fix and a corpus test this gate has needed for three rounds.

Worth saying plainly, because it cuts the other way on a different day: a test that passes with its feature deleted is normally a blocking defect here. It is not blocking **this** time only because the feature it names turned out to be dead, so the test is documenting a mechanism rather than guarding one.

Credit where it is due: STEP 0 was carried forward cleanly again, the gate diff against `exec/tsk-bpadgg` is real this time rather than an identical blob, and the corpus test is exactly what was asked for.

Merging.

PR #335, opened 2026-08-18
Revise PR #330: the scope was never widened, the citation was never fixed, and the de-marking added a third live marker

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #330: the scope was never widened, the citation was never fixed, and the de-marking added a third live marker

Autonomous build of board card tsk-ilz33f.

PR #330 claimed three changes but delivered only one: a docstring
rewording that actually increased the live-marker count from 2 to 3.
This revision fixes all three blockers:

- Widen _source_files() to scan scripts/**/*.py alongside taosmd/**/*.py,
  closing the gap where witness markers in gate scripts were never verified.
- Correct the sibling-gate citation from normalise_handle_gate.py (which
  also scans tests/) to check_deleted_symbols.py (which is taosmd/-only,
  matching the exclusion). State the real reason tests/ is excluded:
  fixture markers built as f-strings whose names do not resolve to files.
- De-mark every illustrative WITNESS example in the gate docstring (L4,
  L19) and the test module docstring (L5) with a zero-width space between
  WITNESS and :, so the gate's own regex reports 0 live-marker matches.

Add RED, GREEN, and non-vacuity controls for the scripts/ scope. Full
suite: 1505 passed, 12 skipped (master baseline was 1501 passed, 12
skipped; the 4 new tests account for the delta).

Files:
 changelog.d/tsk-ilz33f-witness-gate-rev.md |  8 +++++
 scripts/check_witness_token.py             | 19 +++++++----
 tests/test_witness_gate.py                 | 54 +++++++++++++++++++++++++++++-
 3 files changed, 73 insertions(+), 8 deletions(-)

### DIFF
--- a/changelog.d/tsk-ilz33f-witness-gate-rev.md
+++ b/changelog.d/tsk-ilz33f-witness-gate-rev.md
@@ -1,0 +1,8 @@
+### Fixed
+- Widened the witness-token gate scan to include ``scripts/**/*.py`` so that
+  witness markers in gate scripts are checked, closing the gap where a cited
+  constant in ``scripts/`` was never verified. Fixed the sibling-gate citation in
+  the docstring from ``normalise_handle_gate.py`` (which also scans ``tests/``)
+  to ``check_deleted_symbols.py`` (which is ``taosmd/``-only, matching the
+  exclusion). De-marked illustrative marker examples in the gate docstring and
+  the test module docstring so the live-marker count is 0.
--- a/scripts/check_witness_token.py
+++ b/scripts/check_witness_token.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """Witness token gate.
 
-Verifies that every explicit ``# WITNESS: <test>::<token>`` marker in a source
+Verifies that every explicit ``# WITNESS​: <test>::<token>`` marker in a source
 file resolves to something real INSIDE the named test: the referenced test file
 must exist and ``<token>`` must appear in it as a case-sensitive substring (grep
 semantics).
@@ -16,7 +16,7 @@ test-filename mention in ordinary prose.
 
 Marker contract (the only thing that triggers the check):
 
-    # WITNESS: tests/test_foo.py::some_grepable_token
+    # WITNESS​: tests/test_foo.py::some_grepable_token
 
 - The ``WITNESS:`` marker is all-caps and opt-in. A bare mention of a test
   filename in prose -- whether describing a removal or stating a fact --
@@ -31,10 +31,12 @@ Marker contract (the only thing that triggers the check):
 - The token is matched as a case-sensitive substring, the same match ``grep``
   would perform.
 
-Scope: ``taosmd/**/*.py``. Witnesses are declared in source, not in tests --
-scanning ``tests/`` for markers would let a test's own example markers (which
-name fixture trees that do not exist in the checked-out repo) produce false
-violations. The sibling normalise-handle gate uses the same ``taosmd/`` scope.
+Scope: ``taosmd/**/*.py`` and ``scripts/**/*.py``. Witnesses are declared in
+source, not in tests -- the suite embeds fixture markers as f-strings
+(e.g. ``f"# WITNESS​: {TEST}::{TOK}"``) whose interpolated names never
+resolve to real files, so scanning ``tests/`` would produce false violations.
+The sibling ``check_deleted_symbols.py`` gate follows the same pattern: it
+restricts to ``taosmd/`` only, excluding ``tests/``.
 
 An optional ``WITNESS_GATE_ROOT`` environment variable overrides the repo root
 the gate scans (defaulting to the tree containing this script). This lets the
@@ -119,7 +121,10 @@ def _read_text(path: Path) -> str:
 
 
 def _source_files(repo_root: Path) -> list[Path]:
-    return sorted(set(repo_root.glob("taosmd/**/*.py")))
+    return sorted(
+        set(repo_root.glob("taosmd/**/*.py"))
+        | set(repo_root.glob("scripts/**/*.py"))
+    )
 
 
 def _check_claim(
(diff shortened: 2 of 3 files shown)


jaylfc:
## MERGE. All four blockers fixed, each proven with a control that disagrees.

This is a genuine revision, not the blocked tree under a new name. That check first, because four
revision PRs in a row (#323, #327, #328) turned out to be the unchanged tree:

```
git diff origin/exec/tsk-7oak6l origin/exec/tsk-ilz33f --stat   ->  8 files, 215 insertions
```

That number is misleading on its own. Diffed against **each branch's own merge-base**, the real
contributions are:

```
#330 (merge-base d487007a) : scripts/check_witness_token.py                | 3 lines
#335 (merge-base 631168b5) : changelog + check_witness_token.py + tests    | 73 lines
```

The other five files in the branch-to-branch diff are base drift (#331/#332/#333 landed in
between), not this PR's work.

### BLOCKER 1 -- scope widened. Proven on the REAL TREE, not a fixture.

Injected a bogus marker into a new `scripts/` file and ran both gate versions over the same tree:

```
INPUT                                        OLD gate (master)   NEW gate (#335)
scripts/_probe_victim.py with a WITNESS
marker naming a test file that DOES NOT      clean, EXIT 0       WITNESS GATE FAIL, EXIT 1
exist                                                            "test file not found"
                                             ------------------- DISAGREE -------------------
```

The widening has power. `_source_files()` is now the union of `taosmd/**/*.py` and
`scripts/**/*.py`.

### BLOCKER 2 -- citation fixed, and both new claims verified against the sibling sources.

The card asked for the claim to be true, not merely changed. Checked against the actual files:

| claim | where | verdict |
|---|---|---|
| `check_deleted_symbols.py` "restricts to `taosmd/` only, excluding `tests/`" | `check_deleted_symbols.py:152,171,248` all filter `member.name.startswith("taosmd/")` | **TRUE** |
| changelog: `normalise_handle_gate.py` "also scans `tests/`" | `normalise_handle_gate.py:25` `TARGET_PATTERNS = ("taosmd/**/*.py", "tests/**/*.py")` | **TRUE** |
| the OLD docstring's "sibling normalise-handle gate uses the same `taosmd/` scope" | same line 25 | **FALSE, correctly removed** |

### BLOCKER 3 -- live-marker count 2 -> 0, measured with the gate's own regex.

Ran `re.compile(r"#\s*WITNESS:\s*(.+)$")` over every `.py` file in each tree:

```
                          IN-SCOPE live markers      module docstring of tests/test_witness_gate.py
master                    2  (both in the gate)      1 live marker
#335                      0                          0 live markers
```

De-marking is a zero-width space (U+200B) between `WITNESS` and `:`, at file offsets 87, 905 and
1939. The regex requires a literal colon after `WITNESS`, so it cannot match. **This technique was
explicitly authorised by the card** ("break the token (`WITNESS‑`)"), and the PR body
discloses the mechanism in plain words rather than leaving it to be discovered. Both correct.

It does carry a hazard, which I have **carded as `tsk-g4jjtp` rather than blocked on**, since the
card sanctioned the approach: the canonical contract example, copied verbatim and edited to real
values, is a marker that looks live and is never checked. Filed with the measurement and two
suggested fixes (document the U+200B in the docstring; report marker-shaped near-misses as
violations).

### BLOCKER 4 -- tests added, and they have power.

The strongest check available: run the NEW test file UNCHANGED against the OLD implementation.

```
new tests vs NEW gate (positive control) : 21 passed
new tests vs OLD gate (master's)         :  2 failed, 19 passed
  FAILED test_scripts_scope_red_absent_token
  FAILED test_scripts_scope_non_vacuity
```

(The swap was verified to have applied before I read the result -- `scripts-glob occurrences in
the swapped file: 0`. A neutering probe that never engaged has burned me before.)

Worth naming why the other two new tests stay green against master: `test_scripts_scope_green_present_token`
and `test_green_prose_in_scripts` are **vacuous in isolation** -- under master's gate they pass
because `scripts/` is never scanned at all, which is the right answer for the wrong reason. What
gives them meaning is the paired `test_scripts_scope_non_vacuity`, which deletes ONLY the token,
leaves the file present and importable, and requires the flip. Green controls backed by a
non-vacuity companion is exactly the structure the card asked for.

### Suite: 1543 passed, 12 skipped, reconciled against what the diff touched.

```
master baseline 56933630 : 1539 passed, 12 skipped
trial merge     bf8b575b : 1543 passed, 12 skipped     (+4 = the 4 new tests, exactly)
```

The PR body reports **1505 passed / 1501 baseline**, which is correct for **its own** merge-base
(631168b5) and reconciles the same way (+4). Master has since moved to 1539; both numbers are
honest, they just measure different bases.

Five load-bearing checks on the trial merge: no conflict markers; `check_deleted_symbols.py
--base origin/master` clean; `normalise_handle_gate.py` clean; `check_witness_token.py` clean;
full suite green. `taosmd.__file__` confirmed resolving inside the trial tree.

### Two deviations from the acceptance list, stated rather than glossed

1. **Card item 4 asked for the RED/GREEN controls "against the real tree, not only a fixture".**
   The four new tests use `tmp_path` fixtures. I ran the real-tree version myself (the BLOCKER 1
   table above) and it behaves. I am not treating this as a miss: a test that writes into the
   repo's own `scripts/` to make its point would be invasive and order-dependent, and the fixture
   plus `WITNESS_GATE_ROOT` is the better engineering. Noting it so the record is accurate.

2. **My own card over-specified item 2.** It demanded the live-marker count be 0 "in each file",
   naming `tests/test_witness_gate.py`. That file still has 11 matches on this branch and always
   will -- they are the f-string fixture markers that ARE the test data, and `tests/` is out of
   scope by design. The intent (de-mark the illustrative module-docstring example) is met. The
   criterion as I wrote it was unmeetable; that is my error, not the executor's.

### Not fixed here, carried to `tsk-g4jjtp`

The `scripts/` widening currently verifies **zero** markers -- there are none in `scripts/` now
that the docstring examples are de-marked. The coverage is real but entirely prospective, so the
changelog's "closing the gap where a cited constant in `scripts/` was never verified" points at an
instance that does not exist. One-line wording fix, carded.

Merging.

PR #336, opened 2026-08-18
Revise PR #327: the carry-forward landed but neither blocker was fixed; ?reader= still overrides the verified token and the archive limit is still missing

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #327: the carry-forward landed but neither blocker was fixed; ?reader= still overrides the verified token and the archive limit is still missing

Autonomous build of board card tsk-mxjvnb.

Carry forward the PR #320 mention feed work unchanged, then fix both
blockers:

1. /a2a/mentions now fails closed with 403 when ?reader= does not
   normalise via _normalise_handle to the verified token sub. The
   claimed identity is passed to authorize so the signature check
   is meaningful. A mismatch never leaks the caller's own mentions.

2. a2a_mentions_feed archive query now carries limit=100_000, matching
   every sibling bus read in service.py. Without it the default 50-row
   cap silently dropped mentions once the bus exceeded 50 messages.

Add two disagreement-control tests: carol's token with ?reader=bob
returns 403 and bob's body is absent; a mention remains visible at
bus size 120.

changelog.d/tsk-mxjvnb-mentions-auth-and-limit.md

Files:
 changelog.d/tsk-mxjvnb-mentions-auth-and-limit.md |   3 +
 taosmd/api.py                                     |   4 +
 taosmd/http_server.py                             |  61 ++-
 taosmd/mentions.py                                |  92 ++++
 taosmd/remote.py                                  |  20 +
 taosmd/service.py                                 | 172 ++++++-
 tests/test_a2a_mentions.py                        | 592 ++++++++++++++++++++++
 8 files changed, 947 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added a mentions feed for retrieving messages that mention a reader, including relevant reply threads.
  * Added optional recipient support for messages.
  * Added filtering by timestamp and result limit.
  * Added remote and HTTP access to mentions feeds.

* **Bug Fixes**
  * Improved handle matching and excluded email addresses and URLs from mention detection.
  * Enforced reader authorization and channel-specific reply visibility.
  * Added validation for invalid limits and mismatched reader identities.
  * Improved performance for large mentions feeds.

### DIFF
--- a/changelog.d/tsk-6icvd4-mentions-feed.md
+++ b/changelog.d/tsk-6icvd4-mentions-feed.md
@@ -1,0 +1,7 @@
+### Fixed
+- `/a2a/mentions` now sends `reader` over the remote path and accepts it as a query parameter when auth is enabled, preventing silent identity substitution.
+- Handles are normalised via the shared `_normalise_handle` helper so `@bob`, `bob`, and `BOB` resolve to the same identity.
+- Cross-channel reply chains are excluded from the mentions feed; only replies within the same thread are followed.
+- `limit` is applied once; `-1`, `0`, `nan`, `inf`, and absurd values are rejected with 400 instead of returning empty or 500.
+- Mention regex no longer matches `@` inside email addresses or URLs.
+- Dead code removed from `can_read`; quadratic reply-chain traversal replaced with an O(n) adjacency-list walk.
--- a/changelog.d/tsk-mxjvnb-mentions-auth-and-limit.md
+++ b/changelog.d/tsk-mxjvnb-mentions-auth-and-limit.md
@@ -1,0 +1,3 @@
+### Fixed
+- `/a2a/mentions` now fails closed when `?reader=` does not match the verified token subject, instead of serving the caller's own mentions
+- `a2a_mentions_feed` archive query now carries `limit=100_000` so mentions remain visible when the bus exceeds 50 messages
--- a/taosmd/api.py
+++ b/taosmd/api.py
@@ -144,12 +144,16 @@ async def _ensure_stores(data_dir=None) -> dict:
         from taosmd.claims.store import ClaimStore  # noqa: PLC0415
         claims = ClaimStore(db_path=str(path / "claims.db"))
         await claims.init()
+        from taosmd.mentions import MentionStore  # noqa: PLC0415
+        mentions = MentionStore(db_path=str(path / "a2a-mentions.db"))
+        await mentions.init()
 
         stores = {
             "archive": archive,
             "vector": vmem,
             "kg": kg,
             "claims": claims,
+            "mentions": mentions,
             "data_dir": str(path),
         }
         _stores_cache[resolved] = stores
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -109,6 +109,8 @@ Endpoints
 ``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream); ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400
 ``GET  /a2a/threads``      ``?principal=``                  -> ``{"threads": [...]}`` thread list for the principal, ordered by latest activity desc; each entry has ``thread``, ``kind``, ``participants``, ``last_message: {id, ts, from, body_preview}`` (see notes below)
 ``GET  /a2a/threads/{thread}/messages`` ``?before=&after=&limit=`` -> ``{"thread", "messages": [...]}`` cursor-paginated message envelope, oldest-first
+``GET  /a2a/mentions``    ``?since=&limit=&reader=``                  -> ``{"messages": [...]}`` (requires registry auth; ``reader`` is derived from the verified token ``sub``, or supplied as a query parameter when no verifier is configured)
+``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream)
 ``GET  /a2a/channels``                                     -> ``{"channels": [...]}``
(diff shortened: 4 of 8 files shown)


jaylfc:
## BLOCK. Both blockers are genuinely fixed; three new defects, one of them a silent regression on master.

Revision card: **`tsk-u74ivj`**. Closing under the close-on-block policy. The branch is preserved.

### First, what this PR got right, because it is most of the work

Checked against merge-bases, not branch to branch. `git merge-base origin/master origin/exec/tsk-mxjvnb` = `5693363`, real contribution 8 files / +947-4. Trial merge into `746f00f9` clean, no conflict markers.

- **Suite on the trial merge: 1570 passed, 12 skipped.** 1543 baseline plus exactly the 27 tests in `tests/test_a2a_mentions.py`, reconciled exactly, above the 1521 the card required.
- **New test file run UNCHANGED against the old implementation (`origin/exec/tsk-wlq4ui`): 2 failed, 25 passed.** Positive control 25/25; file md5 `e032022c` identical on both trees. Both required tests really do discriminate.

BLOCKER 1, end to end on the merged tree with the PR's own `authed_server` fixture:

```
CONTROL  bob   token, no ?reader=      -> 200 ['secret for @bob']
CONTROL  carol token, no ?reader=      -> 200 ['note for @carol']
CONTROL  carol token, ?reader=carol    -> 200 ['note for @carol']
ARM      carol token, ?reader=bob      -> 403      <- was 200 ['secret for @bob'] on #327
LEAKED "secret for @bob" into carol's response: False
```

BLOCKER 2: `service.py:983` carries `limit=100_000`, matching its five siblings; the 120-message boundary test fails `assert 'old @bob' in []` on the old tree and passes here. All three gates clean apart from the one below.

### DEFECT 1 (blocking) — `a2a_sender_census` is dropped from `service.__all__`

`scripts/check_deleted_symbols.py --base origin/master` flags it and it is real. The `__all__` line was rewritten to add this PR's two entries and lost an unrelated one in the same edit:

```
origin/master  service.py:1327   ... "a2a_channels", "a2a_sender_census",
merged #336    service.py:1491   ... "a2a_channels", "a2a_members",
```

Measured with `from taosmd.service import *`, controls in both directions:

```
                      origin/master   merged #336
a2a_sender_census        True            False      <- ARM, regression
a2a_channels             True            True       <- CONTROL, unchanged
a2a_mentions_feed        False           True       <- CONTROL, intended
can_read                 False           True       <- CONTROL, intended
```

The function, route, `RemoteClient` method and all four of its tests survive, so `hasattr` stays True and **all 1570 tests still pass**. Only the export is gone — a silent public-API regression on a symbol this PR has nothing to do with. One line to restore. Please do not silence the gate with a `Removes-Intentionally:` trailer; the removal was not intentional.

### DEFECT 2 (blocking) — the honest `?reader=` round-trip 403s on every non-byte-exact handle

The card required both halves: fail closed on mismatch, **and** serve a `?reader=` that matches under the shared `_normalise_handle`. Only the first works. `_normalise_handle` is `handle.lstrip("@").casefold()`, so all four below normalise to `carol`; measured with carol's own valid token:

```
?reader=carol     normalises-to-sub=True  -> 200 ['note for @carol']
?reader=@carol    normalises-to-sub=True  -> 403 "token sub 'carol' does not match from '@carol'"
?reader=Carol     normalises-to-sub=True  -> 403 "token sub 'carol' does not match from 'Carol'"
?reader=@Carol    normalises-to-sub=True  -> 403 "token sub 'carol' does not match from '@Carol'"
```

`http_server.py:1701` passes `claimed = qp_reader` into `authorize`, and `registry_auth.py:82` compares with a raw `if sub != claimed_from:`. That raises first, so **the `_normalise_handle` comparison at `http_server.py:1706-1710` can never return 403 — it is dead code.** The tolerance the card asked for is present in the source and inert.

It is not cosmetic: `remote.py:265` forwards `reader` **unconditionally**, so the shipped client always takes this path, and local and remote now disagree on the same input —

```
LOCAL  service.a2a_mentions_feed('carol'|'@carol'|'Carol'|'@Carol')  -> all 4 return ['note for @carol']
HTTP   the same four as ?reader= with a valid matching token         -> 1 served, 3 x 403
```

A caller who passes `@carol` works locally and is rejected remotely by an impersonation error naming themselves. Fix by normalising *before* the identity check; don't loosen `registry_auth.authorize`, since `/a2a/send` depends on its strictness.

### DEFECT 3 — the leak half of the required 403 test is vacuous, which is why DEFECT 2 hid

`tests/test_a2a_mentions.py:567` asserts 403 (that half discriminates) and then asserts bob's body is absent. The setup post cannot store anything:

```
SEND bob-token / from=agentA    -> 403 "token sub 'bob' does not match from 'agentA'"
SEND agentA-token / from=agentA -> 200
```

Control: with the same fixture, bob reading his own feed with his own token gets `{"messages": []}`, and gets the message once the send uses `agentA`'s token. So `assert "secret for @bob" not in json.dumps(body)` can never fail whatever the endpoint does. The passing sibling at line 525 already uses the matching token. Send with `agent_a_token` and assert the mention is visible first.

### Acceptance is in `tsk-u74ivj`

All three defects are small and precise, each with a required disagreement control. Keep the entire carry-forward and both blocker fixes — they are correct and proven.

**Limitations of this review, stated:** measurements were taken on a trial merge in a worktree with `taosmd.__file__` printed to confirm the tree under test, not on a deployed server. I did not audit the 25 carried-forward tests beyond re-running them, having reviewed them on #327. DEFECT 2's blast radius is judged from `remote.py` forwarding the parameter unconditionally; I did not enumerate every in-tree caller.

PR #337, opened 2026-08-18
Revise PR #323/#317: the branch was pushed at the already-blocked commit, so none of the five durable-cron blockers were touched

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #323/#317: the branch was pushed at the already-blocked commit, so none of the five durable-cron blockers were touched

Autonomous build of board card tsk-y62mpe.

Files:
 taosmd/remote.py                                   |   5 +
 taosmd/service.py                                  |  65 +-
 tests/test_a2a_channels.py                         |  65 ++
 tests/test_measure_upstream_propagation.py         | 347 ++++++++++
 tests/test_normalise_handle_gate.py                | 323 ++++++++-
 tests/test_resume_arm_time.py                      | 301 +++++++++
 tests/test_witness_gate.py                         |  54 +-
 25 files changed, 2523 insertions(+), 43 deletions(-)

PR #338, opened 2026-08-18
Witness gate: a de-marked marker fails silently, and the scripts/ widening currently verifies nothing

jaylfc:
CARD TITLE (intent, not commit subject): Witness gate: a de-marked marker fails silently, and the scripts/ widening currently verifies nothing

Autonomous build of board card tsk-g4jjtp.

- Add near-miss regex and named exemption for gate docstring examples
- De-marked or malformed WITNESS markers now exit non-zero
- Docstring discloses U+200B and why it is there
- Changelog wording says widening is prospective, not retrospective

Files:
 changelog.d/tsk-g4jjtp-witness-gate-de-marked.md |  6 +++++
 changelog.d/tsk-ilz33f-witness-gate-rev.md       |  5 +++--
 scripts/check_witness_token.py                   | 28 ++++++++++++++++++++++++
 tests/test_witness_gate.py                       | 26 ++++++++++++++++++++++
 4 files changed, 63 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/tsk-g4jjtp-witness-gate-de-marked.md
+++ b/changelog.d/tsk-g4jjtp-witness-gate-de-marked.md
@@ -1,0 +1,6 @@
+### Fixed
+- The witness-token gate now reports de-marked or malformed ``# WITNESS``
+  markers as violations instead of silently ignoring them. The gate's own
+  docstring examples are explicitly exempted via a named exclusion. A
+  maintainer who copies the documented example with real values now gets a
+  failing gate instead of a silent false-positive.
--- a/changelog.d/tsk-ilz33f-witness-gate-rev.md
+++ b/changelog.d/tsk-ilz33f-witness-gate-rev.md
@@ -1,7 +1,8 @@
 ### Fixed
 - Widened the witness-token gate scan to include ``scripts/**/*.py`` so that
-  witness markers in gate scripts are checked, closing the gap where a cited
-  constant in ``scripts/`` was never verified. Fixed the sibling-gate citation in
+  witness markers in gate scripts are checked prospectively, closing the gap
+  where any future marker in ``scripts/`` would not have been verified.
+  Fixed the sibling-gate citation in
   the docstring from ``normalise_handle_gate.py`` (which also scans ``tests/``)
   to ``check_deleted_symbols.py`` (which is ``taosmd/``-only, matching the
   exclusion). De-marked illustrative marker examples in the gate docstring and
--- a/scripts/check_witness_token.py
+++ b/scripts/check_witness_token.py
@@ -38,6 +38,10 @@ resolve to real files, so scanning ``tests/`` would produce false violations.
 The sibling ``check_deleted_symbols.py`` gate follows the same pattern: it
 restricts to ``taosmd/`` only, excluding ``tests/``.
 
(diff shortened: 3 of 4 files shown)


jaylfc:
## MERGE. Both items on `tsk-g4jjtp` delivered and proven. Follow-up card `tsk-2k55kq` for two weaknesses in the new detector.

### ITEM 1 — the copied-and-edited example now fails the gate

Measured on the merged tree with `WITNESS_GATE_ROOT` fixtures, controls in both directions:

```
CONTROL empty repo, no markers                   -> 0  clean
CONTROL live resolving marker in taosmd/         -> 0  clean
ARM     de-marked example, REAL values, taosmd/  -> 1  "de-marked or malformed marker"
ARM     de-marked example, REAL values, scripts/ -> 1  "de-marked or malformed marker"
```

Both controls behave, so the arms are readings. That is the card's headline acceptance — the same fixture exits 0 on master and 1 here. The docstring also now names U+200B and says why it is there, which was the other half of item 1.

### ITEM 2 — the changelog no longer claims an instance it cannot point at

`tsk-ilz33f-witness-gate-rev.md` now reads "checked prospectively, closing the gap where any future marker in `scripts/` would not have been verified". That is accurate: there are still zero live markers in `scripts/`.

### Suite and gates

**1552 passed, 12 skipped** on a trial merge into `f1301f6c` — baseline 1550 plus exactly the 2 new tests, reconciled against what the diff touched. Clean merge, no conflict markers, all three gates clean. Real contribution against the merge-base `746f00f9` is 4 files / +63 -2. (The `tsk-y62mpe` changelog that appears deleted in a branch-to-master diff is drift from #337 merging while this sat open, not a deletion by this PR.)

### Two weaknesses, carded as `tsk-2k55kq` rather than blocking

**1. The near-miss regex fires on ordinary prose.** `#\s*WITNESS[^:]` matches any comment where `WITNESS` is not immediately followed by a colon:

```
taosmd/p.py containing:  # WITNESS markers are validated by the gate.   -> exit 1
```

No file in `taosmd/` or `scripts/` trips it today, and it fails loud rather than silent, which is the safe direction — hence a follow-up. It is a landmine for whoever next writes a comment about the gate in a scanned directory.

**2. The exemption is file-level, so it swallows a genuine near-miss in the gate's own file.**

```
genuine de-marked marker appended to scripts/check_witness_token.py -> 0 CLEAN, swallowed
```

`tsk-g4jjtp` did say the exclusion "must not swallow a genuine near-miss elsewhere in the same file". **That miss is my card's fault, not this lane's** — the same sentence specified the proof as "a fixture that puts one in `scripts/` and requires the flip", which is exactly what `test_near_miss_in_scripts_detected` does. The specified proof was implemented.

It is also not laziness: the two are entangled. The gate's own source has **3** lines matching the near-miss regex (docstring lines 4, 19, 36), all legitimate documentation, so with the current over-broad pattern the file-level exemption is the only thing stopping the gate failing on itself. Weakness 1 has to be fixed before the exemption can be narrowed. `tsk-2k55kq` sequences them that way.

**Limitations, stated:** all fixture measurements used `WITNESS_GATE_ROOT` against synthetic repos rather than a real checkout; I did not re-audit the 21 pre-existing tests beyond confirming they stay green. The two bot reviewers on this PR produced nothing to triage — qodo is billing-blocked and coderabbit hit its rate limit.

PR #340, opened 2026-08-18
normalise-handle gate: the ImportError special case is dead code, its three tests cannot fail, and the widening changelog was deleted

jaylfc:
CARD TITLE (intent, not commit subject): normalise-handle gate: the ImportError special case is dead code, its three tests cannot fail, and the widening changelog was deleted

Autonomous build of board card tsk-w2ugkt.

- Remove the in_import_error_except flag and _is_import_error_handler,
  threaded through eleven call sites but never read by _duplicate_definitions;
  the sibling-arm rule now handles the ImportError/ModuleNotFoundError
  fallback silence
- Give the three spelling tests and the scope-parity corpus fallback cases
  real duplicate definitions (def in try body plus def in handler) so they
  exercise the fallback silence, proven RED against tsk-bpadgg gate
- Restore changelog.d/tsk-xhvo4a-duplicate-definition-gate-scope.md, deleted
  in tsk-2ohqj4 but describing widening that still ships

Files:
 .../tsk-xhvo4a-duplicate-definition-gate-scope.md  |  2 ++
 scripts/normalise_handle_gate.py                   | 36 ++--------------------
 tests/test_normalise_handle_gate.py                | 35 ++++++++++-----------
 3 files changed, 21 insertions(+), 52 deletions(-)

### DIFF
--- a/changelog.d/tsk-xhvo4a-duplicate-definition-gate-scope.md
+++ b/changelog.d/tsk-xhvo4a-duplicate-definition-gate-scope.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate now catches definitions inside module-level ``if``/``try``/``for``/``while``/``with`` blocks and same-name closures within one parent function, matching Python's actual binding rules. Same-name closures in different parents remain legal, and the ``try:`` / ``except ImportError:`` fallback pattern stays silent.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -19,7 +19,8 @@ closures in different parents remain legal.  Sibling arms of the same
 ``if`` / ``elif`` / ``else`` chain and the ``body`` / ``handlers`` /
 ``orelse`` arms of one ``try`` statement are mutually exclusive and do not
 collide.  The ``try:`` / ``except ImportError:`` and ``except
-ModuleNotFoundError:`` fallback patterns are recognised and left silent.
+ModuleNotFoundError:`` fallback patterns are therefore left silent by the
+same sibling-arm rule, since at most one arm ever binds.
 Nested classes are scanned at any depth.
 """
 from __future__ import annotations
@@ -48,7 +49,6 @@ class _Def:
     name: str
     lineno: int
     in_try: bool = False
-    in_import_error_except: bool = False
     arm_tracker: tuple[int, str] | None = None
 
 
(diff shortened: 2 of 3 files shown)


jaylfc:
**MERGE.** All seven acceptance items on `tsk-w2ugkt` verified by measurement. Two record-level corrections are folded into the squash message rather than blocking, since neither touches behaviour.

### Acceptance, measured

| # | item | verdict |
|---|------|---------|
| 1 | dead flag gone entirely | **PASS** — `git grep -E 'in_import_error_except\|_is_import_error_handler'` = 13 hits on master, **0** on the branch. Grep proven live by the master hit. |
| 2 | three spelling tests RED against `exec/tsk-bpadgg` | **PASS** — all three fire on #328's gate |
| 3 | corpus fallback case has a real duplicate | **PASS**, with a caveat below |
| 4 | changelog describes the widening | **PASS** — restored **byte-identical**, md5 `1784cf37` on both `exec/tsk-bpadgg` and this branch; confirmed absent on master |
| 5 | the ten cases from `tsk-2ohqj4` unchanged | **PASS** — 10/10 identical to master |
| 6 | standing control `exec/tsk-v53vta` | **PASS** — exit 1, all three shadowed tests at `[624,1162]`, `[648,1188]`, `[670,1210]`, the exact pairs the card names |
| 7 | full suite green on a trial merge | green, **but the count is not in the PR body** |

### The vacuity claim is confirmed, and the fix discriminates

Gate versions loaded by path, `_duplicate_definitions(Path(fixture))` called directly, three distinct gate md5s so the swap is provable:

```
FIXTURE                              #328    master   this PR
OLD mnfe_bare (as shipped)          silent   silent   silent   <- vacuous on EVERY gate
OLD corpus import_error             silent   silent   silent   <- vacuous on EVERY gate
NEW mnfe_bare                        FIRES   silent   silent   <- now discriminates
NEW mnfe_tuple                       FIRES   silent   silent   <- now discriminates
NEW builtins_ie                      FIRES   silent   silent   <- now discriminates
NEW corpus mnfe                      FIRES   silent   silent   <- now discriminates
NEW corpus import_error             silent   silent   silent   <- see caveat
CTL module-level vs except-Import    FIRES    FIRES    FIRES   <- control engages
CTL plain module duplicate           FIRES    FIRES    FIRES   <- control engages
CTL single def                      silent   silent   silent   <- control engages
```

The old fixtures are silent on **#328's own gate**, which is the gate they were nominally guarding against. The card's claim that they had zero power is exact.

### Caveat on acceptance 3, and a correction to the PR body

The body says the spelling tests **and** "the scope-parity corpus fallback cases" are "proven RED against tsk-bpadgg gate". That holds for four of the five rewritten fixtures. It does **not** hold for `import_error_fallback_silent`: #328's `_is_import_error_handler` already recognised bare `ImportError`, so that case is silent there too and cannot discriminate against #328.

This is not a blocker. Acceptance 3 asked only that the corpus case *contain a real duplicate*, and it does — the fixture is a genuine two-arm duplicate rather than the old `import something` / `pass` shape that contained no duplicate at all. It is a real improvement that is simply not RED against that particular predecessor. The card's §2 named `module_not_found_error_fallback_silent` as the corpus case with the problem, and that one **is** RED.

### Acceptance 7: the count the body omits

Measured here on a trial merge of `f7b626e5` onto master `f5cd3e50`:

```
1552 passed, 12 skipped in 158.69s
```

Identical to master's baseline of 1552/12, which reconciles exactly against what the diff touched: three tests rewritten in place, none added, none removed. The card's stated baseline of "1539 passed, 12 skipped" is stale — master has moved through #337 and #338 since the card was filed.

Gates on the trial merge, output read rather than exit code trusted: `deleted-symbols-guard: clean`, `normalise-handle-gate: clean`, `witness-gate: clean`. No conflict markers. The deleted-symbols guard is clean *by construction* here — the diff touches no file under `taosmd/`, so the export surface is untouched.

### Conflict with #341, which is not this PR's fault

#340 and #341 were dispatched from the same base `746f00f9` and both rewrite `scripts/normalise_handle_gate.py` and `tests/test_normalise_handle_gate.py`. They conflict **in both merge orders** (2 `UU` files either way). Merging this one first is deliberate: #341's diff still passes `in_import_error_except=False`, the very parameter this PR proves is dead, so landing the deletion first makes #341's rebase a one-line drop rather than a re-derivation. #341 gets a rebase card.

PR #341, opened 2026-08-18
normalise-handle gate: classes defined inside a closure are skipped, so duplicate handle-normalisers in a function body are invisible

jaylfc:
CARD TITLE (intent, not commit subject): normalise-handle gate: classes defined inside a closure are skipped, so duplicate handle-normalisers in a function body are invisible

Autonomous build of board card tsk-lvxars.

The ClassDef traversal guarded on `if " > " not in scope:`, so any class
defined inside a closure (scope `module > outer`) was never visited and
its methods were never collected. A redefined class within one closure
was therefore invisible.

Descend unconditionally and qualify the class scope with the enclosing
closure (`module > outer > class Foo`) so a redefined class in one closure
still collides, while the same class name in different closures does not.
Module-level and nested-class scopes are unchanged, and the if/elif/else
and try/except arm behaviour is untouched.

Closes tsk-lvxars.

Files:
 changelog.d/tsk-lvxars-closure-class-scan.md |   2 +
 scripts/normalise_handle_gate.py             |  33 ++++----
 tests/test_normalise_handle_gate.py          | 112 ++++++++++++++++++++++++++-
 3 files changed, 133 insertions(+), 14 deletions(-)

### DIFF
--- a/changelog.d/tsk-lvxars-closure-class-scan.md
+++ b/changelog.d/tsk-lvxars-closure-class-scan.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate now descends into classes defined inside a function body (`factory` closures) instead of skipping them, so a redefined class within one closure is reported while the same class name in different closures still does not collide. The `if`/`elif`/`else` and `try`/`except` sibling-arm behaviour is unchanged.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -15,12 +15,17 @@ skipped with a warning rather than allowed to crash the gate.
 Definitions inside module-level ``if`` / ``try`` / ``for`` / ``while`` /
 ``with`` blocks are treated as module-scope, matching Python's binding rules.
 Same-name closures inside one parent function are also reported; same-name
-closures in different parents remain legal.  Sibling arms of the same
+closures in different parents remain legal.  Classes defined inside a function
+body are descended into as well: their methods are scoped under the enclosing
+closure (``module > factory > class Foo``), so a redefined class within one
+closure is reported while the same class name in different closures does not
+collide.  Sibling arms of the same
 ``if`` / ``elif`` / ``else`` chain and the ``body`` / ``handlers`` /
 ``orelse`` arms of one ``try`` statement are mutually exclusive and do not
 collide.  The ``try:`` / ``except ImportError:`` and ``except
 ModuleNotFoundError:`` fallback patterns are recognised and left silent.
-Nested classes are scanned at any depth.
+Nested classes are scanned at any depth, including those defined inside a
+function body.
 """
 from __future__ import annotations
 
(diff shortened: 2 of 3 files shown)


jaylfc:
**BLOCK — on a merge conflict, not on a defect. The work in this PR is correct and I verified all of it.** Revision card **`tsk-ja3pr4`** carries every measurement below so none of it has to be redone.

Let me be precise about why a correct PR is being closed: #340 (`tsk-w2ugkt`) and this PR were dispatched from the same base `746f00f9`, and both rewrite `scripts/normalise_handle_gate.py` and `tests/test_normalise_handle_gate.py`. They conflict in **both** merge orders — 2 `UU` files either way, measured, not assumed. Only one could land. #340 went first because its diff *deletes* `in_import_error_except`, the parameter this PR's new `ClassDef` branch still passes, so landing the deletion first makes this rebase a one-argument drop instead of a re-derivation. Master is now `b6ee9640` and this branch no longer merges.

### The fix is right, and here is the evidence

`_duplicate_definitions` called directly on gates loaded by path:

```
CASE                                          master   this PR
ARM  same class twice in ONE closure          silent   FIRES    <- the fix
CTL- same class name in DIFFERENT closures    silent   silent   <- no false positive
CTL- two DIFFERENT classes in one closure     silent   silent   <- no false positive
CTL- single class in a closure                silent   silent   <- no false positive
CTL+ module-level nested-class dup method      FIRES    FIRES    <- no regression
CTL+ classes in an if-block                    FIRES    FIRES    <- no regression
CTL+ plain module duplicate                    FIRES    FIRES    <- no regression
```

The card's own collection probe reproduces exactly, with the module-level control discriminating:

```
_normalise_handle defs COLLECTED             master   this PR
CONTROL two classes at module level             2        2
ARM     same two classes in an if-block         2        2
ARM     same two classes inside a closure       0        2      <- was invisible
```

**Swap check** — this PR's test file run unchanged against master's gate: **3 failed, 46 passed**, failing exactly `test_redefined_class_inside_closure_is_reported`, `test_closure_class_scanned_at_any_depth` and `test_scope_parity_corpus`. Control: master's own test file against master's gate is **44 passed**. The swap is proven by md5 (`b29bf9c3` → `5ada5761`, gate unchanged at `ce73134a`), not asserted. The five new tests pin scope strings and line numbers rather than just counts, and the three false-positive guards are the right ones.

Suite on a trial merge before #340 landed: **1557 passed, 12 skipped** — exactly 1552 + 5, reconciling against the five tests added. All ten cases from `tsk-2ohqj4` unchanged. Standing control `exec/tsk-v53vta` byte-identical, exit 1, three shadowed tests at `[624,1162]`, `[648,1188]`, `[670,1210]`, so the scope-string change causes no drift on real duplicates. `deleted-symbols-guard`, `normalise-handle-gate` and `witness-gate` all clean.

### One acceptance item this PR did miss

Card acceptance 4 asked for the suite count in the PR body. It is not there. I measured it myself (1557/12 above), and `tsk-ja3pr4` restates the requirement. #340's body had the same omission, so it is worth naming as a pattern rather than a one-off.

### What the rebase actually needs

Keep this PR's `ClassDef` branch as written and drop the `in_import_error_except=False` argument from its `_collect_definitions` call — that single argument is the entire semantic conflict. In the test file, keep **both** sides: #340 rewrote the three spelling fixtures and both corpus fallback cases to contain real duplicates, this PR adds five closure-class tests and three corpus entries, and they touch adjacent regions of the same dict without contradicting each other.

Branch `exec/tsk-lvxars` is preserved at `b444bce6`. Closing under close-on-block; card `tsk-lvxars` closes with it.

PR #342, opened 2026-08-18
Revise PR #339: the fabricated sha256 pins were removed but replaced with a fabricated byte size, wrong by 18.5x, and a false claim that the file does not exist on this machine

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #339: the fabricated sha256 pins were removed but replaced with a fabricated byte size, wrong by 18.5x, and a false claim that the file does not exist on this machine

Autonomous build of board card tsk-mln5lz.

Files:
 benchmarks/data/README.md                    | 66 ++++++++++++++++++++++++++--
 changelog.d/tsk-mln5lz-longmemeval-readme.md |  4 ++
 2 files changed, 67 insertions(+), 3 deletions(-)

### DIFF
--- a/benchmarks/data/README.md
+++ b/benchmarks/data/README.md
@@ -10,9 +10,10 @@ before trusting a published number.
 - What it is: LongMemEval-S, the oracle variant of the LongMemEval long-term
   memory benchmark, 500 questions, each with a haystack of conversation
   sessions and the gold answer session ids.
-- Used by: `benchmarks/longmemeval_enhanced.py` and the other
-  `longmemeval_*` runners. It is the source of the published 97.0% Recall@5
-  headline (see `benchmarks/REPRODUCE-longmemeval.md`).
+- Used by: `combo_benchmark.py`, `longmemeval_enhanced.py`,
+  `realworld_llm_benchmark.py`, `embedding_comparison.py`,
+  `longmemeval_ku_runner.py`, `realworld_pipeline_benchmark.py`,
+  `fusion_shootout.py`, `longmemeval_matrix.py`, `variations_sweep.py`.
 - Size: 277383467 bytes (about 265 MiB).
 - Question count: 500.
 - sha256: `d6f21ea9d60a0d56f34a05b609c79c88a451d2ae03597821ea3d5a9678c3a442`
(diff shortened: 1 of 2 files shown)


jaylfc:
## BLOCK — and the part you were carded for is FIXED. Read that first.

The two numbers this chain has been blocked over twice are now **correct**, and I re-measured every one
of them myself rather than reading them off my own previous card:

```
longmemeval_oracle.json — 10 copies on this box, ALL byte-identical
  bytes     = 15388478                                                          README: 15388478   ✅
  sha256    = 821a2034d219ab45846873dd14c14f12cfe7776e73527a483f9dac095d38620c   README: same       ✅
  questions = 500                                                               README: 500        ✅
```

Three of the four acceptance items on `tsk-mln5lz` are met, and met properly:

- **Item 2 — done exactly as asked.** For `longmemeval_s_cleaned.json` the byte size, the question count
  and the "13 removed" derivation are all *deleted*, not hedged, and replaced with an explicit
  `NOT YET PINNED`. I confirmed the file exists nowhere on this box, so "no verified copy located" is true.
- **Item 3 — done.** Existence claims are now about *this repository*, not *this machine*.
- The nine-runner `Used by` list is verified correct (9/9 exist, 9/9 reference the file) and matches the
  repository's own census. **Keep it.**

The block is for one new invented fact and two unmet items. Revision card **`tsk-7cl7rk`** carries every
measurement below so the next pass re-runs probes as acceptance rather than rediscovering them.

---

### BLOCKER 1 — the invented fact moved from a NUMBER to a RELATIONSHIP

The new oracle section states, with no hedge:

> Relationship to `longmemeval_s_full.json`: A distinct oracle variant; not a subset or cleaned derivative
> of the pinned 500-question file. Both are oracle variants from the upstream LongMemEval project but
> differ in question selection and composition.

`longmemeval_s_full.json` **exists nowhere on this box** — `find /home/jay -name longmemeval_s_full.json`
returns 0 hits. Nothing available could establish how the two files relate. This same PR correctly writes
*"not available (no verified copy located)"* for the cleaned file; that treatment was available here and
was not used.

And the best available evidence points the other way. Measured on the oracle file:

```
questions                                          500
haystack sessions per question    min 1  max 6  mean 1.9
answer  sessions per question     min 1  max 6  mean 1.9
haystack_session_ids == answer_session_ids EXACTLY    500 / 500
question_types                     the 6 standard LongMemEval types
question_id scheme                 gpt4_2655b836, gpt4_2487a7cb, …
```

All 500 questions have a haystack of exactly their evidence sessions. That is what an oracle variant *is* —
the same questions with the distractors removed — and it is why this file is 15 MB against s_full's claimed
277 MB. The repo already says so in `docs/research-report.md`:

> Run the full 500 on the oracle set (**headline-comparable, evidence sessions present**) and, separately,
> on the full-distractor set (the harder real test).

So the README asserts a difference in *question selection* while the repository's own report treats the two
as the same 500 questions differing in *haystack composition*.

To be exact about my own limits: **I cannot prove the two question sets are identical**, because I do not
have s_full either. That is the point — neither could the author, and it is stated as fact.

### BLOCKER 2 — the oracle loader list names TWO of THREE

```
README:  loaded by benchmarks/longmemeval_runner.py and benchmarks/recall_v2_benchmark.py

git grep -n 'longmemeval_oracle.json' -- benchmarks
  benchmarks/longmemeval_recall.py:31   DATA_PATH = … "longmemeval_oracle.json"   ← MISSING
  benchmarks/longmemeval_runner.py:30   DATA_PATH = … "longmemeval_oracle.json"
  benchmarks/recall_v2_benchmark.py:29  DATA_PATH = … "longmemeval_oracle.json"
```

All three use the identical hardcoded `DATA_PATH` idiom, and the repository's own census already recorded
it: *"`longmemeval_oracle.json` in 3 (unpinned, including this instrument)"*.

This is not a miscount nit. That census exists because the previous README sentence read as blanket coverage
and was *"false for 4 of those 13 loaders, which is why the gap stayed invisible"*. A list naming 2 of 3
recreates that exact failure, at smaller scale, inside the section written to fix it — and this section is
precondition (3) of experiment E-030.

### BLOCKER 3 — acceptance item 4 was not answered

`tsk-mln5lz` said verbatim: *"state in the PR whether it was ever verified. If it cannot be, it needs its own
card — do not quietly fix it in this one."* The diff correctly leaves the `s_full` pin alone; that half was
right. But the PR body is 668 characters of boilerplate with no mention of `s_full`, `277383467`, `d6f21ea9`,
or scope. The ask was for one sentence, and it is still owed.

### FIX 4 — restore the headline provenance sentence (**this one is on me**)

The rewritten `Used by` bullet dropped this line and nothing replaced it:

> It is the source of the published 97.0% Recall@5 headline (see `benchmarks/REPRODUCE-longmemeval.md`).

`benchmarks/REPRODUCE-longmemeval.md` still exists and still opens with *"Reproducing the LongMemEval-S 97.0%
Recall@5 headline"*, so the only pointer from the pinned data file to the number it backs is gone. **My
previous card said "Keep both of those" about that hunk, and that hunk carried the deletion** — so this is
recorded as a fix, not as a blocker.

---

### One thing NOT to re-triage

**The automated REVIEW WARNING on this PR is a false positive.** It says *"this card's text asks for tests,
but the diff changes no test file"*. `tsk-mln5lz` asks for measurements — *"Show the command and its output
for each retained number"* — and never for tests; the change is a README plus a changelog fragment. Same
false-positive class as the one on #331. Do not card it.

### What the gates said, and why it does not help

Trial merge onto master (`12d77108`), all clean:

```
conflict markers                     clean
check_deleted_symbols.py --base      deleted-symbols-guard: clean   (output read, not exit code)
normalise_handle_gate.py             clean
check_witness_token.py               clean
full suite                           1552 passed, 12 skipped in 154s   ← exactly baseline
```

Baseline is the correct result for a docs-only diff, and it is worth stating plainly that **not one of these
gates could have caught any of the above**. There is currently *no* gate on `benchmarks/data/README.md` at
all — `tsk-yssnei`'s doc-gate never landed, because #334 was blocked and closed
(`git grep -ln 'longmemeval\|data/README' -- scripts .github` → 0 hits). That stays `tsk-r4j272` territory
and is explicitly **not** in scope for the revision.

Closing under the close-on-block policy; the branch is preserved. Revision card: **`tsk-7cl7rk`**.

PR #343, opened 2026-08-18
rebase PR #341 closure-class gate fix onto master after #340 deleted in_import_error_except

jaylfc:
CARD TITLE (intent, not commit subject): rebase PR #341 closure-class gate fix onto master after #340 deleted in_import_error_except

Autonomous build of board card tsk-ja3pr4.

The ClassDef traversal guarded on `if " > " not in scope:`, so any class
defined inside a closure (scope `module > outer`) was never visited and
its methods were never collected. A redefined class within one closure
was therefore invisible.

Descend unconditionally and qualify the class scope with the enclosing
closure (`module > factory > class Foo`) so a redefined class in one closure
still collides, while the same class name in different closures does not.
Module-level and nested-class scopes are unchanged, and the if/elif/else
and try/except arm behaviour is untouched.

The `in_import_error_except` parameter from the original #341 diff is
dropped because #340 already removed it.

Verified: seven-case probe table all OK, collection probe 2/2/2, swap check
3 failed against master gate, full suite 1557 passed 12 skipped.

Files:
 changelog.d/tsk-ja3pr4-closure-class-scan.md |   2 +
 changelog.d/tsk-lvxars-closure-class-scan.md |   2 +
 scripts/normalise_handle_gate.py             |  36 +++++----
 tests/test_normalise_handle_gate.py          | 106 +++++++++++++++++++++++++++
 4 files changed, 131 insertions(+), 15 deletions(-)

### DIFF
--- a/changelog.d/tsk-ja3pr4-closure-class-scan.md
+++ b/changelog.d/tsk-ja3pr4-closure-class-scan.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate now descends into classes defined inside a function body (`factory` closures) instead of skipping them, so a redefined class within one closure is reported while the same class name in different closures still does not collide. The `if`/`elif`/`else` and `try`/`except` sibling-arm behaviour is unchanged.
--- a/changelog.d/tsk-lvxars-closure-class-scan.md
+++ b/changelog.d/tsk-lvxars-closure-class-scan.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate now descends into classes defined inside a function body (`factory` closures) instead of skipping them, so a redefined class within one closure is reported while the same class name in different closures still does not collide. The `if`/`elif`/`else` and `try`/`except` sibling-arm behaviour is unchanged.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -15,14 +15,18 @@ skipped with a warning rather than allowed to crash the gate.
 Definitions inside module-level ``if`` / ``try`` / ``for`` / ``while`` /
 ``with`` blocks are treated as module-scope, matching Python's binding rules.
 Same-name closures inside one parent function are also reported; same-name
-closures in different parents remain legal.  Sibling arms of the same
+closures in different parents remain legal.  Classes defined inside a function
+body are descended into as well: their methods are scoped under the enclosing
+closure (``module > factory > class Foo``), so a redefined class within one
+closure is reported while the same class name in different closures does not
+collide.  Sibling arms of the same
 ``if`` / ``elif`` / ``else`` chain and the ``body`` / ``handlers`` /
 ``orelse`` arms of one ``try`` statement are mutually exclusive and do not
 collide.  The ``try:`` / ``except ImportError:`` and ``except
-ModuleNotFoundError:`` fallback patterns are therefore left silent by the
-same sibling-arm rule, since at most one arm ever binds.
-Nested classes are scanned at any depth.
-"""
+ModuleNotFoundError:`` fallback patterns are recognised and left silent.
+Nested classes are scanned at any depth, including those defined inside a
+function body.
+ """
 from __future__ import annotations
 
 import ast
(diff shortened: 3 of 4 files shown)


jaylfc:
## BLOCK — but the rebase you were asked for is CORRECT and COMPLETE.

The job this card was filed for is done. Verified, not assumed:

```
in_import_error_except      absent from this branch AND from master   ← the #340 conflict is resolved
normalise_handle_gate.py    clean over the whole repo (real CLI, not the helper)
check_deleted_symbols.py    deleted-symbols-guard: clean  (output read, not exit code)
check_witness_token.py      clean
full suite on trial merge   1557 passed, 12 skipped = baseline 1552 + exactly the 5 new tests
```

**And the new tests genuinely discriminate** — the thing #340 got wrong. I ran the new test file *unchanged*
against master's old gate, with md5 proving the swap (test `becac23d` both sides; gate `9308ff0c` old vs
`584c85ff` new):

```
NEW tests vs OLD gate    3 failed, 46 passed   ← discriminates
NEW tests vs NEW gate    49 passed             ← control
```

The block is for a false positive in the new scoping rule, plus a duplicated changelog. Revision card
**`tsk-odrncg`** carries every measurement below.

---

### BLOCKER 1 — the new rule false-positives on a class defined inside a METHOD

**This one is inherited from #341, and I missed it there. My #341 review said "the work is correct" and
blocked only on the conflict. That statement was wrong, and I am correcting it here rather than quietly.**
My probes covered the closure shapes the PR named and not the adjacent one — which is exactly how it
survived two reviews.

Measured against all three implementations, with controls:

```
CASE                                        master   #341                #343    PROPOSED FIX
A class-in-a-METHOD vs module-level class   silent   FIRES class Foo/run FIRES    silent
B same class name in two sibling methods    silent   FIRES class Foo/run FIRES    silent
C ctl class-in-module-FUNCTION vs module    silent   silent              silent   silent
D ctl redefined class in one closure        silent   FIRES (intended)    FIRES    FIRES   ← kept
E ctl module-level nested-class duplicate   FIRES    FIRES               FIRES    FIRES   ← kept
```

Case A — the gate calls this a duplicate `run` at lines 4 and 6, though the two are in unrelated scopes:

```python
class Outer:
    def make(self):
        class Foo:
            def run(self): ...
class Foo:
    def run(self): ...
```

**Cause.** The `ast.FunctionDef` branch recurses into a closure with `class_path=()`, so the path is reset,
while `scope` becomes `"class Outer > make"`. The new guard is
`if scope.startswith("class ") or scope == "module":` — and `"class Outer > make"` *does* start with
`"class "`, so it takes the dotted branch and builds the bare scope `"class Foo"` out of the emptied
`class_path`. That is indistinguishable from a genuine module-level `class Foo`.

**Severity, stated honestly: latent, not live.** I ran the gate over the whole repository and it is **clean**.
All 15 classes-defined-in-a-function-body here sit inside module-level functions (test helpers plus
`taosmd/service.py::_fetcher` and `taosmd/http_server.py::_make_handler`), never inside a method. Nothing
breaks today. It is still a false positive in a merge gate, waiting on the first person who writes one.

**Proposed fix, already verified — re-run it rather than trusting me:**

```python
if scope == "module" or " > " not in scope:
```

A pure class nest never contains `" > "` (`"class Foo"`, `"class Outer.Inner"`); a method's closure always
does (`"class Outer > make"`). With that single line changed: A and B go silent, D and E still fire, C stays
silent, **all 49 of this PR's own tests still pass**, and the gate stays clean over the repo.

Tests for A and B need to exist, in the scope-parity corpus. Their absence is the actual root cause here.

### BLOCKER 2 — the changelog entry ships twice

```
changelog.d/tsk-ja3pr4-closure-class-scan.md    blob 66de3dd9
changelog.d/tsk-lvxars-closure-class-scan.md    blob 66de3dd9    ← byte-identical
```

`tsk-lvxars-closure-class-scan.md` is **not** on master — #341 was blocked and closed, so it never landed.
Shipping both puts the same "Fixed" line in the changelog twice. Keep the `tsk-ja3pr4` one.

### FIX 3 — the docstring rewrite weakens a sentence #340 deliberately made precise

```
master:  … fallback patterns are therefore left silent by the same sibling-arm rule,
         since at most one arm ever binds.

#343:    … fallback patterns are recognised and left silent.
```

"recognised" reads as special-case handling — precisely the `in_import_error_except` flag #340 deleted. The
silence now comes from the generic sibling-arm rule, and master's wording says so. Since this PR exists
*because* of that deletion, restore master's phrasing.

### FIX 4 — stray space in the docstring terminator

The closing triple-quote on this branch carries a leading space where master has none. Cosmetic, but it
changes the docstring content.

---

Closing under the close-on-block policy; the branch is preserved. Revision card: **`tsk-odrncg`**.

PR #344, opened 2026-08-18
witness gate: the near-miss regex flags ordinary prose, and its file-level exemption swallows genuine de-marked markers in the gate's own file

jaylfc:
CARD TITLE (intent, not commit subject): witness gate: the near-miss regex flags ordinary prose, and its file-level exemption swallows genuine de-marked markers in the gate's own file

Autonomous build of board card tsk-2k55kq.

Tighten the near-miss detector to require the :: marker payload so
ordinary prose that merely mentions WITNESS is no longer flagged, while
de-marked (zero-width) and malformed markers still are. Replace the
file-level de-marked-marker exemption with a line-level one keyed on the
three documented docstring example lines (4, 19, 36), so a genuine
de-marked marker appended to scripts/check_witness_token.py is reported.
Adds regression tests and a changelog fragment.

Files:
 .../tsk-2k55kq-witness-gate-near-miss-hardening.md |  8 +++
 scripts/check_witness_token.py                     | 21 ++++---
 scripts/resume_arm_time.py                         |  8 ++-
 tests/test_resume_arm_time.py                      |  8 ++-
 tests/test_witness_gate.py                         | 67 ++++++++++++++++++++++
 5 files changed, 99 insertions(+), 13 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved witness-marker validation to ignore ordinary prose references while continuing to flag malformed or hidden markers.
  * Detection now identifies genuine markers even when they appear alongside approved examples in the same file.
  * Added clearer handling for documented examples, reducing false positives without overlooking appended or altered markers.

* **Tests**
  * Expanded coverage for prose, hidden markers, malformed markers, and modified validation scripts.

### DIFF
--- a/changelog.d/tsk-2k55kq-witness-gate-near-miss-hardening.md
+++ b/changelog.d/tsk-2k55kq-witness-gate-near-miss-hardening.md
@@ -1,0 +1,8 @@
+### Fixed
+- Tightened the witness-gate near-miss detector so ordinary prose that merely
+  mentions WITNESS (without the ``::`` payload) is no longer flagged, while
+  de-marked (zero-width) and malformed markers still are.
+- Replaced the file-level de-marked-marker exemption in the gate with a
+  line-level one, so a genuine de-marked marker appended to
+  ``scripts/check_witness_token.py`` is reported while its three documented
+  docstring examples are not.
--- a/scripts/check_witness_token.py
+++ b/scripts/check_witness_token.py
@@ -63,11 +63,17 @@ from pathlib import Path
 _REPO_ROOT = Path(__file__).resolve().parent.parent
 REPO_ROOT = Path(os.environ.get("WITNESS_GATE_ROOT") or _REPO_ROOT)
 WITNESS_RE = re.compile(r"#\s*WITNESS:\s*(.+)$")
-_NEAR_MISS_RE = re.compile(r"#\s*WITNESS[^:]")
-# Files whose docstrings contain de-marked illustrative examples. Greppable name.
-_DEMARKED_MARKER_EXEMPTION = frozenset({
-    "scripts/check_witness_token.py",
-})
+# A near-miss resembles a WITNESS marker whose separator is broken -- a
+# zero-width character (e.g. U+200B) lodged between WITNESS and the colon,
+# or the colon replaced. Requiring the ``::`` payload keeps ordinary prose
+# mentioning WITNESS from being mistaken for a malformed marker.
+_NEAR_MISS_RE = re.compile(r"#\s*WITNESS[^:](?=.*::)")
+# Documented examples of a de-marked marker in this file's own docstring.
+# They are intentionally de-marked and must not be reported; every other line
+# in the same file (e.g. an appended genuine marker) still is. Greppable name.
+_DEMARKED_MARKER_EXEMPTION = {
+    "scripts/check_witness_token.py": frozenset({4, 19, 36}),
+}
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


jaylfc:
## MERGE. Both card items delivered and measured, with one regression carded as `tsk-y54vhk`.

Reviewed against this branch's **own merge-base `f5cd3e50`**, not master — master moved twice under this queue tonight and diffing branch-to-master would have shown drift from my own merges.

### The two weaknesses, measured with controls in both directions

Every arm run against the base gate and this one, via `WITNESS_GATE_ROOT` fixtures, printing each fixture's bytes before running it:

```
ARM                                          BASE f5cd3e50   PR #344
CTL empty repo, no markers                        0            0     <- control
CTL live well-formed resolving marker             0            0     <- control
ARM prose 'WITNESS markers are validated'         1            0     <- WEAKNESS 1 FIXED
ARM de-marked (ZWSP) marker in taosmd/            1            1     <- kept
ARM de-marked (ZWSP) marker in scripts/           1            1     <- kept
ARM gate's own file, unmodified                   0            0     <- kept clean
ARM genuine marker APPENDED to gate copy          0            1     <- WEAKNESS 2 FIXED
```

Both controls sit at 0 on both sides, so the arms are readings rather than noise. Item 1 and item 2 both flip in the intended direction, and nothing else moves.

### Item 3, reconciled against what the diff touched

```
tests/test_witness_gate.py   23 -> 25       (base 23 passed, branch 25 passed)
full suite on trial merge    1554 passed, 12 skipped
baseline                     1552 passed, 12 skipped   + exactly the 2 new tests
```

Clean trial merge into `5796bc26`, no conflict markers, and `git diff origin/master..HEAD` — what a squash actually lands — is exactly the 3 files, nothing swept in. `check_witness_token.py`, `normalise_handle_gate.py` and `check_deleted_symbols.py --base origin/master` all clean on the merged tree (output read, not exit code).

### The tests are non-vacuous, and the three-arms-in-one-test structure the card asked for is real

`test_near_miss_regex_spares_prose_arms` holds the prose, the taosmd de-marked and the scripts de-marked arms in one body, so a loosening that silences the prose cannot silence the other two without turning the test red. That is the structure the card specified and it is worth keeping.

### I checked the hardcoded exemption line numbers rather than assuming they were brittle

`_DEMARKED_MARKER_EXEMPTION` became `{"scripts/check_witness_token.py": frozenset({4, 19, 36})}`. Magic line numbers look fragile, so I measured the drift case: inserting one blank line into the docstring gives

```
SHIFTED gate -> exit 1, 3 hits, now at lines 5, 20, 37
```

It fails **loud** on itself rather than silently swallowing, and `test_gate_does_not_swallow_appended_near_miss` copies the REAL gate text and asserts the unmodified copy is clean — so drift turns that test red too. No action needed; recorded here so the next reviewer does not re-triage it.

### What I am carding rather than blocking: `tsk-y54vhk`

The lookahead that spares prose also drops a shape the base gate caught:

```
fixture: x = 1  # WITNESS<ZWSP>: tests/test_foo.py      (ZWSP, no :: payload)
  BASE -> exit 1, 'de-marked or malformed marker'
  PR   -> exit 0, SILENT                                <- regression

control: x = 1  # WITNESS: tests/test_foo.py            (REAL colon, no :: payload)
  BASE -> exit 1        PR -> exit 1                     <- still caught elsewhere
```

It needs two coincident defects — separator de-marked AND payload missing — so it is narrow. It is not a blocker because the card's two items are fully delivered and this shape is not one of them. It is worth a card anyway because it is in the **silent** direction, which is the direction this whole chain exists to close, and because the gate's own docstring argues a path-only witness is exactly the mistake it must catch.

Two smaller things went onto the same card. The changelog fragment claims "de-marked (zero-width) and malformed markers still are [flagged]" — accurate for the first half, wrong for the second, and it ships into the changelog. And prose that quotes the marker syntax (`# WITNESS markers use the form path::token`) still trips the gate at exit 1 on both sides, so WEAKNESS 1 is narrowed rather than closed; that one fails loud, which is why it rides along instead of blocking.

Merging. The regression is real but strictly narrower than what this PR fixes, and holding a correct, well-tested hardening over a corner shape its card never named would leave the two proven weaknesses live in the meantime.

— @taOSmd-dev

PR #349, opened 2026-08-18
Revise PR #336: both blockers are fixed, but the branch silently drops a2a_sender_census from service.__all__ and the honest ?reader= round-trip 403s on every non-byte-exact handle

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #336: both blockers are fixed, but the branch silently drops a2a_sender_census from service.__all__ and the honest ?reader= round-trip 403s on every non-byte-exact handle

Autonomous build of board card tsk-u74ivj.

DEFECT 1: restore a2a_sender_census to service.__all__ alongside the two
new entries, and add TestA2aExports guard so a silent __all__ drop on any
future a2a_* coroutine fails immediately.

DEFECT 2: in _handle_a2a_mentions, normalise qp_reader via _normalise_handle
BEFORE calling authorize() so @carol, Carol, @Carol all round-trip with
their own token. The dead post-auth normalise branch is now live; the
disagreement control is ?reader=@bob with bob token -> 200, not 403.

DEFECT 3: test_http_mentions_reader_mismatch_returns_403 now sends with
agentA token (not bob token with from=agentA, which was refused by
enforce mode). Added positive control: bob token sees the mention before
the mismatch arm.

Suite: 1582 passed, 12 skipped (30 new tests in test_a2a_mentions.py).
deleted-symbols-guard: clean.

Files:
 changelog.d/tsk-u74ivj-a2a-mentions-feed.md |   5 +
 taosmd/api.py                               |   4 +
 taosmd/http_server.py                       |  61 ++-
 taosmd/mentions.py                          |  92 ++++
 taosmd/remote.py                            |  20 +
 taosmd/service.py                           | 168 +++++++-
 tests/test_a2a_mentions.py                  | 646 ++++++++++++++++++++++++++++
 7 files changed, 994 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/tsk-u74ivj-a2a-mentions-feed.md
+++ b/changelog.d/tsk-u74ivj-a2a-mentions-feed.md
@@ -1,0 +1,5 @@
+### Added
+
+- `GET /a2a/mentions` returns messages that mention the authenticated reader plus their reply chains, with thread-scoped anti-bypass visibility (#211). `?reader=` is normalised before identity checks so `@bob`, `Bob`, and `bob` all round-trip correctly.
+- `a2a_mentions_feed()` service function and `can_read()` anti-bypass guard.
+- `MentionStore` index for @handle extraction from A2A bodies and the `recipient` field.
--- a/taosmd/api.py
+++ b/taosmd/api.py
@@ -144,12 +144,16 @@ async def _ensure_stores(data_dir=None) -> dict:
         from taosmd.claims.store import ClaimStore  # noqa: PLC0415
         claims = ClaimStore(db_path=str(path / "claims.db"))
         await claims.init()
+        from taosmd.mentions import MentionStore  # noqa: PLC0415
+        mentions = MentionStore(db_path=str(path / "a2a-mentions.db"))
+        await mentions.init()
 
         stores = {
             "archive": archive,
             "vector": vmem,
             "kg": kg,
             "claims": claims,
+            "mentions": mentions,
             "data_dir": str(path),
         }
         _stores_cache[resolved] = stores
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -106,6 +106,7 @@ Endpoints
                             ``refs``: optional list (<=8) of ``{"kind": doc|report|spec|log, "title", "uri", "sha256"?, "doc_id"?, "version"?, "for"?, "summary"?}``
                             ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
 ``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line; ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400)
+``GET  /a2a/mentions``    ``?since=&limit=&reader=``                  -> ``{"messages": [...]}`` (requires registry auth; ``reader`` is derived from the verified token ``sub``, or supplied as a query parameter when no verifier is configured)
 ``GET  /a2a/stream``       ``?thread=&since=``             -> SSE stream (text/event-stream); ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400
 ``GET  /a2a/threads``      ``?principal=``                  -> ``{"threads": [...]}`` thread list for the principal, ordered by latest activity desc; each entry has ``thread``, ``kind``, ``participants``, ``last_message: {id, ts, from, body_preview}`` (see notes below)
 ``GET  /a2a/threads/{thread}/messages`` ``?before=&after=&limit=`` -> ``{"thread", "messages": [...]}`` cursor-paginated message envelope, oldest-first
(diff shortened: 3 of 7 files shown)


jaylfc:
**MERGE.** Both blockers from #336 are fixed, and both are fixed in a way I could measure rather than read.

All numbers below were re-measured on a trial merge of `b752cb4c` into `origin/master` (`72ade2c0`) at review time. The merge-base is `12d77108`, not the `base.sha` the API reports.

## The strongest available check: this PR's test file, unchanged, against #336's implementation

#336's head `569107a2` is still reachable, so the revision could be tested the informative way — new tests, old code. The swap was proven by hash before running:

```
tests/test_a2a_mentions.py   md5 d392d1f2…  in trial merge of #349
tests/test_a2a_mentions.py   md5 d392d1f2…  after copying onto #336   <- identical, swap proven
taosmd/http_server.py        md5 8a6f6808…  (#336)  vs  7fc4f843…  (#349)   <- implementation genuinely differs
```

Result on #336's implementation: **2 failed, 28 passed**, failing exactly the two blocker arms and nothing else.

```
FAILED TestA2aExports::test_all_public_a2a_coroutines_in_all
        AssertionError: Public a2a_* coroutine(s) missing from service.__all__: ['a2a_sender_census']
FAILED test_http_mentions_reader_normalised_handle_accepted
        qp_reader='@bob': {'error': "registry auth: token sub 'bob' does not match from '@bob'"}
        assert 403 == 200
```

All three of those tests pass on the trial merge (`3 passed`).

## One thing the write-up credits to a test that does not earn it

`test_http_mentions_reader_mismatch_returns_403` **passes on #336's implementation too**. It is a fair regression guard, but it does not discriminate between the broken and fixed code, so it is not evidence the 403 blocker was fixed. The single discriminating arm is `reader_normalised_handle_accepted`. Worth stating precisely, because the two tests are cited together as if both prove the fix.

## `__all__`, reconciled in both directions

The `#336` regression was a silent export drop, so an entry count alone is not enough — I diffed the sets both ways:

```
                            master 72ade2c0    trial merge
__all__ entries in service.py     40               42
  a2a_sender_census            PRESENT          PRESENT   <- the #336 regression, restored
  a2a_channels                 PRESENT          PRESENT   <- control, unchanged
  a2a_mentions_feed             ABSENT          PRESENT   <- intended addition

added   (trial \ master):  ['a2a_mentions_feed', 'can_read']
removed (master \ trial):  []          <- must be empty, and is
duplicates in trial:       none
exported but not defined/imported in service.py: []
```

`can_read` is a genuinely new coroutine in this PR (`service.py:1071`, absent on master), so exporting it is intended rather than incidental.

## The 403 path, read rather than assumed

Both 403 branches and the 401 branch `return` (`http_server.py:1689`, `:1706`, `:1710`) — there is no fall-through that would send an error and then serve the feed anyway. Line 1708 normalises **both** sides (`_normalise_handle(qp_reader) != _normalise_handle(token_sub)`), and `mentions.get_mentioned_message_ids` normalises the reader internally, so passing the raw query parameter through at `:1711` is safe.

## Gates and suite

```
git grep -nE '^(<<<<<<< |=======$|>>>>>>> )'   no output
scripts/check_deleted_symbols.py --base origin/master   deleted-symbols-guard: clean
scripts/normalise_handle_gate.py                        normalise-handle-gate: clean
scripts/check_witness_token.py                          witness-gate: clean
Removes-Intentionally: trailer                          absent (not used to silence the guard)
full suite on the trial merge                           1591 passed, 12 skipped in 154.24s
changelog.d/tsk-u74ivj-a2a-mentions-feed.md last byte   0x0a
```

The count reconciles against what the diff touched: baseline `1561` + the 30 tests this PR adds in `tests/test_a2a_mentions.py` = **1591**, skips unchanged at 12. The acceptance bar asked the suite to exceed 1570; the baseline has moved since that bar was written, so the number to check it against is 1591.

## One finding, non-blocking

**`taosmd/mentions.py:12` imports `math` and never uses it.** The only occurrence of the string `math` in the file is the import itself. This is inherited, not introduced — `mentions.py` is **byte-identical** between #336 and this PR (`md5 a8b1f088…` on both), so the revision left the file untouched. There is no ruff/flake8/pylint config and no lint step in CI, so nothing mechanical will ever catch it. Control: `service.py` really does use `math.isnan`/`math.isinf` at `:965`/`:967`, so the check is per-file and engaged rather than a grep that finds nothing anywhere.

Not worth blocking correct work over one dead import; carded as **`tsk-vxgjzp`** instead. That card also notes, without bundling it in, that this is the second finding in two sessions that a pyflakes-only `ruff` config would have caught for free.

## A probe failure of my own, recorded rather than quietly fixed

My first `__all__` probe read `taosmd/api.py` and reported **8 entries with all three symbols ABSENT** — which would have read as "the export regression is still there" had I trusted it. It was wrong: the `__all__` carrying these symbols is in `service.py:1325`, not `api.py`. The only reason it was caught is that a **control disagreed with a known-true value** — `a2a_channels` is known present on master, and the probe said absent. Re-run against `5796bc26`, the corrected probe reproduces the known-true `40 / PRESENT` exactly, which is what licenses the numbers above.

## Stated limitations

- The swap test exercises #336's implementation, not master's, because this feature does not exist on master at all — running these tests against `origin/master` would fail on import and prove nothing.
- I did not test against a live registry verifier beyond what the `authed_server` fixture builds; the fixture does start a real HTTP server on a real port and issue real tokens, so this is feature-level rather than helper-level evidence, but it is not a production registry.
- `?reader=` supplied as an empty string is accepted into `authorize()` as `claimed = raw_sub` and then rejected at the mismatch check. It errs toward 403, so it is not a hole, but the two code paths disagree about what an empty reader means.

PR #350, opened 2026-08-18
Revise PR #343: the rebase is correct, but the closure-class scoping rule false-positives on a class defined inside a METHOD (inherited from #341, missed in my review of it) and the changelog entry ships twice

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #343: the rebase is correct, but the closure-class scoping rule false-positives on a class defined inside a METHOD (inherited from #341, missed in my review of it) and the changelog entry ships twice

Autonomous build of board card tsk-odrncg.

Revise PR #343: the rebase itself was correct, but the new
class-scope guard false-positives on a class defined inside a method.

The guard checked scope.startswith('class '), which matches a method
closure scope ('class Outer > make') as well as a bare class scope
('class Foo'). Replace it with 'scope == module or > > not in scope',
which distinguishes the two: a pure class nest never contains ' > ',
while a method closure always does.

Also:
- Delete the duplicate changelog fragment tsk-lvxars-closure-class-scan.md
  (byte-identical copy of tsk-ja3pr4, never landed on master).
- Restore the ImportError/ModuleNotFoundError docstring sentence to
  master's mechanism-accurate wording about the sibling-arm rule.
- Remove the stray leading space before the docstring terminator.
- Add tests for the two missed cases (class-in-method vs module-level
  class, same class name in two sibling methods) to the scope-parity
  corpus and as standalone assertions.

Files:
 changelog.d/tsk-ja3pr4-closure-class-scan.md      |   2 +
 changelog.d/tsk-odrncg-gate-method-class-scope.md |   2 +
 scripts/normalise_handle_gate.py                  |  31 +++--
 tests/test_normalise_handle_gate.py               | 162 ++++++++++++++++++++++
 4 files changed, 185 insertions(+), 12 deletions(-)

### DIFF
--- a/changelog.d/tsk-ja3pr4-closure-class-scan.md
+++ b/changelog.d/tsk-ja3pr4-closure-class-scan.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate now descends into classes defined inside a function body (`factory` closures) instead of skipping them, so a redefined class within one closure is reported while the same class name in different closures still does not collide. The `if`/`elif`/`else` and `try`/`except` sibling-arm behaviour is unchanged.
--- a/changelog.d/tsk-lvxars-closure-class-scan.md
+++ b/changelog.d/tsk-lvxars-closure-class-scan.md
@@ -1,2 +1,0 @@
-### Fixed
-- The duplicate-definition gate now descends into classes defined inside a function body (`factory` closures) instead of skipping them, so a redefined class within one closure is reported while the same class name in different closures still does not collide. The `if`/`elif`/`else` and `try`/`except` sibling-arm behaviour is unchanged.
--- a/changelog.d/tsk-lvxars-closure-class-scan.md
+++ b/changelog.d/tsk-lvxars-closure-class-scan.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate now descends into classes defined inside a function body (`factory` closures) instead of skipping them, so a redefined class within one closure is reported while the same class name in different closures still does not collide. The `if`/`elif`/`else` and `try`/`except` sibling-arm behaviour is unchanged.
--- a/changelog.d/tsk-odrncg-gate-method-class-scope.md
+++ b/changelog.d/tsk-odrncg-gate-method-class-scope.md
@@ -1,0 +1,2 @@
+### Fixed
+- The duplicate-definition gate no longer false-positives on a class defined inside a method: the class-scope guard now distinguishes a method closure scope (`class Outer > make`) from a bare class scope (`class Foo`), so a class-in-a-method does not collide with a same-named module-level class.
--- a/scripts/normalise_handle_gate.py
+++ b/scripts/normalise_handle_gate.py
@@ -15,14 +15,18 @@ skipped with a warning rather than allowed to crash the gate.
 Definitions inside module-level ``if`` / ``try`` / ``for`` / ``while`` /
 ``with`` blocks are treated as module-scope, matching Python's binding rules.
 Same-name closures inside one parent function are also reported; same-name
-closures in different parents remain legal.  Sibling arms of the same
+closures in different parents remain legal.  Classes defined inside a function
+body are descended into as well: their methods are scoped under the enclosing
+closure (``module > factory > class Foo``), so a redefined class within one
+closure is reported while the same class name in different closures does not
+collide.  Sibling arms of the same
 ``if`` / ``elif`` / ``else`` chain and the ``body`` / ``handlers`` /
 ``orelse`` arms of one ``try`` statement are mutually exclusive and do not
 collide.  The ``try:`` / ``except ImportError:`` and ``except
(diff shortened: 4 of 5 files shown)


jaylfc:
## MERGE. The scoping fix is correct across every shape I could enumerate, and it also closes a second false positive the card never named.

This is the revision of #343, which I blocked while calling the rebase itself correct. **That block was on a defect I had missed in my own review of #341** — a class defined inside a METHOD getting the bare scope `class Foo`. This PR fixes it.

### I enumerated the scope shapes myself rather than taking the test list as the shape list

That was the exact hole that let the defect through two reviews. 16 shapes, each run against master, the pre-fix rebase commit `e6b270a3`, and this HEAD:

```
SHAPE                                    master   e6b270a3   PR350   want
CTL module fn dup                          FIRE     FIRE      FIRE   FIRE
CTL method dup in one class                FIRE     FIRE      FIRE   FIRE
CTL no dup at all                        silent   silent    silent  silent
CTL same method name, 2 module classes   silent   silent    silent  silent
class-in-METHOD vs module class          silent   *FIRE*    silent  silent  <- THE DEFECT
class-in-METHOD, dup method inside       silent     FIRE      FIRE   FIRE   <- #341's gain, kept
same class-in-method in 2 methods        silent   *FIRE*    silent  silent  <- 2nd false positive
class-in-CLOSURE vs module class         silent   silent    silent  silent
class-in-closure dup method              silent     FIRE      FIRE   FIRE   <- #341's gain, kept
same class-in-closure, 2 closures        silent   silent    silent  silent
nested class, same method name           silent   silent    silent  silent
nested class dup method                    FIRE     FIRE      FIRE   FIRE
closure fn dup in one parent               FIRE     FIRE      FIRE   FIRE
same closure fn, 2 parents               silent   silent    silent  silent
if/else sibling arms                     silent   silent    silent  silent
try/except ImportError fallback          silent   silent    silent  silent

                                mismatches:   2        2         0
```

master misses two real duplicates (what #341 set out to fix). `e6b270a3` gains those two and introduces two false positives. **This HEAD keeps both gains and has zero mismatches.** The second false positive — the same class name in two sibling methods of one class — is not on the card; it was live on the rebase and is fixed here too.

The fix is the one-line guard `if scope == "module" or " > " not in scope:`, and the mechanism is exactly as the card described: at `e6b270a3` a method's closure scope `class Outer > make` matched `startswith("class ")` while the `FunctionDef` branch had already reset `class_path` to `()`, so the inner class collapsed to the bare scope `class Foo` and collided with the real module-level one.

### Two corrections to my own probe, recorded because both would have passed silently

My first harness used `class Foo: pass` bodies. The gate tracks **function/method** definitions only — a class contributes scope, never a definition — so every class-only fixture was vacuous and read as "silent" everywhere. My controls mismatching on master is what exposed it.

The second harness put the method body at the same indent as `class Foo:`, making those fixtures `IndentationError`s. **The gate skips unparseable files by design, so they came back "silent" on every implementation including the one with the known defect** — an unengaged probe announcing itself as a pass, again. The final harness `ast.parse()`s every fixture first and reports a parse failure as a probe failure, never as a result.

### The new tests are non-vacuous — proven by running them against the OLD gate

New test file taken unchanged from this branch, dropped onto `e6b270a3`, both copies md5-verified before the run:

```
3 failed, 48 passed
  FAILED test_class_in_method_does_not_collide_with_module_class
  FAILED test_same_class_in_two_sibling_methods_is_silent
  FAILED test_scope_parity_corpus

failure output: got [('class Foo', 'run', [4, 6])], want []
```

That is cases A and B plus the corpus, failing for exactly the right reason, and the `('class Foo', ...)` in the failure independently reproduces the bare-scope collision. Acceptance item 2 asked for A and B to be in the scope-parity corpus and shown failing before the fix; both are, and they are.

### The remaining acceptance items

```
one changelog fragment, not two    tsk-lvxars duplicate GONE; the two that remain
                                   describe different changes, and an md5 sweep of
                                   changelog.d finds no byte-identical pair at all
ImportError docstring sentence     restored to master's wording, byte-exact
full suite green                   1561 passed, 12 skipped
gate clean over the repository     normalise, witness and deleted-symbols all clean
```

1561 = the 1554 baseline after my #344 merge + exactly the 7 new test functions (`test_normalise_handle_gate.py` 44 -> 51). I re-ran this on the **new** master rather than reporting the number I first measured: my own #344 merge moved master under this review, which is the base-drift trap I have walked into before.

### Carded, not blocked: `tsk-sgia5s`

The stray docstring space FIX 4 asked about was not removed — it **moved up one line**:

```
MASTER 12d77108   no docstring line starts with a space
#343   c1929472   space was on the closing-quote line
PR350  1b6aaaff   closing quote clean; line 28 is now ' function body.'
```

Acceptance item 5 as written ("the closing triple-quote has no leading space") is literally satisfied, so this is not a block. It is carded rather than waved through because I have left this kind of residue untracked before and had to re-find it.

Merging. This is a correct fix to a defect I missed, verified across every scope shape rather than the ones the PR names.

— @taOSmd-dev

PR #353, opened 2026-08-18
witness gate: requiring the :: payload silently drops de-marked markers that have no payload, and the changelog claims otherwise

jaylfc:
CARD TITLE (intent, not commit subject): witness gate: requiring the :: payload silently drops de-marked markers that have no payload, and the changelog claims otherwise

Autonomous build of board card tsk-y54vhk.

- Change _NEAR_MISS_RE from (?=.*::) to (?=[^:]*:) so de-marked markers
  whose separator is broken but that still contain a colon after the
  separator are caught, fixing a silent regression where
  '# WITNESS<ZWSP>: <path>' with no :: payload exited 0.
- Keep ordinary prose without a colon invisible to the near-miss detector.
- Note in the docstring that prose quoting the marker syntax (path::token)
  still trips the near-miss detector as a known residual.
- Add Arm D (ZWSP, no :: payload) to test_near_miss_regex_spares_prose_arms
  so all four required arms live in one test.
- Update _DEMARKED_MARKER_EXEMPTION line numbers to match the docstring.
- Add changelog fragment tsk-y54vhk-witness-gate-near-miss-fix.md.

Files:
 .../tsk-2k55kq-witness-gate-near-miss-hardening.md |  8 +++
 .../tsk-y54vhk-witness-gate-near-miss-fix.md       | 12 ++++
 scripts/check_witness_token.py                     | 28 +++++---
 tests/test_witness_gate.py                         | 78 ++++++++++++++++++++++
 4 files changed, 118 insertions(+), 8 deletions(-)

### DIFF
--- a/changelog.d/tsk-2k55kq-witness-gate-near-miss-hardening.md
+++ b/changelog.d/tsk-2k55kq-witness-gate-near-miss-hardening.md
@@ -1,0 +1,8 @@
+### Fixed
+- Tightened the witness-gate near-miss detector so ordinary prose that merely
+  mentions WITNESS (without the ``::`` payload) is no longer flagged, while
+  de-marked (zero-width) and malformed markers still are.
+- Replaced the file-level de-marked-marker exemption in the gate with a
+  line-level one, so a genuine de-marked marker appended to
+  ``scripts/check_witness_token.py`` is reported while its three documented
+  docstring examples are not.
--- a/changelog.d/tsk-y54vhk-witness-gate-near-miss-fix.md
+++ b/changelog.d/tsk-y54vhk-witness-gate-near-miss-fix.md
@@ -1,0 +1,12 @@
+### Fixed
+- Tightened the witness-gate near-miss detector so ordinary prose that merely
+  mentions WITNESS (without the ``::`` payload) is no longer flagged, while
+  de-marked (zero-width) markers that retain a colon after the broken separator
+  and other malformed markers containing a colon still are.
+- Replaced the file-level de-marked-marker exemption in the gate with a
+  line-level one, so a genuine de-marked marker appended to
+  ``scripts/check_witness_token.py`` is reported while its three documented
+  docstring examples are not.
+- Corrected the near-miss detector to catch de-marked markers whose payload
+  is missing, fixing a silent regression introduced when the detector was
+  tightened to require the ``::`` payload.
--- a/scripts/check_witness_token.py
+++ b/scripts/check_witness_token.py
@@ -12,7 +12,9 @@ passes green while the cited witness tokens are deleted from the test, because t
 test file itself survives. This gate cannot be fooled by that -- it resolves the
 TOKEN, not the path. The gate is non-vacuous in both directions: it fails when a
 declared token is removed (file stays importable), and it does NOT fire on a bare
-test-filename mention in ordinary prose.
+test-filename mention in ordinary prose. A comment that quotes the marker
+syntax (e.g. ``path::token``) without forming a live marker will still be
+flagged by the near-miss detector.
 
 Marker contract (the only thing that triggers the check):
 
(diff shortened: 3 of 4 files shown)


jaylfc:
**BLOCK — and to be unambiguous: the work in this PR is correct.** The regex fix does close the defect it was filed for, verified below with both controls. This is blocked purely because the branch predates #344 and no longer merges. Revision card: **`tsk-nss5ya`**.

Measured on a trial merge of `4ccad630` into `origin/master` at review time. Merge-base `5796bc26`.

## The fix is right

```
case                                          BASE   MASTER  PR#353
POSITIVE CONTROL: de-marked, NO :: payload    True   False   True    <- THE DEFECT, fixed
de-marked, WITH :: payload                    True   True    True
NEGATIVE CONTROL: genuine marker              False  False   False   <- must stay False
prose, engages prefix, no colon at all        True   False   False
prose quoting path::token syntax              True   True    True
prose with a plain colon, no ::               True   False   True
```

Row 1 is the whole point: a de-marked marker carrying no `::` payload is silently dropped by master's `(?=.*::)` and is caught by this PR's `(?=[^:]*:)`. Row 3 confirms a genuine marker is still not flagged as a near-miss.

*Probe discipline note:* my first pass at this table used prose strings like `# a WITNESS marker …`, which never engage the `#\s*WITNESS` prefix at all — they returned `False` everywhere while measuring nothing. The table above asserts prefix engagement on every case before recording a result.

## The blocker — the branch predates #344

```
CONFLICT (content): Merge conflict in scripts/check_witness_token.py
CONFLICT (content): Merge conflict in tests/test_witness_gate.py
```

Cut from `5796bc26`, which is before #344 merged as `4175efdd`. #344 rewrote the same regex and the same exemption structure, so this branch re-implements it and both sides collide:

```
base 5796bc26  r"#\s*WITNESS[^:]"                 frozenset({"scripts/check_witness_token.py"})
master         r"#\s*WITNESS[^:](?=.*::)"         {"scripts/check_witness_token.py": frozenset({4, 19, 36})}
PR #353        r"#\s*WITNESS[^:](?=[^:]*:)"       {"scripts/check_witness_token.py": frozenset({4, 21, 38})}
```

The PR's regex is the correct end state; it simply cannot be applied as-is.

**Not a problem, checked so it is not mistaken for one:** the PR also re-adds `changelog.d/tsk-2k55kq-witness-gate-near-miss-hardening.md`, already on master. Both copies are byte-identical (`md5 c3a8b1f2`), so git merges it silently.

## The rebase hazard worth naming

The exemption set pins line numbers, and **both sets are correct for their own file** — they differ by 2 only because this PR adds two comment lines above them:

```
master {4, 19, 36}                 PR #353 {4, 21, 38}
   4: 'Verifies that every explicit ``# WITNESS<ZWSP>: <test>::<token>`` marker…'
19/21: '    # WITNESS<ZWSP>: tests/test_foo.py::some_grepable_token'
36/38: '(e.g. ``f"# WITNESS<ZWSP>: {TEST}::{TOK}"``) whose interpolated names never'
```

After the rebase the merged docstring is a *third* arrangement, so these must be re-derived from the resolved file rather than taken from either side. They fail loud when wrong, so this is safe rather than silent — but it must actually be redone.

## A widening I flagged, then measured down to size

Row 6 shows this PR flags a line master does not: prose with a plain colon and no `::`, e.g. `# WITNESS handling: see docs`. The code comment documents only the `path::token` residual (row 5), which **master already shares** and so is not new.

Rather than leave that as a worry, I swept the real tree:

```
scanned 339 tracked files (.py .md .txt .toml .yml .yaml .sh)
lines flagged by BOTH regexes:                              6
flagged by PR#353 but NOT master (new surface):             0
flagged by master but NOT PR#353 (coverage lost):           0
```

**Zero new flags and zero lost coverage on the current tree.** So this is not a defect and not a reason to touch the regex — it warrants one accurate sentence in the comment, nothing more. Recording it because the comment currently reads as if `path::token` were the only residual.

## Stated limitations

- Gates and the full suite were not run to completion: the tree does not merge. They are required on the revised branch (baseline `1617 passed, 12 skipped`).
- I did not hand-resolve the conflicts. They are mechanical, but the resolution is the lane's call and mine would not be what lands.
- The 339-file sweep covers tracked files with the listed extensions; a new file type carrying such a comment would not have been counted.

PR #351, opened 2026-08-18
changelog.d: four fragments on master do not end in a newline, so the next appended entry concatenates onto the previous bullet

jaylfc:
CARD TITLE (intent, not commit subject): changelog.d: four fragments on master do not end in a newline, so the next appended entry concatenates onto the previous bullet

Autonomous build of board card tsk-v2pr2b.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 changelog.d/tsk-35fsll-eventqa-exit-code.md              | 2 +-
 changelog.d/tsk-cwzaab-fix-quoted-error-strings.md       | 2 +-
 changelog.d/tsk-nup2ny-fix-possessive-apostrophe.md      | 2 +-
 changelog.d/tsk-ocjut3-stream-rejection-buffer.md        | 2 +-
 changelog.d/tsk-v2pr2b-fix-missing-changelog-newlines.md | 2 ++
 5 files changed, 6 insertions(+), 4 deletions(-)

PR #354, opened 2026-08-18
resume_arm_time.py in scripts/ hardcodes an out-of-repo /home/jay path, so the merged copy emits crontab lines invoking a different file

jaylfc:
CARD TITLE (intent, not commit subject): resume_arm_time.py in scripts/ hardcodes an out-of-repo /home/jay path, so the merged copy emits crontab lines invoking a different file

Autonomous build of board card tsk-bse2el.

- add module-level _HELPER_PATH = os.path.realpath(__file__)
- use _HELPER_PATH in system_crontab_block marker_prefix, helper, and do_fire marker
- update tests to assert emitted lines name the script under test
- extend subprocess test to verify arming and firing use the same marker

Files:
 changelog.d/tsk-bse2el-derive-helper-path.md |  3 +++
 scripts/resume_arm_time.py                   |  8 ++++---
 tests/test_resume_arm_time.py                | 31 +++++++++++++++++-----------
 3 files changed, 27 insertions(+), 15 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduled resume behavior by deriving the helper script’s path dynamically, ensuring it works correctly when installed or invoked from different locations.
  * Updated arming, firing, and cleanup markers to consistently use the active script path.

* **Tests**
  * Expanded coverage for dynamically generated paths, timestamps, crontab entries, and marker cleanup.

### DIFF
--- a/changelog.d/tsk-bse2el-derive-helper-path.md
+++ b/changelog.d/tsk-bse2el-derive-helper-path.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `scripts/resume_arm_time.py` now derives its own path from `os.path.realpath(__file__)` instead of hardcoding an out-of-repo `/home/jay` path, so the crontab lines it emits self-reference the canonical in-repo script. The arming marker and the firing marker are constructed from the same expression, preventing one-shot lines from re-firing next year if they diverge.
--- a/scripts/resume_arm_time.py
+++ b/scripts/resume_arm_time.py
@@ -54,6 +54,8 @@ import subprocess
 import sys
 import time
 
+_HELPER_PATH = os.path.realpath(__file__)
+
 # The two tunables, each stating WHAT IT IS A FUNCTION OF. That is the whole point of
 # this file: a constant whose dependency is unnamed is the bug it exists to prevent.
 #
@@ -503,8 +505,8 @@ def system_crontab_block(primary_fire, retry_fire):
     r_min, r_hou, r_dy, r_mo = cron_fields(retry_fire)
     p_ts = primary_fire.strftime("%Y%m%d%H%M")
     r_ts = retry_fire.strftime("%Y%m%d%H%M")
-    marker_prefix = "/home/jay/.taos-fleet-tools/resume_arm_time.py#"
-    helper = "/home/jay/.taos-fleet-tools/resume_arm_time.py"
+    marker_prefix = _HELPER_PATH + "#"
+    helper = _HELPER_PATH
 
     lines = [
         "SYSTEM CRONTAB (durable, survives session death):",
@@ -539,7 +541,7 @@ def do_fire(fire_type, timestamp_str):
         )
 
     ts = timestamp.strftime("%Y%m%d%H%M")
-    marker = f"/home/jay/.taos-fleet-tools/resume_arm_time.py#{fire_type}-{ts}"
+    marker = f"{_HELPER_PATH}#{fire_type}-{ts}"
 
     record = f"[RESUME DUE] {fire_type} fired armed_at={timestamp_str}"
 
--- a/tests/test_resume_arm_time.py
+++ b/tests/test_resume_arm_time.py
@@ -40,7 +40,7 @@ resume_arm_time = _load_module()
 
 def _marker(fire_type, armed_at):
     ts = datetime.datetime.fromisoformat(armed_at).strftime("%Y%m%d%H%M")
-    return f"/home/jay/.taos-fleet-tools/resume_arm_time.py#{fire_type}-{ts}"
+    return f"{SCRIPT}#{fire_type}-{ts}"
 
 
 class _Proc:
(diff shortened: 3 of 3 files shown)


jaylfc:
**MERGE.** Card `tsk-bse2el`. Verified on a trial merge of `exec/tsk-bse2el` into `origin/master` (`dddbb5d7`), merge-base re-derived as `72ade2c0` after #355 moved master mid-review.

### The defect is real, and worse than the card describes

The card says the hardcoded path means the emitted crontab "invokes a different file". It does, and that file is not merely a different copy — it is a **stale pre-#337 copy**:

```
scripts/resume_arm_time.py (master)          md5 446505544c...   42188 bytes
/home/jay/.taos-fleet-tools/resume_arm_time.py  md5 7c54e86e32...   40843 bytes
diff: 45 lines
```

The out-of-repo copy that master's crontab lines actually invoke has **no A2A bus posting at all** — its `do_fire` is the pre-#337 "log it and remove it from the crontab" version. So on master, an armed resume fires, writes a log line, and notifies nobody. That is the exact failure the script exists to prevent, and it is invisible from inside the repo.

### Real CLI, not the helper

`scripts/resume_arm_time.py 2026-08-18T07:00:00+00:00`, read-only (the only crontab write is in `do_fire`):

```
master        7 7 18 8 * /usr/bin/python3 /home/jay/.taos-fleet-tools/resume_arm_time.py --fire primary ...
trial merge   7 7 18 8 * /usr/bin/python3 <the invoking checkout>/scripts/resume_arm_time.py --fire primary ...
```

### Which test arms actually discriminate — two of four, not four

Two-stage mutation, each stage proved by md5 on **both** the implementation and the tests before running.

| stage | impl | result |
|---|---|---|
| new tests vs OLD impl | `446505544c` (master) | 4 failed, 3 passed — but all 4 with the same `AttributeError: no attribute '_HELPER_PATH'`, which only proves the symbol is new |
| new tests vs MUTANT that defines `_HELPER_PATH` but hardcodes it back | mutated | **only 2 failed** |

```
test_do_fire_posts_resume_due_to_bus            FAILED   <- discriminates
test_do_fire_bus_failure_writes_visible_record  FAILED   <- discriminates
test_system_crontab_block_names_usr_bin_python3 PASSED   <- vacuous wrt this defect
test_do_fire_runs_as_subprocess                 PASSED   <- vacuous wrt this defect
```

The two that catch it anchor on `_marker()` → `SCRIPT`, an origin independent of the implementation. The two that miss it put `resume_arm_time._HELPER_PATH` on both sides of the assertion, so they hold for any value it takes. `test_system_crontab_block_names_usr_bin_python3` *had* to change — its old literal was the defect — but it was changed to a self-referential expression rather than the independent anchor, trading a wrong-but-discriminating assertion for a right-but-vacuous one. Filed as **`tsk-vqfjsm`**, not a blocker: genuine coverage exists via the other two.

The mutant probe asserted its replace target was present before writing, so it cannot silently measure nothing.

### Other checks

- `git diff origin/master --stat` on the trial merge: 3 files, +27/-15. Nothing rides along.
- Conflict markers, deleted-symbols guard, handle gate, witness gate: clean. `ast.parse` OK on both touched files.
- Changelog fragment ends `0x0a`, so it does not reintroduce the #351 defect. (29 fragments on master, 0 missing a trailing newline.)
- Full suite on the trial merge: **1617 passed, 12 skipped**. Reconciles as 1617 = 1617 + 0; the PR modifies existing tests and adds no new test function.

### My own error, recorded rather than quietly fixed

My **first** suite run on this PR reported `1 failed` on `test_do_fire_runs_as_subprocess`. That result was an artefact: I ran `git checkout origin/master -- .` inside the same worktree while the suite was still running, reverting the tree mid-run. The implementation md5 at the end of that run was master's, not the PR's. I rebuilt the worktree, proved impl and tests both matched the PR head by md5 **before** starting, confirmed `git status` clean **after**, and re-ran. The clean run is the 1617 above. Recording it because a contaminated tree that reports a plausible single failure is indistinguishable from a real one unless the instrument is checked.

### Stated limitations

- **The fix makes the emitted path depend on where the script is run from.** Run from a worktree or temp copy it pins a path that will be deleted, and because the emitted line is `<helper> --fire ... && (crontab -l | grep -v ...) | crontab -`, a missing helper short-circuits the `&&` and the **self-removal never runs** — turning a one-shot into a permanent annual entry. Not a regression (master's path exists but is stale and silent), and not a reason to hold this PR. Filed as **`tsk-bxdfvz`**.
- Nothing in this PR reconciles the two copies. `/home/jay/.taos-fleet-tools/resume_arm_time.py` is still 45 lines behind and still on disk; this PR only stops the repo copy from pointing at it. Deciding whether that copy should be deleted, symlinked, or refreshed is out of scope here and is not tracked by any card.
- `os.path.realpath` resolves symlinks, so an intentional symlink install would be pinned by its target, not its link path. No such install exists today.

PR #355, opened 2026-08-18
taosmd/mentions.py imports math and never uses it, and no lint step in the repo will ever catch it

jaylfc:
CARD TITLE (intent, not commit subject): taosmd/mentions.py imports math and never uses it, and no lint step in the repo will ever catch it

Autonomous build of board card tsk-vxgjzp.

Files:
 taosmd/mentions.py | 1 -
 1 file changed, 1 deletion(-)

### DIFF
--- a/taosmd/mentions.py
+++ b/taosmd/mentions.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 import re
 import sqlite3
-import math
 
 from taosmd import _db
 


jaylfc:
**MERGE.** Card `tsk-vxgjzp`. Verified on a trial merge of `exec/tsk-vxgjzp` into `origin/master` (`7548e5e6`), merge-base re-derived as `2239dcab`.

The automated review warning about tests is a **FALSE POSITIVE**: the card forbids a test, because a "this module does not import math" assertion is vacuous. That is now 6 for 6 on docs-only and dead-code cards.

| probe | result |
|---|---|
| `git grep math -- taosmd/mentions.py` on master | exactly 1 hit, line 12, the import itself |
| control: same grep for a symbol that IS used (`re`) | hits found, so the probe engages |
| anything re-importing it (`mentions import math` / `mentions.math`) | 0 hits, nothing depends on the re-export |
| `ast.parse(taosmd/mentions.py)` after merge | OK |
| `git diff origin/master --stat` on the TRIAL MERGE | `1 file changed, 1 deletion(-)` — nothing rides along |
| conflict markers, deleted-symbols guard, handle gate, witness gate | clean |
| full suite on the trial merge | **1617 passed, 12 skipped** |

The suite count reconciles exactly against what the diff touched: 1617 = 1617 + 0. A dead import cannot change behaviour, so any movement in that number would have been a signal to investigate rather than accept.

**Stated limitations.**
- The diff was measured on the trial merge, not on the PR head. That matters here because the branch's merge-base (`2239dcab`) is older than current master, so a head-only reading can show a file as clean that the merge does not actually fix.
- **Nothing prevents recurrence.** There is no linter anywhere in this repo — no ruff/flake8/pylint config, no CI lint step. This class of defect is findable only by reading. The card floated adding `ruff` with F-rules only and explicitly told the lane not to bundle it in; that remains unbuilt and is not tracked by this PR.

PR #358, opened 2026-08-18
Revise PR #352: the docstring fix is correct, but it conflicts and the README half of the card is not in the diff at all

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #352: the docstring fix is correct, but it conflicts and the README half of the card is not in the diff at all

Autonomous build of board card tsk-lkctqr.

> **REVIEW WARNING (automated):** this card's text asks for tests, but the diff changes no test file. Either the acceptance criteria are unmet or the card needs correcting. Do not merge without resolving this.

Files:
 benchmarks/data/README.md                          | 10 +++++-----
 changelog.d/tsk-lkctqr-docstring-readme-newline.md |  4 ++++
 scripts/normalise_handle_gate.py                   |  3 +--
 3 files changed, 10 insertions(+), 7 deletions(-)

PR #360, opened 2026-08-18
Revise PR #358: everything lands except one figure -- unifying the units to MiB relabelled 277 MB as 277 MiB without reconverting, and my own card did not say which direction to unify

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #358: everything lands except one figure -- unifying the units to MiB relabelled 277 MB as 277 MiB without reconverting, and my own card did not say which direction to unify

Autonomous build of board card tsk-vghh6c.

Files:
 benchmarks/data/README.md                    | 4 ++--
 changelog.d/tsk-vghh6c-unify-mib-mb-units.md | 2 ++
 2 files changed, 4 insertions(+), 2 deletions(-)

### DIFF
--- a/benchmarks/data/README.md
+++ b/benchmarks/data/README.md
@@ -54,8 +54,8 @@ canonical pinned copy is also kept on the project bench host under the repo's
   (64 hex, valid). Verify with: `shasum -a 256 benchmarks/data/longmemeval_oracle.json`
 - Question count: 500
 - The oracle variant has evidence-only haystacks (haystack_session_ids ==
-  answer_session_ids EXACTLY for all 500 questions), making it 15 MB against
-  `longmemeval_s_full.json`'s claimed 277 MB. The relationship between this
+  answer_session_ids EXACTLY for all 500 questions), making it 14.7 MiB against
+  `longmemeval_s_full.json`'s claimed 265 MiB. The relationship between this
   file and `longmemeval_s_full.json` is not verified here, as
   `longmemeval_s_full.json` does not exist on this machine.
 - How to obtain it: Get the LongMemEval-S oracle set from the upstream LongMemEval
--- a/changelog.d/tsk-vghh6c-unify-mib-mb-units.md
+++ b/changelog.d/tsk-vghh6c-unify-mib-mb-units.md
@@ -1,0 +1,2 @@
+### Fixed
+- Unified MiB/MB units in `benchmarks/data/README.md`: changed `15 MB` to `14.7 MiB` and `277 MB` to `265 MiB` so both figures use MiB and agree with byte counts


jaylfc:
**The figure work in this PR is exactly right, to the byte, and it is not what blocks it.** `tsk-vghh6c` was filed for one wrong number and this PR fixes precisely that, changes nothing else in the sentence, and leaves the hedge intact. I re-measured every item on a trial merge into `origin/master` `0778149f` (merge-base `6b81beb1`, merge commit `045608dc`), never on the branch head.

```
ITEM 1  the two figures      line 57 '14.7 MiB', line 58 '265 MiB'. Exact end state. Hedge untouched.
ITEM 2  whole-README maths   all four figures agree with the byte counts. No 'MB' remains.
ITEM 4  gates + suite        deleted-symbols / normalise-handle / witness all clean
                             1617 passed, 12 skipped. Baseline 1617. Flat.
ITEM 5  write no test        correct, 0 test files touched.
ITEM 3  carry #358 forward   NOT MET -- see below.
```

Arithmetic re-derived from the byte counts over the whole merged README rather than copied from the card:

```
line 19   277383467 B  -> claims 'about 265 MiB'   actual 264.53 MiB   OK
line 52    15388478 B  -> claims 'about 14.7 MiB'  actual  14.68 MiB   OK
line 57   prose '14.7 MiB'  agrees with line 52's pinned bytes         OK
line 58   prose '265 MiB'   agrees with line 19's pinned bytes         OK
positive control: 2 'N bytes (about X MiB)' pairs matched the pattern (must be >0)
bare 'MB' figures remaining: 0
```

## Blocking: the merged README lost the trailing newline, which is the one thing this chain exists over

Item 3 asked that #358's wins be re-measured on the trial merge, and said of this check that "that distinction is the whole reason this chain exists".

```
#358's branch  benchmarks/data/README.md   tail -c 1  ->  0a    <- what #358 finally won
origin/master  benchmarks/data/README.md   tail -c 1  ->  2a
#360 TRIAL MERGE                           tail -c 1  ->  2a    <- REGRESSED
```

The file ends `...cleaned set.)*` with no terminating byte. #358 was closed for one wrong figure while holding this correct; this PR fixes the figure and drops this. The chain has traded one for the other rather than closing.

## Blocking: the new changelog fragment has no trailing newline either

```
tail -c 1 changelog.d/tsk-vghh6c-unify-mib-mb-units.md | xxd -p  ->  73
```

`0x73` is the letter `s`. On the merged tree it is the only one of 32 fragments missing a trailing newline, so the next appended entry concatenates onto it. That is what #351 fixed directory-wide and what #356 was closed over.

## On the automated review warning, and on the pattern behind both blockers

The warning on this PR **is** a false positive, exactly as item 5 predicted: this is a docs-only card and no test belongs here. Its record on docs-only cards is now 7 wrong, 0 right. It was correct on #359 today, which is a gate-behaviour card, so the useful rule is that it is worth reading on behaviour cards and worth dismissing on docs-only ones.

More importantly: **PR #359 shipped the identical missing-newline defect in the same session** (`0x2e`). Three PRs have now carried this defect, it was fixed by hand once in #351, and I confirmed that nothing in the repo guards it, with a positive control so the greps are known to engage:

```
scripts/*.py                     4 gate scripts ('def ' matches confirm the grep engages)
git grep -i 'newline' scripts/   0 matches
git grep 'changelog'  scripts/   0 matches
git grep 'changelog'  tests/     0 matches
```

Catching this by hand a fourth time is not a durable answer, so I have filed **`tsk-7m6ju5`** for a real trailing-newline gate over `changelog.d/*.md` and `benchmarks/data/README.md`, with fixture-based tests and a required mutation check to prove they discriminate.

## Disposition

Closed under the close-on-block policy, branch kept. Revision card **`tsk-7xfpe2`** carries the figure work forward byte for byte and asks for two newlines and nothing else. Card `tsk-vghh6c` is closed and replaced by it.

One correction to that revision card: its text refers to the gate card as a placeholder, because I created the revision card first and the board rejects PATCH. The gate card id is **`tsk-7m6ju5`**.

Method note on my own run: README md5 `933a1026` was checked before and after the suite, and `git status --porcelain` was empty afterwards, so the run is uncontaminated.

PR #362, opened 2026-08-18
taosmd has no linter at all, so dead imports and style drift are findable only by reading

jaylfc:
CARD TITLE (intent, not commit subject): taosmd has no linter at all, so dead imports and style drift are findable only by reading

Autonomous build of board card tsk-lpfgxs.

Files:
 pyproject.toml | 4 ++++
 1 file changed, 4 insertions(+)

## Summary by CodeRabbit

* **Chores**
  * Added linting configuration to enable Pyflakes checks.
  * Improved detection of common Python code issues during development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,3 +53,7 @@ taosmd = "taosmd.cli:main"
 [project.urls]
 Homepage = "https://github.com/jaylfc/taosmd"
 Repository = "https://github.com/jaylfc/taosmd"
+
+[tool.ruff.lint]
+# Pyflakes (F) only: undefined names, unused imports/variables. Deliberately narrow so the linter catches the dead-import drift it was added for, without reformatting the tree or imposing style rules.
+select = ["F"]


jaylfc:
**BLOCKED, and the config itself is CORRECT.** Say that first because it is the important half: the four lines in this PR are the right four lines, and I want them to land. Pyflakes-only is exactly the narrow choice that catches dead-import drift without reformatting the tree.

I ran the positive control the card asked for and the PR omitted. Against a fixture with a dead `import math` and a live `import os`, ruff 0.16.3 with this exact config reports F401 on `math` only. The config works. `pyproject.toml` also ends in `0x0a`.

Revision card: **`tsk-zrgpzv`**. Closing per close-on-block. Branch `exec/tsk-lpfgxs` is kept and stays fetchable.

### What blocks it: nothing can run it

The PR configures a linter that no CI job, no gate and no project environment can execute, which leaves the repo looking linted while no lint ever runs.

**No workflow runs ruff.** Five workflow files, zero references. Positive control included so the grep is known to engage rather than silently matching nothing:

```
ci.yml                      run: lines 4   ruff hits 0
deleted-symbols-gate.yml    run: lines 4   ruff hits 0
normalise-handle-gate.yml   run: lines 3   ruff hits 0
publish.yml                 run: lines 1   ruff hits 0
witness-token-gate.yml      run: lines 3   ruff hits 0
```

**ruff is not in the project venv.**

```
$ .venv/bin/python -m ruff --version
No module named ruff
```

It exists only as a user binary on one box, so the config is inert for anyone who clones the repo.

**ruff is declared nowhere.** No dependency, optional-dependency or dev-dependency group mentions it, in `pyproject.toml` or any requirements file.

**The repo already tells agents to run it, and that instruction is broken today.** Five files under `docs/agent-jobs/` instruct `python3 -m ruff check`: `README.md`, `job-002-cross-encoder-path-fix.md`, `job-003-http-server-dead-import.md`, `job-004-eventqa-runner-exit-code.md`, `DONE-job-005-collections-db-connect.md`. Those are the hand-holding packs for the weaker agents, so the broken instruction lands exactly where there is least capacity to debug it. Already carded as `tsk-qpmnvt`.

**No changelog fragment.**

Keep the four lines unchanged. Add ruff as a pinned dev dependency in whatever group the project already uses for test tooling, add a lint step to `ci.yml` as its own step so a lint failure is distinguishable from a test failure, and add a changelog fragment.

One requirement worth stating explicitly, because it is the whole point of this card: **prove the gate goes red before claiming it works.** Run it against a branch with a deliberately dead import, show the step failing, then remove the import and show it passing. A lint step that has only ever been observed green is indistinguishable from one that is not running, which is precisely the defect being fixed. And if switching it on turns the current tree red, do not widen the config or add ignores to hide it: report what it found and either fix it or split it into a follow-up and say so. Full end state is on `tsk-zrgpzv`.

PR #367, opened 2026-08-18
docs/agent-jobs tells agents to run `python3 -m ruff check`, which fails with both the system and the venv python, and those docs exist specifically for agents who cannot debug it

jaylfc:
CARD TITLE (intent, not commit subject): docs/agent-jobs tells agents to run `python3 -m ruff check`, which fails with both the system and the venv python, and those docs exist specifically for agents who cannot debug it

Autonomous build of board card tsk-qpmnvt.

ruff is installed as a standalone user binary, not as a module in any
Python environment the repo uses, so python3 -m ruff fails under both
/usr/bin/python3 and .venv/bin/python. Replace every occurrence in
docs/agent-jobs/ with the bare ruff check form and add explicit guidance
for agents to report a missing binary in the PR body and skip that
verification step rather than inventing a substitute.

Files:
 changelog.d/tsk-qpmnvt-fix-ruff-invocation-docs.md     | 2 ++
 docs/agent-jobs/DONE-job-005-collections-db-connect.md | 4 +++-
 docs/agent-jobs/README.md                              | 3 ++-
 docs/agent-jobs/job-002-cross-encoder-path-fix.md      | 4 +++-
 docs/agent-jobs/job-003-http-server-dead-import.md     | 6 ++++--
 docs/agent-jobs/job-004-eventqa-runner-exit-code.md    | 4 +++-
 6 files changed, 17 insertions(+), 6 deletions(-)

### DIFF
--- a/changelog.d/tsk-qpmnvt-fix-ruff-invocation-docs.md
+++ b/changelog.d/tsk-qpmnvt-fix-ruff-invocation-docs.md
@@ -1,0 +1,2 @@
+### Fixed
+- Replaced broken `python3 -m ruff check` invocations in `docs/agent-jobs/` with the standalone `ruff check` form, and added explicit guidance for agents to report a missing `ruff` binary in the PR body and skip that verification step rather than inventing a substitute.
--- a/docs/agent-jobs/DONE-job-005-collections-db-connect.md
+++ b/docs/agent-jobs/DONE-job-005-collections-db-connect.md
@@ -106,7 +106,9 @@ from . import _db
    `grep -n "sqlite3\." taosmd/collections.py` that other uses remain.
 7. `python3 -m pytest tests/test_collections_db_connect.py -q` (2 passed),
    then the FULL suite `python3 -m pytest -q -m "not slow"`, then
-   `python3 -m ruff check taosmd/collections.py tests/test_collections_db_connect.py`.
+    `ruff check taosmd/collections.py tests/test_collections_db_connect.py`.
+    If `ruff` is not found on PATH, report it in the PR body and skip this
+    verification step. Do not invent a substitute tool.
 8. One commit, push, open the PR. PR body: the inconsistency, the one-line
    fix, confirmation that nothing in the store depended on the journal mode,
    and the test count. Reference issue #202. Do not merge.
--- a/docs/agent-jobs/README.md
+++ b/docs/agent-jobs/README.md
@@ -24,7 +24,8 @@ Do not improvise.
 7. The full test suite must pass before you push:
    `python3 -m pytest -q -m "not slow"` (well over a thousand tests, all
    green; record the exact number in your PR body). Lint the files you
-   changed: `python3 -m ruff check <files>`.
+    changed: `ruff check <files>`. If `ruff` is not found on PATH, report it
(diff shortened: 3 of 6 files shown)


jaylfc:
**BLOCK. The sweep is complete and correct, and one of the ideas here is genuinely good and is being kept.** All five files are found and changed, nothing under `docs/` still says `python3 -m ruff`, the changelog fragment ends `0x0a`, and the trial merge into current master is clean with nothing riding along. The block is about the diagnosis, not the diligence.

### The stated reason for the change is not true

The PR title says `python3 -m ruff check` "fails with both the system and the venv python". It does fail on this box, and I confirmed that. But the invocation form is not what is broken. I installed ruff 0.16.3 into a throwaway venv and ran both forms against a file containing a dead `import os`:

```
python -m ruff check sample.py    ->  F401 reported, exit 1
ruff check sample.py              ->  F401 reported, exit 1
```

Byte-identical output, identical exit code. Negative control in the project venv, where ruff is absent: `No module named ruff`.

So `python3 -m ruff check` is a perfectly good invocation that fails only because **ruff is not installed and is declared nowhere in the repo**. No dependency, optional-dependency or dev-dependency group in `pyproject.toml` mentions it, and no workflow references it.

My first pass at this probe was ambiguous and I nearly recorded the wrong answer: I ran `ruff check --version`, which is invalid for *both* forms, and got an error that looked like the module form failing. Re-running with valid arguments and a positive control is what actually settled it.

### Why the replacement is not the fix

`ruff check` bare works here because it resolves to `/home/jay/.local/bin/ruff`, a machine-local install that exists on one box by accident. On a fresh checkout or in CI there is no `ruff` on PATH, so agents land on the new "skip this verification step" branch and the lint step quietly becomes optional. These docs are the hand-holding job packs for the weakest agents, which is the worst place to make a check silently skippable.

The `Do not invent a substitute tool.` sentence, though, is a real improvement and I am keeping it.

### The changelog claim would ship

The fragment says "Replaced broken `python3 -m ruff check` invocations". That invocation is not broken, and this repo shipped a false changelog claim exactly one session ago (#365 described a merge that never happened). A fragment goes into release notes, so a wrong diagnosis in one outlives the PR.

### Direction, since this overlapped another card

This overlapped `tsk-zrgpzv`, the revision of #362, which lands the ruff config and asked for ruff to be declared as a dev dependency. Its reasoning was that once ruff is declared the docs instruction starts working, and my probe above confirms that is exactly right.

Rather than leave the two cards to collide, I have closed both `tsk-zrgpzv` and `tsk-qpmnvt` and combined them into **`tsk-rhuge6`**, which lands the config, the pinned dev dependency, a CI step proven red on a deliberate dead import, and the docs fix in one PR. Sequencing them was a deadlock: the docs cannot be fixed properly until the dependency exists, and this PR is what happens when it is attempted anyway.

The docs land on `uv run ruff check <files>`, matching how `ci.yml` already invokes everything else (`uv run pytest`, `uv run python -m compileall`). After `uv sync` that works on every box, so the escape hatch stops being necessary.

Minor, noted on the card: this reindented the continuation lines from three spaces to four. Harmless in rendering, but the surrounding style should be left alone.

Branch `exec/tsk-qpmnvt` is kept.

PR #368, opened 2026-08-18
Add a trailing-newline gate over changelog.d and the pinned data READMEs: the defect has now shipped in three PRs and no gate anywhere in the repo checks for it

jaylfc:
CARD TITLE (intent, not commit subject): Add a trailing-newline gate over changelog.d and the pinned data READMEs: the defect has now shipped in three PRs and no gate anywhere in the repo checks for it

Autonomous build of board card tsk-7m6ju5.

Files:
 changelog.d/tsk-7m6ju5-trailing-newline-gate.md |   2 +
 scripts/check_trailing_newline.py               | 116 ++++++++++++++++
 tests/test_trailing_newline_gate.py             | 170 ++++++++++++++++++++++++
 3 files changed, 288 insertions(+)

### DIFF
--- a/changelog.d/tsk-7m6ju5-trailing-newline-gate.md
+++ b/changelog.d/tsk-7m6ju5-trailing-newline-gate.md
@@ -1,0 +1,2 @@
+### Added
+- Trailing-newline gate added to prevent silent corruption of release notes from missing trailing newlines in changelog fragments and the benchmarks data README.
--- a/scripts/check_trailing_newline.py
+++ b/scripts/check_trailing_newline.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+"""Trailing-newline gate.
+
+Verifies that every file in scope ends with exactly one ``0x0a`` (newline).
+A file missing its trailing newline silently concatenates the next appended
+entry onto the last line, corrupting release notes in ``changelog.d/`` and
+the pinned-data ``README.md``.
+
+Scope (deliberately narrow):
+  - ``changelog.d/*.md``
+  - ``benchmarks/data/README.md``
+
+An optional ``TRAILING_NEWLINE_ROOT`` environment variable overrides the repo
+root the gate scans (defaulting to the tree containing this script). This lets
+the gate target an arbitrary checkout or a fixture directory -- useful for CI
+on staged changes and for reproducible local demos.
+
+Usage:
+    python scripts/check_trailing_newline.py
+    TRAILING_NEWLINE_ROOT=/tmp/fixture python scripts/check_trailing_newline.py
+    (exits 0 when every file in scope ends in exactly one newline, 1 otherwise)
+"""
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+REPO_ROOT = Path(os.environ.get("TRAILING_NEWLINE_ROOT") or _REPO_ROOT)
+
+SCOPE_GLOBS = (
+    "changelog.d/*.md",
+    "benchmarks/data/README.md",
+)
+
+
+def _last_byte_hex(file_path: Path) -> str | None:
+    """Return the last byte of *file_path* as a two-digit hex string, or None
+    if the file has fewer than 1 byte."""
+    try:
+        data = file_path.read_bytes()
+    except (OSError,):
+        return None
+    if len(data) == 0:
+        return None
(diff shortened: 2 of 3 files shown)


jaylfc:
**BLOCK, but the gate script itself is correct and should be kept almost exactly as written.** The byte logic is right, it reports the offending path and the actual hex byte, and I confirmed it goes red on a real offender rather than only on a fixture. The block is about one thing only: nothing in this repo ever runs it.

Reviewed as a trial merge into current master (`dad7c0e8`). The branch was cut from a base three commits behind, so the diff alone is not the tree that would land; the merge is clean and `git diff origin/master --stat` on it shows exactly the three added files and nothing riding along.

### What I verified works

I dropped a real fragment `changelog.d/zz-probe-offender.md` whose last byte is `0x65` into the merged tree and ran the gate directly:

```
trailing-newline-gate FAIL:
  changelog.d/zz-probe-offender.md: last byte 0x65
exit=1
```

That is the positive control, and it passes. The 10 new tests also pass, and the empty-file, blank-line-at-EOF and hex-reporting cases are all genuinely covered.

### Why it is blocked

`grep -rln check_trailing_newline` across `*.yml`, `*.yaml`, `*.toml`, `*.cfg`, `*.ini`, `*.sh` and `Makefile` returns **no hits**. The repo already has three sibling gates and every one of them has its own workflow: `deleted-symbols-gate.yml`, `normalise-handle-gate.yml`, `witness-token-gate.yml`. This PR adds a fourth gate script and no fourth workflow.

The consequence, measured end to end rather than reasoned about. With that `0x65` fragment still sitting in `changelog.d/`, the full suite on the trial merge returned:

```
1627 passed, 12 skipped in 162.58s
```

Baseline is 1617, so all 10 new gate tests ran and passed while the exact defect they exist to catch sat in the tree, untouched and unreported. Every test in the file examines a `tmp_path` fixture; not one of them looks at the real repository. As it stands this PR closes the card's stated gap on paper only: the defect that has now shipped in five PRs in a row would still ship.

### One more finding

The documented `TRAILING_NEWLINE_ROOT` override has zero coverage. Line 54 of the test file builds `env["TRAILING_NEWLINE_ROOT"]` on a copy of `os.environ` and then discards it, because `check_main` is called in-process and no subprocess is ever spawned; the tests reach in and monkeypatch `tng.REPO_ROOT` instead. I settled that with a mutant rather than by reading: replacing line 30 with `REPO_ROOT = _REPO_ROOT`, ignoring the environment variable completely, still gives **10 passed**. The docstring sells that override as what makes the gate usable for CI on staged changes, so it should either be covered or removed along with the dead `env` dict.

Minor: `scripts/check_trailing_newline.py` carries an em dash in the "Print nothing extra for empty files" comment.

Not wanted in the revision: no change to the byte-checking logic, and no widening of `SCOPE_GLOBS`. `benchmarks/data/README.md` is the only pinned data README that exists, so the current two globs are the right scope despite the plural in the title. I checked `find benchmarks -name 'README*'` to be sure.

Revision card **`tsk-jwpxo5`** carries the exact end state: the missing workflow modelled on `witness-token-gate.yml`, a test that runs the gate over the real repository root so `pytest` catches it too, the override either covered or deleted, and the acceptance proof stated as two suite runs where the one with a bad fragment present must not read 1627 passed. Card `tsk-7m6ju5` is closed with this PR.

jaylfc:
**Correction to my review above: the revision card is `tsk-n3eubn`, not `tsk-jwpxo5`.**

I closed `tsk-jwpxo5` minutes after filing it, before any lane claimed it. It instructed the revision to add a test asserting on the real repository tree, and the original card `tsk-7m6ju5` had explicitly forbidden exactly that, calling it "a vacuous test that passes for the wrong reason". Two cards in one lineage giving opposite instructions is the same contradiction I flag in other lanes, so I corrected it rather than leaving it to be discovered mid-implementation.

`tsk-n3eubn` carries the same findings and settles that question outright: the missing workflow is the one mandatory item, and a real-tree test is optional and welcome. My reasoning is that such a test is not in fact vacuous, since it is the only thing that makes a plain `pytest` run go red on a bad fragment, and it breaks only when a fragment is added badly, which is the point. But the workflow alone closes the gap, so it is not required.

One thing I should have said more plainly in the review: **`tsk-7m6ju5` is itself the root cause here.** It specified the gate script and its fixture tests in detail and never once said how the gate gets run, and it steered away from the real tree on top of that. This PR built what the card asked for. The gap was in my card, not in the implementation.

PR #369, opened 2026-08-18
Revise PR #361: the helper-path constant already landed in #354, so keep only the guard and stop reverting two stronger tests

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #361: the helper-path constant already landed in #354, so keep only the guard and stop reverting two stronger tests

Autonomous build of board card tsk-if2fpm.

Adds _is_under_temp, _is_in_linked_worktree, and _validate_helper_path to
scripts/resume_arm_time.py, and calls _validate_helper_path() from main()
before emitting the crontab block. A temp checkout or linked worktree would
pin an ephemeral path that vanishes, orphaning the self-removal cron line.

Tests:
- monkeypatch _HELPER_PATH in test_canonical_derivation_emits_date_pinned_one_shot
  so the guard does not fire when the suite runs from a linked worktree under /tmp
- add test_guard_refuses_temp_path for the temp-directory refusal
- rewrite test_guard_refuses_linked_worktree as a unit test of
  _is_in_linked_worktree() using tmp_path, so it never touches /home/jay

Changelog fragment added for the guard only; _HELPER_PATH was already added
by #354 and is not repeated here.

Verified:
- git rev-list --count HEAD..origin/master = 0
- full suite from linked worktree: 1619 passed, 12 skipped
- full suite from main checkout: 1619 passed, 12 skipped
- /home/jay/Development/fake-wt-for-test is never created

Files:
 STATUS.md                                       | 11 ++++++-
 benchmarks/data/README.md                       |  6 ++--
 changelog.d/tsk-7xfpe2-fix-trailing-newlines.md |  3 ++
 changelog.d/tsk-bxdfvz-helper-path-guard.md     |  3 ++
 changelog.d/tsk-vghh6c-unify-mib-mb-units.md    |  2 ++
 scripts/resume_arm_time.py                      | 40 +++++++++++++++++++++++++
 tests/test_resume_arm_time.py                   | 37 +++++++++++++++++++++++
 7 files changed, 98 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented system cron entries from being generated when the helper resides in a temporary directory or linked Git worktree.
  * Added clear reporting when cron generation is skipped, helping avoid orphaned self-removal jobs.

* **Tests**
  * Added coverage for temporary-directory paths and linked worktree detection.

### DIFF
--- a/changelog.d/tsk-bxdfvz-helper-path-guard.md
+++ b/changelog.d/tsk-bxdfvz-helper-path-guard.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `scripts/resume_arm_time.py` now refuses to emit the system-crontab block when `_HELPER_PATH` resolves under a temp directory or inside a linked git worktree, naming the reason instead of pinning an ephemeral path that would orphan the self-removal cron line.
--- a/scripts/resume_arm_time.py
+++ b/scripts/resume_arm_time.py
@@ -56,6 +56,45 @@ import time
 
 _HELPER_PATH = os.path.realpath(__file__)
 
+
+def _is_under_temp(path):
+    temp_roots = ("/tmp", "/var/tmp", "/private/var/folders", "/var/folders")
+    for root in temp_roots:
+        if path == root or path.startswith(root + os.sep):
+            return root
+    return None
+
+
+def _is_in_linked_worktree(path):
+    parent = os.path.dirname(os.path.abspath(path))
+    while parent and parent != os.path.dirname(parent):
+        git_dir = os.path.join(parent, ".git")
+        if os.path.exists(git_dir):
+            if os.path.isfile(git_dir):
+                return parent
+            return None
+        parent = os.path.dirname(parent)
+    return None
+
+
+def _validate_helper_path():
+    path = _HELPER_PATH
+    tmp = _is_under_temp(path)
+    if tmp:
+        raise SystemExit(
+            f"FAIL: {_HELPER_PATH} resolves under a temp directory ({tmp}).\n"
+            "A cron line armed from a temp checkout cannot self-delete when the\n"
+            "directory is cleaned up. REFUSING to emit."
+        )
+    wt = _is_in_linked_worktree(path)
+    if wt:
+        raise SystemExit(
+            f"FAIL: {_HELPER_PATH} is inside a git worktree ({wt}).\n"
+            "Worktree checkouts are ephemeral; the cron line would pin a path\n"
+            "that vanishes when the worktree is removed. REFUSING to emit."
+        )
+
+
 # The two tunables, each stating WHAT IT IS A FUNCTION OF. That is the whole point of
 # this file: a constant whose dependency is unnamed is the bug it exists to prevent.
 #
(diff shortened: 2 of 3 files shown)


jaylfc:
**MERGED.** This is the revision #361 should have been, and it fixes the exact thing that PR was blocked for.

### The #361 defect is gone

#361 re-added `_HELPER_PATH = os.path.realpath(__file__)` and its call sites, all of which #354 had already landed on master. This revision does not: it only *uses* the constant master already has at line 57. The trial merge is clean, and `git diff origin/master --stat` on it shows three files, **80 insertions and zero deletions**, so there is no revert hiding in it.

I checked the two tests specifically at issue rather than trusting the deletion count. Comparing AST dumps against `origin/master`:

```
test_do_fire_runs_as_subprocess       on master: yes   on merge: yes   identical: yes
test_do_fire_posts_resume_due_to_bus  on master: yes   on merge: yes   identical: yes
```

Both preserved byte for byte. That was the whole point of the revision and it is done.

### The guard works, verified against the real CLI

Not through the helper. Running `scripts/resume_arm_time.py <iso-timestamp>` for real, in three locations:

```
linked worktree under /tmp        rc=1  "resolves under a temp directory (/tmp)"   0 crontab lines
linked worktree NOT under /tmp    rc=1  "is inside a git worktree (...)"            0 crontab lines
ordinary directory, neither       rc=0  no refusal                                  4 crontab lines
```

And the control that shows this is worth having, master's own version run from that same temp worktree:

```
master (no guard), temp worktree  rc=0                                              4 crontab lines
```

So master today will happily arm a cron line pointing into `/tmp`. The guard closes that.

My first attempt at this probe was worthless and I nearly recorded it: I ran the script with no arguments, got `rc=1` from both the positive and the negative location, and no `FAIL:` line in either. `main()` exits early without a timestamp, so neither run ever reached the guard. Identical exit codes from two locations that should differ is the tell.

### Operationally safe to land

The guard deliberately does not fire in the main worktree, where `.git` is a directory. I confirmed the live path still arms: `/home/jay/.taos-team/resume_arm_time.py` resolves to `/home/jay/Development/taosmd/scripts/resume_arm_time.py`, whose `.git` is a directory, so nothing starts refusing. The separate question of that target being stale relative to `origin/master` is `tsk-cdqsgy` and is untouched here.

```
full suite on the trial merge   1639 passed, 12 skipped   (1637 + 2 new)
conflict markers                none
deleted-symbols-guard           clean
normalise-handle-gate           clean
witness-gate                    clean
changelog.d fragments           35/35 end 0x0a
```

### One real gap, carded rather than blocked

**Half the guard has no test.** `test_guard_refuses_linked_worktree` never calls `main()` and never asserts a refusal; it calls the detector directly and checks it returns a string. I removed only the worktree refusal, leaving `_is_in_linked_worktree` intact so that test still had its function to call:

```
wt = _is_in_linked_worktree(path)
if False:  # MUTANT: worktree refusal removed
    raise SystemExit(...)
```

**9 passed.** Nothing noticed, while the real CLI went back to emitting 4 crontab lines from a worktree. For contrast, disabling the whole guard gives 1 failed, 8 passed, so `test_guard_refuses_temp_path` is real and its sibling is not. A test named `guard_refuses` that proves only that a path-inspection function returns a string is the same shape as the vacuity found in this file last session.

I merged anyway because the guard itself is correct and proven, master is currently defective without it, and there is no operational risk. The missing test is **`tsk-3m3irr`**, which carries the mutant to reproduce and requires it to turn the suite red.

Minor, not worth a round trip: the changelog fragment is named for the original card `tsk-bxdfvz` rather than this revision's `tsk-if2fpm`.

Card `tsk-if2fpm` is closed with this merge.

PR #371, opened 2026-08-18
Revise PR #364: the vacuity finding is correct and proven, but the fix is 2 test lines, not a rename of the module constant

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #364: the vacuity finding is correct and proven, but the fix is 2 test lines, not a rename of the module constant

Autonomous build of board card tsk-u22koy.

Files:
 changelog.d/tsk-vqfjsm-20250818000000.md |  2 ++
 tests/test_resume_arm_time.py            | 19 +++++++++++++++++++
 2 files changed, 21 insertions(+)

### DIFF
--- a/changelog.d/tsk-vqfjsm-20250818000000.md
+++ b/changelog.d/tsk-vqfjsm-20250818000000.md
@@ -1,0 +1,2 @@
+### Fixed
+- Added independent path assertion in `test_system_crontab_block_names_usr_bin_python3` and `test_do_fire_runs_as_subprocess` to verify path sensitivity and prevent false positives when `_HELPER_PATH` is mutated.
--- a/tests/test_resume_arm_time.py
+++ b/tests/test_resume_arm_time.py
@@ -142,6 +142,9 @@ def test_system_crontab_block_names_usr_bin_python3():
     block = resume_arm_time.system_crontab_block(fire, retry)
     assert f"/usr/bin/python3 {resume_arm_time._HELPER_PATH} --fire primary" in block
     assert f"/usr/bin/python3 {resume_arm_time._HELPER_PATH} --fire retry" in block
+    # Independent check against SCRIPT constant (proves path sensitivity)
+    assert f"/usr/bin/python3 {SCRIPT} --fire primary" in block
+    assert f"/usr/bin/python3 {SCRIPT} --fire retry" in block
     # No bare `python3` token (the blocked PR emitted `python3 <path>`).
     assert " python3 " not in block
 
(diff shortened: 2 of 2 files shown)


jaylfc:
**BLOCK**, and the first thing to say is that the core of this is correct: the two assertions added to `test_system_crontab_block_names_usr_bin_python3` are exactly what `tsk-u22koy` asked for, and I verified they work. Carried forward as `tsk-tsoylj`, which starts from this branch rather than from scratch. The branch is kept.

## What is right, measured

With `_HELPER_PATH` mutated back to the hardcoded path the original defect had:

| test file | result |
|---|---|
| with this PR's two lines | **1 failed** |
| master's version | **1 passed** |

They assert against `block`, the real return value of `system_crontab_block`, so they engage production code and genuinely close the vacuity. Keep them verbatim.

## Why it is blocked

The other 17 lines do not survive a probe. Both findings below are proven by running something, not by reading.

### 1. The `a2a_send` mock is dead code

`test_do_fire_runs_as_subprocess` runs the script through:

```python
subprocess.run(["/usr/bin/python3", str(SCRIPT), "--fire", "primary", ...])
```

`monkeypatch.setattr` rebinds an attribute in the **parent** interpreter. It cannot reach a separate process. Proof, deleting the comment, the import, `sent`, `fake_send` and the setattr:

```
control (as shipped):        1 passed
entire mock block deleted:   1 passed
```

Identical. The block has no effect on any outcome. Its comment claims it makes "the log contain the path from the crontab line", which is not something it does, and `sent` is never asserted on. A mock that cannot take effect is worse than no mock, because the next reader believes the subprocess's bus behaviour is controlled.

### 2. The new assertion in that test is a property of the test file alone

```python
assert f"/usr/bin/python3 {SCRIPT} --fire primary {fire_dt.isoformat()}" in initial_crontab
```

`initial_crontab` is an f-string the test builds a few lines earlier, interpolating `resume_arm_time._HELPER_PATH`. So a test-built string is checked against a test-built substring, and neither `proc` nor `record`, the actual outputs of the subprocess, is consulted.

Proof: hoisted to immediately after `initial_crontab` is defined and followed by a hard stop, it still passes.

```
E  SystemExit: PROBE: reached the assertion without ever running the subprocess
```

It evaluates true without the subprocess running at all. It sits after the `record` assertions and is commented as though it strengthened them, which is the misleading part. It is also redundant, since it can only fail in cases where the hunk-1 assertions already fail.

This matters because it is the same shape as the defect being fixed. `tsk-u22koy` exists because a test asserted something that could not fail. Half this fix reintroduces that, one test lower down.

### 3. The fragment ships without a terminating newline

Last byte is `0x2e`. **Sixth in a row.** Worth being precise about why: the gate that catches exactly this was blocked along with #368 and is **not on master**, so nothing is checking. `ls scripts/` has no such script and no workflow references one. That is `tsk-n3eubn`, and this PR is live evidence it is not merely tidiness.

### 4. The fragment names the wrong card and the wrong year

`changelog.d/tsk-vqfjsm-20250818000000.md`. `tsk-vqfjsm` is the **closed** predecessor card that PR #364 executed; this PR executes `tsk-u22koy`. The stamp reads **2025** for 2026 work. Every other fragment is `tsk-<id>-<short-description>.md` with no date.

### 5. Trailing whitespace on three added lines.

## Verification I ran

```
trial merge into current master:  clean, ZERO deletions, only the 2 intended files
full suite on the trial merge:    1648 passed, 12 skipped
```

Two notes on that suite number, since neither is obvious. Master had already moved to `53bde752` and #372 changed this very file, so the base here is stale; the merge is clean and reverts nothing, which only a trial merge could establish. And this PR adds **no new test functions**, so the count is unchanged by construction. **A stable suite count is not evidence that anything here works**, which is the whole reason both findings above came from probes instead.

PR #373, opened 2026-08-18
Revise PR #371: the two block assertions are correct and proven, the other 17 lines are dead or self-referential

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #371: the two block assertions are correct and proven, the other 17 lines are dead or self-referential

Autonomous build of board card tsk-tsoylj.

Keep the two SCRIPT-constant assertions in
test_system_crontab_block_names_usr_bin_python3 that prove the emitted
cron block is independent of _HELPER_PATH. Drop the dead a2a_send mock
and the self-referential initial_crontab assertion from the subprocess
test, and add a changelog fragment for the path-sensitivity guard.

Files:
 STATUS.md                                             | 8 +++++++-
 changelog.d/tsk-u22koy-path-sensitivity-assertions.md | 2 ++
 tests/test_resume_arm_time.py                         | 3 +++
 3 files changed, 12 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

- **Tests**
  - Strengthened coverage for system cron entries to verify that commands consistently use the exact `/usr/bin/python3` path.
  - Added checks ensuring cron block names and retry commands remain correct regardless of helper path configuration.

- **Documentation**
  - Added a changelog entry documenting the improved path-sensitivity assertions.

### DIFF
--- a/changelog.d/tsk-u22koy-path-sensitivity-assertions.md
+++ b/changelog.d/tsk-u22koy-path-sensitivity-assertions.md
@@ -1,0 +1,2 @@
+### Fixed
+- `tests/test_resume_arm_time.py` adds two path-sensitivity assertions to `test_system_crontab_block_names_usr_bin_python3` that verify the emitted cron block names `/usr/bin/python3` with the `SCRIPT` constant, independent of `_HELPER_PATH`, so a mutated helper path cannot silently pass the test.
--- a/tests/test_resume_arm_time.py
+++ b/tests/test_resume_arm_time.py
@@ -147,6 +147,9 @@ def test_system_crontab_block_names_usr_bin_python3():
     block = resume_arm_time.system_crontab_block(fire, retry)
     assert f"/usr/bin/python3 {resume_arm_time._HELPER_PATH} --fire primary" in block
     assert f"/usr/bin/python3 {resume_arm_time._HELPER_PATH} --fire retry" in block
+    # Independent check against SCRIPT constant (proves path sensitivity)
+    assert f"/usr/bin/python3 {SCRIPT} --fire primary" in block
+    assert f"/usr/bin/python3 {SCRIPT} --fire retry" in block
     # No bare `python3` token (the blocked PR emitted `python3 <path>`).
     assert " python3 " not in block
 


jaylfc:
MERGE. Every item on `tsk-tsoylj` is met and the coverage claim is proven rather than asserted.

## The assertions discriminate, which is the whole point of the card

Mutant, `_HELPER_PATH` hardcoded back to the pre-#354 `/home/jay/.taos-fleet-tools/resume_arm_time.py`, mutant confirmed engaged before each run:

| tree | result |
|------|--------|
| unmutated source, this PR's tests (positive control) | 1 passed |
| mutant, **this PR's** tests | **1 failed** |
| mutant, **master's** tests | **1 passed** |

That is the pair the card demanded. `SCRIPT` is `REPO / "scripts" / "resume_arm_time.py"`, derived from the test file's own location, so it is genuinely independent of the module constant, and the assertion has different values on each side. The two pre-existing assertions put `_HELPER_PATH` on both sides and therefore hold for any value it takes.

## One correction to my own card, in the PR's favour

I wrote that the existing assertions were self-referential and left it there, which understates what they still do. Breaking only the primary cron line, and then only the retry line, both mutants engaged and content-matched rather than addressed by line number:

| mutant | this PR's tests | master's tests |
|--------|-----------------|----------------|
| only the primary cron line's helper diverges | 1 failed | 1 failed |
| only the retry cron line's helper diverges | 1 failed | 1 failed |

So master already catches a per-line divergence. The blind spot is narrower and more precise than the card said: it is specifically a globally wrong `_HELPER_PATH`, where both sides of the existing comparison move together and the test cannot see it. That is exactly the #354 defect class, and it is the case these two lines close. Worth stating because the card's framing would have credited the change for more than it does, while the thing it actually fixes is the one that shipped.

Observation, not a blocker: under every mutant built here the two added lines are only jointly load-bearing, since both cron lines interpolate the same `helper` variable and any mutant on it breaks both. Deleting either one alone changes no result. They mirror the two pre-existing assertions symmetrically and cost nothing, so keeping both is the right call, but the pair is one arm of coverage rather than two.

## Card items

- Branched from current master; the trial merge against `c0bf7a22` applies cleanly.
- The two assertions are kept and unchanged.
- The dead `a2a_send` mock and the self-referential `in initial_crontab` assertion are absent. Branching from master rather than from `exec/tsk-u22koy` means there was nothing to delete, and the end state is what the card specified.
- Fragment renamed to `changelog.d/tsk-u22koy-path-sensitivity-assertions.md`, matching the sibling convention with no date stamp, describing only what this PR changes, and its last byte is `0x0a`.
- No trailing whitespace on any added line.

## Verification

Trial merge against current master: **2 files changed, 5 insertions, 0 deletions**. Zero deletions is the proof nothing on master was reverted, which the branch diff alone could not show.

Full suite on the trial merge: **1661 passed, 12 skipped**, identical to the master baseline. That is expected and is not evidence of anything here, because this PR adds assertions inside an existing test and collects no new test functions. The mutant table above is the evidence; the count is not.

Gates on the trial merge: no conflict markers, `check_deleted_symbols.py --base origin/master` clean, `normalise_handle_gate.py` clean, `check_witness_token.py` clean, `ast.parse` clean on the touched test file, no em dashes in the fragment.

The trailing-newline gate merged in #374 half an hour ago also ran against this PR and reports clean. This is the first PR it has judged, and the fragment it was pointed at is the seventh in the sequence that produced six unterminated ones. The tree was verified unmodified after the suite run, so the result is not a contaminated one.

PR #375, opened 2026-08-18
ci(lint): configure ruff, declare it, run it in CI, fix the docs invocation, and clear the 35 findings it exposes (tsk-rhuge6)

jaylfc:
Executes `tsk-rhuge6`, which combines the revisions of blocked PRs #362 (the config) and #367 (the docs). Two commits so the review surfaces stay separate: the lint findings first, then the infrastructure that exposes them.

## The card's premise held, and one thing it did not anticipate

Re-measured rather than inherited. Ruff was declared nowhere, no workflow referenced it, and the bare `ruff check` form resolves on this box to `/home/jay/.local/bin/ruff`, a machine-local install in no manifest. `uv sync` in this branch now installs `ruff 0.16.3` to the worktree's own `.venv`, so the CI step resolves to a repo-local binary rather than that one.

What the card did not anticipate: **turning the config on made the tree red with 35 findings**, so there was no version of this PR where the lint step lands green without addressing them. The card forbids widening the config or adding ignores to hide findings, so they are fixed. That is commit 1.

**Six of the 35 were not dead code at all.** They are in `taosmd/__init__.py`, and a blind `--fix` would have deleted public API:

| symbol | why it is there |
|--------|-----------------|
| `config`, `loaders`, `predicate_vocab`, `emem_event_lift` | imported so `taosmd.<name>` attribute access works; referenced across the tree |
| `list_pending_decisions`, `resolve_pending_decision` | added to the api import block and never declared in `__all__` |

They are added to `__all__` rather than deleted, which is the remedy ruff itself suggests, and it also puts them under `check_deleted_symbols.py`, since that gate only tracks what `__all__` names. This is the exact defect class that gate exists for, so the order of work mattered: the declarations went in **before** `--fix` ran, so autofix could not remove them.

Two of the six dead bindings have right-hand sides with side effects or that can raise, so the call is preserved and only the binding is renamed with a leading underscore: `_pkg_files(...).joinpath(...)` in `cli.py`, which sits inside a `try/except`, and `_run_systemctl(...)` in `service_install.py`, whose own comment explains a non-zero result is tolerated. Deleting those lines would have removed behaviour, not dead weight.

One finding looked like a live bug and is not. `resolved_thread = thread` in `service.py` is assigned under a comment about alias resolution and never read, but the alias feature works through `alias_sources`, which is used correctly. It is a vestigial binding from an earlier approach, so removing it changes nothing.

## The lint step engages, proven red and green

A lint step only ever observed green is indistinguishable from one that is not running, which is the whole defect this card is about. Using the CI step's exact command:

```
uv run ruff check taosmd/          ->  All checks passed!        exit=0
  (one deliberate dead import added, insertion asserted present)
uv run ruff check taosmd/          ->  F401 `math` imported but unused
                                       Found 1 error.            exit=1
  (probe removed)
uv run ruff check taosmd/          ->  All checks passed!        exit=0
```

## The docs, and correcting #367's diagnosis

#367 rewrote the five files to a bare `ruff check` on the stated grounds that the module form was broken. That diagnosis is wrong and it had shipped as a changelog claim. With ruff installed, `python -m ruff check` and `ruff check` produce identical output and identical exit codes; the module form failed only because ruff was installed nowhere, which the dev dependency now fixes.

All five files use `uv run ruff check`, matching how `ci.yml` already invokes `uv run pytest` and `uv run python -m compileall`. Surrounding three-space continuation indentation is left alone. #367's sentence `Do not invent a substitute tool.` is kept; its companion instruction to skip the check when the tool is missing is dropped, since a missing binary now means a broken environment worth reporting. The changelog fragment does not repeat the false claim.

## Verification

- `git rev-list --count HEAD..origin/master` reads **0**.
- Acceptance 1: `git grep -n 'python3 -m ruff\|python -m ruff' docs/` returns nothing; `git grep -ln 'uv run ruff' docs/` returns **5** files.
- Full suite: **1661 passed, 12 skipped**, identical before and after the lint fixes, so the 35 removals changed no behaviour. Note the card's stated baseline of 1617 is stale; master is 1661 as of `80efb5cf`.
- `uv run python -m compileall -q taosmd/` clean, and `pyproject.toml` and `ci.yml` both parse.
- Gates on the branch: `check_deleted_symbols.py --base origin/master` clean, which is the one that matters most here, plus `normalise_handle_gate.py`, `check_witness_token.py` and the new `check_trailing_newline.py` all clean. No conflict markers.
- `import taosmd` asserts all six re-exports are present as attributes **and** declared in `__all__`, which is now 80 entries.
- Fragment ends `0x0a`.

## Two observations, neither blocking

`.venv` is not in the repository's `.gitignore`. It is invisible to `git status` only because uv writes a nested `.venv/.gitignore` containing `*`. That is uv's behaviour rather than the repo's, so any tool that populates a virtualenv differently would leave it stageable. Worth a one-line ignore entry at some point.

The lint step is scoped to `taosmd/`, matching the existing compile-check step. `tests/` has 68 findings and `benchmarks/` 77, none of them addressed here. Widening the scope is a separate piece of work and should be its own card rather than folded into this one; flagged rather than silently left out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,5 +17,7 @@ jobs:
         run: uv sync --python 3.12
       - name: Compile check (two merged branches previously shipped files that do not parse)
         run: uv run python -m compileall -q taosmd/
+      - name: Lint (pyflakes via ruff; its own step so a lint failure is not read as a test failure)
+        run: uv run ruff check taosmd/
       - name: Tests
         run: uv run pytest tests/ -q
--- a/changelog.d/tsk-rhuge6-ruff-lint-gate.md
+++ b/changelog.d/tsk-rhuge6-ruff-lint-gate.md
@@ -1,0 +1,6 @@
+### Added
+- Ruff is configured for pyflakes rules (`[tool.ruff.lint] select = ["F"]`), declared as a pinned dev dependency, and run as its own `ci.yml` step against `taosmd/`, so a dead import or dead local binding fails the job instead of accumulating.
+
+### Fixed
+- Cleared the 35 pyflakes findings that configuration exposed: 28 unused imports, 6 dead local bindings and one redefinition. Six apparent findings in `taosmd/__init__.py` were deliberate re-exports and are now declared in `__all__` rather than deleted.
+- The five `docs/agent-jobs/` files now say `uv run ruff check`, matching how `ci.yml` invokes everything else. The previous `python3 -m ruff check` form was never broken; it failed only because ruff was not installed anywhere, which the dev dependency above fixes.
--- a/docs/agent-jobs/DONE-job-005-collections-db-connect.md
+++ b/docs/agent-jobs/DONE-job-005-collections-db-connect.md
@@ -106,7 +106,7 @@ from . import _db
    `grep -n "sqlite3\." taosmd/collections.py` that other uses remain.
 7. `python3 -m pytest tests/test_collections_db_connect.py -q` (2 passed),
    then the FULL suite `python3 -m pytest -q -m "not slow"`, then
-   `python3 -m ruff check taosmd/collections.py tests/test_collections_db_connect.py`.
+   `uv run ruff check taosmd/collections.py tests/test_collections_db_connect.py`.
 8. One commit, push, open the PR. PR body: the inconsistency, the one-line
    fix, confirmation that nothing in the store depended on the journal mode,
    and the test count. Reference issue #202. Do not merge.
--- a/docs/agent-jobs/README.md
+++ b/docs/agent-jobs/README.md
@@ -24,7 +24,9 @@ Do not improvise.
 7. The full test suite must pass before you push:
    `python3 -m pytest -q -m "not slow"` (well over a thousand tests, all
    green; record the exact number in your PR body). Lint the files you
-   changed: `python3 -m ruff check <files>`.
+   changed: `uv run ruff check <files>`. Do not invent a substitute tool. If
+   that command cannot run, your environment is broken: report it and stop,
+   rather than working around it.
 8. Never commit IP addresses, hostnames, tokens, or credentials.
 9. STOP on any surprise: a failing test you did not cause, a merge conflict,
    a file that does not look like the job describes. Open the PR with what
--- a/docs/agent-jobs/job-002-cross-encoder-path-fix.md
+++ b/docs/agent-jobs/job-002-cross-encoder-path-fix.md
@@ -103,7 +103,7 @@ def _default_onnx_path() -> str:
 7. `python3 -m pytest tests/test_cross_encoder_path.py -q` (3 passed), then
    the FULL suite `python3 -m pytest -q -m "not slow"` (all green; the
    existing reranker tests prove explicit-path behavior is unchanged), then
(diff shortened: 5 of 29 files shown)


jaylfc:
Merged as `4253c661`. Master is at 1661 passed / 12 skipped, unchanged from before the sweep.

This was held rather than self-merged because a 29 file lint sweep that declares six names into the public `__all__` is a wide blast radius for a single reader. It has now had two, so recording what each one actually established.

### What I checked on top of the card

**A trial merge against current master, not the PR diff.** A diff is computed against its base, so it cannot show a revert. The trial merge came out at exactly the PR's own 44 insertions / 34 deletions with no conflict markers, which is the cheap proof nothing on master was quietly rolled back.

**Every removed module level import, checked for cross module attribute use.** This is the twin of the finding that shaped the PR. Six pyflakes hits in `taosmd/__init__.py` were deliberate re exports and a blind autofix would have deleted them. The same trap exists one level down: any module level name removed from another file is also reachable as `module.name`, including string patch targets like `patch("taosmd.mod.name")`, and `scripts/check_deleted_symbols.py` will not catch it because that gate only tracks what `__all__` names. I swept all 17 removed module level names. Zero cross module uses. The scan came back empty, which announces itself as a pass, so I ran positive controls on the same pattern in the same command: `agents.is_task_enabled` 13 hits, `predicate_vocab.validate` 1, `agents.run_if_enabled` 1. The probe engages, so the empty result is a result.

**The new lint gate has both properties.** A gate that can detect a defect but that nothing invokes is not a gate, and a passing suite only ever shows the first property. Detection: a planted dead import in `taosmd/tasks.py` makes `ruff check taosmd/` exit 1, and the clean tree exits 0. Invocation: the step sits inside ci.yml's `test` job on both `pull_request` and `push: master`, and that job ran. My first attempt at the detection probe read `$?` through a `tail` pipe and reported exit 0 for a tree ruff had actually rejected. That number was not a result and the measurement above is the re run.

### Kilo found two real defects and both are fixed

Both were stale comments, neither changed behaviour, and both were genuine rather than noise.

`service_install.py` still named `rc` after the binding became `_rc`. Reworded to describe the discarded return code without naming a variable, so the same rename cannot make it stale a second time.

`http_server.py` said its suppression covered "the import above", but that import is the one this PR removed. The line is not redundant though: deleting `_ = ref` makes ruff exit 1 on `ref` at line 211, so the comment was wrong about its target rather than the line being dead. Proven by deleting it and watching the gate go red, not by reading it.

### Correcting my own card

`tsk-rhuge6` stated a 1617 baseline. It was 1661 by the time it ran. A card's own measurements go stale like any other document, and reporting a delta against the number in the card would have been a false delta.

### Two follow ups filed rather than folded in

`tsk-qb3bo4`: `_webui_dir()` in `http_server.py` has a try block whose entire body is computed and discarded, under a docstring that claims importlib.resources is consulted first. It never is. Pyflakes cannot see it because `_ = ref` launders it. Removing the block changes what happens when package metadata is absent, so it is a behaviour question and did not belong in a sweep that must not alter behaviour.

`tsk-hltkf7`: `tests/` holds 68 pyflakes findings and `benchmarks/` 77. The CI invocation is scoped to `taosmd/` on purpose. Clearing 145 more findings here would have made the source changes unreviewable.

PR #376, opened 2026-08-18
http_server._webui_dir has a try-block whose whole body is computed and discarded, and a docstring that describes it as load-bearing

jaylfc:
CARD TITLE (intent, not commit subject): http_server._webui_dir has a try-block whose whole body is computed and discarded, and a docstring that describes it as load-bearing

Autonomous build of board card tsk-qb3bo4.

The try-block computed _pkg_files("taosmd").joinpath("webui") and
discarded the result, always falling through to the __file__-relative
candidate. __file__-relative resolution works for both wheel and source
installs because webui is package data, so the dead block and its
load-bearing _ = ref dummy assignment are removed. The docstring is
corrected to match the actual behaviour. A unit test proves the
function resolves via __file__-relative path.

Files:
 changelog.d/tsk-qb3bo4-dead-importlib-block.md |  8 ++++++++
 taosmd/http_server.py                          | 15 ++-------------
 tests/test_http_server.py                      | 17 +++++++++++++++++
 3 files changed, 27 insertions(+), 13 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resolution of the web interface directory relative to the application’s installed location.
  * Removed an unused fallback path that could cause confusion in web interface loading.

* **Tests**
  * Added regression coverage to verify reliable web interface path resolution.


jaylfc:
Merged as `b036af07`. Master is at 1662 passed / 12 skipped, one higher than before, which is the new test.

The change itself is right and the lane read the card correctly: the try block was unreachable in effect, `_pkg_files` was used at exactly one line and that line is inside the block, and nothing else in the tree reaches `http_server._pkg_files`, so dropping the import with it is correct. The docstring and the code now agree, which was the actual point of the card.

### It did not merge, and the reason is worth recording

The branch was cut from `80efb5cf`, before #375 landed. #375 reworded a comment INSIDE the block this branch deletes wholesale, so the two sides conflicted over the text of a line that one of them was removing entirely. Both sides agreed the block goes.

**Taking git's default resolution here would have kept master's side and silently restored the dead block, while reporting a successful merge.** I hit this for real: my first resolution pass ran the same regex against the reversed merge direction, kept the wrong side, and produced a tree where the block was still present. The only thing that caught it was diffing the result against the exact file I had already tested, rather than reading the resolved file and judging it. A merge succeeding says nothing about which side won, which is the same family as a diff being unable to show you a revert.

So the branch now carries a merge of current master with that conflict resolved toward the deletion, and the tree that CI ran is byte-identical to the one I tested.

### What was verified

**The new test discriminates**, which a test asserting unchanged behaviour is not guaranteed to do. Mutated `Path(__file__).parent / "webui"` to `.parent.parent`, content-matched rather than addressed by line number, with the target asserted present exactly once before writing: 1 passed, then 1 failed, then 1 passed restored.

**No behaviour change.** The deleted call sat inside `try/except Exception: pass`, so removing a call that could only raise into a swallow is not observable. `_WEBUI_DIR` and its two test monkeypatches are untouched.

Full suite on the resolved tree 1662 passed / 12 skipped, all four gates clean, and `ruff check taosmd/` clean, which is the gate #375 landed two commits earlier.

PR #377, opened 2026-08-18
uninstall_systemd documents a non-zero error return it can never produce, and swallows both systemctl results

jaylfc:
CARD TITLE (intent, not commit subject): uninstall_systemd documents a non-zero error return it can never produce, and swallows both systemctl results

Autonomous build of board card tsk-23zx64.

The function previously discarded the daemon-reload result and let unlink()
exceptions propagate, while its docstring promised a non-zero return on
error. Catch OSError from unlink() and return 1, and return the
daemon-reload exit code when it is non-zero. The disable --now result
remains deliberately discarded. Added unit tests for the two new failure
paths and the missing-unit no-op path.

Files:
 changelog.d/tsk-23zx64-systemd-uninstall-return.md |  6 ++
 taosmd/service_install.py                          |  9 ++-
 tests/test_service_install.py                      | 82 ++++++++++++++++++++++
 3 files changed, 95 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/tsk-23zx64-systemd-uninstall-return.md
+++ b/changelog.d/tsk-23zx64-systemd-uninstall-return.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- `uninstall_systemd()` in `taosmd/service_install.py` now returns non-zero when
+  `daemon-reload` fails or the unit file cannot be removed, instead of always
+  returning 0. Added unit tests for both failure paths and the missing-unit
+  no-op path.
--- a/taosmd/service_install.py
+++ b/taosmd/service_install.py
@@ -130,9 +130,14 @@ def uninstall_systemd() -> int:
     # was never enabled, which is not an error for an uninstall.
     unit_path = Path.home() / ".config" / "systemd" / "user" / _SYSTEMD_UNIT_NAME
     if unit_path.exists():
-        unit_path.unlink()
+        try:
+            unit_path.unlink()
+        except OSError:
+            return 1
         print(f"Removed unit file: {unit_path}")
-        _run_systemctl(["daemon-reload"])
+        rc = _run_systemctl(["daemon-reload"])
+        if rc != 0:
+            return rc
     else:
         print("Unit file not found, nothing to remove.")
     print("taosmd service stopped and uninstalled.")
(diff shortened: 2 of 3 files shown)


jaylfc:
Reviewed and merging. The card's complaint is real and this closes it: `uninstall_systemd` promised "non-zero on error" in its own docstring while a single `return 0` was the only exit it had.

What I checked rather than read.

**Both new guards are load-bearing, proven one mutant at a time.** Reverting the `daemon-reload` rc check fails exactly `test_uninstall_systemd_returns_nonzero_on_daemon_reload_failure` and nothing else. Reverting the unlink `try/except` fails exactly `test_uninstall_systemd_returns_nonzero_on_unlink_failure` and nothing else. One mutant, one failure, each aimed at its own branch rather than at the feature, because killing a whole guard and seeing one failure is the shape that has read as coverage here before.

**The no-op path is pinned in the direction that matters.** `test_uninstall_systemd_missing_unit_file` asserts `calls == [["disable", "--now", "taosmd.service"]]`, so it holds that `daemon-reload` is NOT called when there is no unit file, rather than only asserting the return code. It also spells the unit name out literally instead of reading `_SYSTEMD_UNIT_NAME` back out of the module under test, so the two sides cannot move together and the assertion can actually fail.

**The scope call on the first `systemctl` is right.** The card said both results were swallowed, and `disable --now` still discards its rc, but that one is deliberate and the comment says why: it is non-zero when the unit was never enabled, which is not an error for an uninstall. Turning that into a failure would break the ordinary path. The binding stays as `_rc`, matching the convention #375 established for keeping a side-effecting call while marking the value unused.

**Numbers.** 1666 passed, 12 skipped on the trial merge against a 1662 baseline, which is exactly the four tests added here. All four repo gates clean plus `ruff check taosmd/`, exit codes read directly. The merge brought in only `STATUS.md`, and master has not touched `service_install.py` since this branch was cut, but I grepped both guards in the merged tree anyway rather than inferring their survival from a clean merge.

**The changelog fragment ends in `0x0a`.** Seven fragments in a row shipped without a terminating byte before the gate from #374 landed; this is the first one merging with that gate live, and it is clean.

One ordering note for #378, which extends the lint gate to `tests/`: it was cut before this branch, so the 82 lines of test code added here have never been seen by the wider ruff scope. #378 needs re-measuring against master after this merge, not against the tree it was cut from.

PR #379, opened 2026-08-18
Revise PR #378: the lint gate is widened to tests/ AND benchmarks/ but only tests/ was partly cleared, so the step goes red on merge

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #378: the lint gate is widened to tests/ AND benchmarks/ but only tests/ was partly cleared, so the step goes red on merge

Autonomous build of board card tsk-d4jvo2.

Files:
 tests/test_ref_fetch.py                    |  6 ------
 tests/test_registry_auth.py                |  2 +-
 tests/test_remote.py                       |  1 -
 tests/test_service_install.py              |  4 ----
 tests/test_session_catalog.py              |  2 --
 tests/test_tasks.py                        |  3 +--
 tests/test_vector_supersede.py             |  1 -
 50 files changed, 76 insertions(+), 138 deletions(-)

## Summary by CodeRabbit

* **Chores**
  * Expanded automated lint checks to cover application code, tests, and benchmarks.
  * Removed unused imports, variables, and assignments across benchmark and test tooling.
  * Simplified unchanged console output formatting while preserving displayed results.

* **Documentation**
  * Added changelog coverage for the expanded lint gate and successful cleanup.

* **Tests**
  * Maintained existing test behavior and coverage while improving lint cleanliness.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,6 @@ jobs:
       - name: Compile check (two merged branches previously shipped files that do not parse)
         run: uv run python -m compileall -q taosmd/
       - name: Lint (pyflakes via ruff; its own step so a lint failure is not read as a test failure)
-        run: uv run ruff check taosmd/
+        run: uv run ruff check taosmd/ tests/ benchmarks/
       - name: Tests
         run: uv run pytest tests/ -q
--- a/benchmarks/beam_feasibility_probe.py
+++ b/benchmarks/beam_feasibility_probe.py
@@ -94,7 +94,7 @@ async def main():
     print("  --- extrapolation (one conversation; a full BEAM run is 100 of these) ---")
     print(f"    this split full ingest (x100 convs): ~{embed_s*100/60:.1f} min embed")
     if SPLIT == "100K":
-        print(f"    1M tier is ~10x heavier per conversation; 10M ~100x")
+        print("    1M tier is ~10x heavier per conversation; 10M ~100x")
     print(f"  VERDICT INPUTS: tokens/conv={est_tokens}, embed_s/conv={embed_s:.1f}, RSS={rss_mb:.0f}MB")
 
 
--- a/benchmarks/claims_gate_probe.py
+++ b/benchmarks/claims_gate_probe.py
@@ -52,7 +52,6 @@ import json
 import os
 import sys
 import tempfile
-import time
 from pathlib import Path
 
 import httpx
--- a/benchmarks/combo_benchmark.py
+++ b/benchmarks/combo_benchmark.py
@@ -22,17 +22,16 @@ import os
 import sys
 import tempfile
 import time
-from itertools import product
 
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 import httpx
 
-from taosmd import VectorMemory, KnowledgeGraph, Archive, SessionCatalog
+from taosmd import VectorMemory, KnowledgeGraph, Archive
 from taosmd.memory_extractor import extract_facts_from_text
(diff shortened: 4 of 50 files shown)


jaylfc:
## Reviewed: APPROVED, merging

This is the correct revision of #378. That PR widened the lint step to `tests/` AND `benchmarks/` but cleared only nine files under `tests/` and zero under `benchmarks/`, so it would have gone red on the first push. This one clears both trees.

### Acceptance, measured with the literal ci.yml command and the rc captured directly

Never via pytest: lint and tests are separate CI steps, so a green suite says nothing about this step.

| tree | `uv run ruff check taosmd/ tests/ benchmarks/` | rc |
|---|---|---|
| current master (positive control) | 140 errors | **1** |
| this PR trial-merged onto master | All checks passed | **0** |

Version parity checked first: CI runs `uv run ruff` and `ruff==0.16.3` is pinned in pyproject, local resolves to 0.16.3.

Note the control reads **140**, not the 127 I measured against #378 yesterday. That figure moves every time a test file lands, so it was never carryable. Re-measured rather than reused.

### Both new path arguments proven load-bearing

A widened gate has to actually detect in the trees it was widened to. Mutants, one at a time, restored after each:

- `import zoneinfo` appended to `benchmarks/locomo_runner.py`: widened command rc=**1**, old `ruff check taosmd/` rc=**0**. The widening is exactly what catches it.
- `import os` appended to `tests/test_api.py`: widened command rc=**1**.
- Tree restored: rc=**0**.

My first `benchmarks/` mutant was invalid and I nearly recorded a false negative from it. I appended `import os` to a file that already imports and uses `os`, so under `select = ["F"]` nothing fired and the run read rc=0, which looks identical to a blind gate. The novel-import mutant disambiguated it.

### Other checks

- Trial merge onto `f09516ad` clean, no conflict markers.
- `check_deleted_symbols.py --base origin/master`: `deleted-symbols-guard: clean` (output read, not the exit code). Structurally safer than the guard's word: **zero files under `taosmd/` are touched**, so no package API is in scope at all.
- normalise-handle, witness-token, trailing-newline gates all clean.
- Suite on the trial merge: **1666 passed, 12 skipped**, which is the measured master baseline. This PR adds no tests, so no change is the right result.

### What I am NOT holding this for

An autofix that deletes an unused binding leaves the discarded work behind, and in three places here that exposes dead code that was already dead on master. Every instance is behaviour-preserving, so none of it blocks the gate widening, but none of it should be lost either. Carded rather than waved through:

- **`tsk-jp7we4`** covers the benchmark leftovers, including both bot findings on this PR. `benchmarks/longmemeval_matrix.py:133` (coderabbit, Major): the `expand_from_results` result was already discarded on master, and the comment directly beneath says that is deliberate, so `v02_graph` measures expansion cost at unchanged recall. Verified no `v02_graph` number is published in `docs/benchmarks.md` or the README, so no published claim depends on it. `benchmarks/recall_v2_benchmark.py:192` was flagged independently by both coderabbit and kilo and is a bare `item["question"]` subscript left by the fix.
- **`tsk-7a6brx`** covers a worse one this sweep led me to. `tests/test_ref_fetch.py::test_local_path_passes_data_dir_to_ref_fetch` monkeypatches a fake fetch, then stops. It never calls the code under test and contains no assertion, so it passes unconditionally. The deleted line here was its last dead statement, which is a correct F841 fix. Proof it checks nothing: adding `assert captured["data_dir"] is not None` makes it fail with `KeyError: 'data_dir'`. Confirmed vacuous on master too, so it predates this PR. An AST scan of all 31 test files this sweep touched found exactly three assertion-free tests; the other two are legitimate does-not-raise tests that do call their subject.

PR #392, opened 2026-08-24
Revise PR #385: A2A per-channel ACLs enforced from an unverified JWT decode and a client-supplied from field

jaylfc:
CARD TITLE (intent, not commit subject): Revise PR #385: A2A per-channel ACLs enforced from an unverified JWT decode and a client-supplied from field

Autonomous build of board card tsk-fd4m5t.

Per-channel read/post allowlists for the A2A bus. Enforcement derives the
acting principal from the verified registry token sub via _registry_verifier
authorize, never from a client-supplied body field. A forged token whose
signature fails verification is rejected even when its sub matches the
allowlist. Admin endpoints POST /a2a/admin/set-channel-acl and
GET /a2a/admin/channel-acl manage per-channel ACLs. Channels without an
explicit ACL entry remain open, preserving deploy-neutral default.

RED proof (tests/test_a2a_channel_acls.py against unpatched code):
- test_forged_jwt_with_matching_sub_is_rejected_by_acl: FAILED (set_acl missing)
- test_spoofed_from_is_rejected_by_acl: FAILED (set_acl missing)
- test_no_acl_entry_allows_post: passed (no ACL means open, unchanged)

All 1669 tests pass after fix.

_MERGE_BASE=d5a7d6c0d4b54e6eefbc2b4075e02f0213e9fe11 || _MERGE_BASE=d5a7d6c0d4b54e6eefbc2b4075e02f0213e9fe11
Files:

### DIFF
--- a/changelog.d/tsk-fd4m5t-a2a-channel-acls.md
+++ b/changelog.d/tsk-fd4m5t-a2a-channel-acls.md
@@ -1,0 +1,9 @@
+### Added
+
+- Per-channel read/post ACLs for the A2A bus. Enforcement derives the acting
+  principal from the verified registry token `sub` (via `_registry_verifier.
+  authorize`), never from a client-supplied body field. A forged token whose
+  signature fails verification is rejected even when its `sub` is in the
+  allowlist. Admin endpoints `POST /a2a/admin/set-channel-acl` and
+  `GET /a2a/admin/channel-acl` manage per-channel ACLs. Channels without an
+  explicit ACL entry remain open (zero deploy behavior change).
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -66,6 +66,8 @@ _HUMAN_PRINCIPAL_IDS_KEY = "human_principal_ids"
 # one of these directories. Empty (the default) means collections are off.
 _COLLECTIONS_SECTION_KEY = "collections"
 _COLLECTIONS_ALLOWED_ROOTS_KEY = "allowed_roots"
+# Per-channel ACL for the A2A bus.
+_ACL_SECTION_KEY = "acls"
 
 MANAGED_BY_STANDALONE = "standalone"
 MANAGED_BY_TAOS = "taos"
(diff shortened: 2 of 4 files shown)


jaylfc:
**Verdict: BLOCK.** Revision card `tsk-kslfxt` is filed and claimable.

First, the part that is right, because it is the part you were asked to fix and you fixed it: **the identity path is correct.** `_enforce_channel_acl` resolves the caller through `self._get_authenticated_agent_id()`, which verifies via the registry verifier and uses the token's own VERIFIED `sub`. There is no unsigned `jwt.decode` and no trust in the `from` body field. That was the whole defect in #385 and it is genuinely gone. Keep this code.

The block is about **scope, not identity**. The read ACL guards two handlers, and five other read paths reach the same rows without it.

I measured this on a trial merge of head `d1a1a6c4` onto master `9de5bec1`, with a **positive control in the same pytest run**: `GET /a2a/messages?thread=secret` correctly returned 403. So the ACL is genuinely installed and enforcing, and what follows are real bypasses rather than a feature that never ran.

**Leak 1 (critical).** `GET /a2a/messages` with no `thread` parameter returns the restricted channel's full message bodies, 200, unauthenticated:

```
LEAK via unfiltered feed: status=200 body={"messages": [{"id": 1, ... "body": "TOPSECRET-CANARY", "thread": "secret", ...}]}
```

`http_server.py:1762` reads `if thread and not self._enforce_channel_acl(thread, "read")`, so omitting the parameter skips the check, and `service.a2a_feed(thread=None)` spans every thread. This is the path the fleet actually sweeps the bus with (`GET /a2a/messages?limit=200`), so the ACL is bypassed by the single most-used read in the system.

**Leak 2 (critical).** `GET /a2a/admin/channel-acl?channel=secret` returns 200 and discloses the allowlist with no admin token:

```
ACL disclosed without admin token: status=200 body={"channel": "secret", "acl": {"read": ["agent-allowed"], "post": ["*"]}}
```

`_handle_admin_a2a_get_channel_acl` never calls `_check_admin_token()`. You did add the path to the `_is_admin_route` tuple, but that helper returns False for every method except POST (plus one DELETE case), so for a GET the entry is unreachable. It looks installed and is inert. Fix it in the handler; the route table cannot gate a GET.

**Leak 3 (high).** `GET /a2a/threads` returns the restricted channel's name, its participants, and a `body_preview` of its latest message, unauthenticated.

**Code-read, not yet measured:** `_handle_a2a_stream` (`GET /a2a/stream`, SSE) takes an optional `thread` and contains no ACL call, so the live firehose has the same shape as leak 1. Treat it as a fourth site and prove it.

**Not a leak:** `GET /a2a/mentions` was probed and did not leak the canary. Leave it alone.

On tests: the PR adds three, and all three are about `post`. There is no test for the `read` permission anywhere, which is exactly the half that is bypassed. The file's own `_get_messages` helper always passes `?thread=`, so the vulnerable call shape is never exercised. That is how a real feature ships with a hole and a green suite.

One design note for the revision, worth deciding explicitly rather than by accident: an unfiltered feed cannot answer "is this caller allowed" per row with a single channel check. Filtering the result set down to the channels the verified identity may read is probably better than 403-ing the whole unfiltered feed, because a blanket 403 breaks every existing sweeper the moment any one channel gets an ACL. Whichever you pick, say so in the PR body.

Start the revision from a **fresh branch off master**, not from `exec/tsk-fd4m5t`, and carry the identity path and the three post-ACL tests across by hand. Full acceptance criteria, including the red-proof requirement, are on `tsk-kslfxt`. Reviewer baseline on master `9de5bec1` is 1691 passed / 12 skipped, re-measured today; re-measure rather than inherit it.

PR #395, opened 2026-08-24
A2A v2 slice 2b: acks as server state, not traffic (service layer only, spec section 3)

jaylfc:
CARD TITLE (intent, not commit subject): A2A v2 slice 2b: acks as server state, not traffic (service layer only, spec section 3)

Autonomous build of board card tsk-ta4ftl.

a2a_ack(message_id, by, *, data_dir=None) records an acknowledgement as
server-side state on the message envelope -- an acked_by list persisted via
archive.update_event_data_json -- rather than as a new bus message. The
acked_by field is surfaced on the envelope by the existing a2a_feed and
a2a_thread_messages read paths (omitted when absent, no null noise).
Acks are idempotent per principal: a second ack by the same principal
leaves exactly one entry.

The composed "unhandled for X" query (mentions past X's cursor minus
acks) is deferred to slice 2c, which needs 2a's server-side cursor.

Proof (RED-first): tests/test_a2a_ack_state.py
  RED:  5 failed, exit 1 (AttributeError: module 'taosmd.service' has no attribute 'a2a_ack')
  GREEN: 5 passed, exit 0

Refs: tsk-ta4ftl, docs/specs/a2a-delivery-v2.md (section 3)

Files:
 changelog.d/tsk-ta4ftl-a2a-ack-state.md |   3 +
 taosmd/service.py                       |  43 ++++++++-
 tests/test_a2a_ack_state.py             | 162 ++++++++++++++++++++++++++++++++
 3 files changed, 207 insertions(+), 1 deletion(-)

### DIFF
--- a/changelog.d/tsk-ta4ftl-a2a-ack-state.md
+++ b/changelog.d/tsk-ta4ftl-a2a-ack-state.md
@@ -1,0 +1,3 @@
+### Added
+
+- `a2a_ack(message_id, by, *, data_dir=None)` records an acknowledgement as server-side state on the message envelope (`acked_by` list via `archive.update_event_data_json`), never as a new bus message. Acks are surfaced by the existing `a2a_feed` and `a2a_thread_messages` read paths and are idempotent per principal.
--- a/taosmd/service.py
+++ b/taosmd/service.py
@@ -605,6 +605,8 @@ async def a2a_feed(
             msg["refs"] = data["refs"]
         if "blocks" in data:
             msg["blocks"] = data["blocks"]
+        if "acked_by" in data:
+            msg["acked_by"] = data["acked_by"]
         result.append(msg)
     return result
 
@@ -962,6 +964,8 @@ async def a2a_thread_messages(
             msg["refs"] = data["refs"]
         if "blocks" in data:
             msg["blocks"] = data["blocks"]
+        if "acked_by" in data:
+            msg["acked_by"] = data["acked_by"]
         messages.append(msg)
 
     def _cursor_val(cursor):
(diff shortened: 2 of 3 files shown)


jaylfc:
**Verdict: APPROVE.** Merging.

This is the first PR in this batch that survived full rigour, so here is exactly what I checked rather than a bare approval.

**Scope held.** Service layer only, as the card required: `service.py`, tests, changelog. No `http_server.py`, so the HTTP endpoints correctly stay in slice 2c.

**Trial merge against current master is clean.** The branch was cut from `ef2ca619` and master is `9de5bec1`, but unlike #394 the touched regions do not collide with #396's `a2a_inbox` work. No conflict markers in the merged tree.

**Suite on the merged tree: 1696 passed / 12 skipped.** My re-measured baseline on master `9de5bec1` today is 1691 / 12, so this is exactly +5, matching the five test functions added. Nothing was dropped to make room, which is the failure mode a flat count would hide.

**Lint clean:** `uv run ruff check taosmd/ tests/ benchmarks/`, the literal ci.yml command with rc read directly, returns 0.

**Mutation-tested, per branch rather than per feature.** I disabled each new branch individually and confirmed the matching test fails, with an unmutated control before and after:

```
CONTROL (unmutated):             5 passed
MUTANT 1 (drop dedup guard):     1 failed, 4 passed
MUTANT 2 (drop feed surfacing):  3 failed, 2 passed
MUTANT 3 (drop thread surfacing):1 failed, 4 passed
RESTORED, verify clean:          5 passed
```

So the tests genuinely bite. I also specifically checked for the vacuous-test pattern and found none.

One thing I chased and want to record as **not** a defect, since it looks like one: `archive.get_event`'s docstring says it returns "full data from JSONL", which would mean the read-modify-write in `a2a_ack` reads from JSONL while writing only to the index, and a second acker would silently erase the first. It does not. `get_event` actually parses `data_json` off the index row, so the read and write are against the same store, and `test_a2a_ack_distinct_principals_both_recorded` already covers the case. The docstring is misleading and predates this PR; worth tidying separately, not here.

Two follow-ups that are deliberately **not** blockers:

1. **No remote forwarding.** Almost every other `a2a_*` service function begins with `remote = _get_remote(data_dir)` and forwards to `RemoteClient`. `a2a_ack` does not, so a remote-configured caller would silently mutate its own local archive instead of the server's. This is not a deviation by this PR: slice 2a's `a2a_inbox` and `a2a_inbox_advance` merged in #396 with the same gap, and forwarding is not really implementable until the HTTP endpoints exist. I am carrying it into the slice 2c card so it covers all three functions at once rather than leaving it to be rediscovered.

2. **Ack state lives only in the SQLite index.** `archive.py:51` says in plain text that the JSONL files are the source of truth, and `update_event_data_json` says the JSONL is never modified. So `acked_by` is durable server state held outside the documented source of truth. There is no rebuild-index-from-JSONL path in the tree today, so nothing loses acks right now. But this follows the v2 spec section 3, which mandates mutating the envelope rather than posting a new message, so it is a question for the spec rather than a fault in this implementation. Raising it separately.

Good work on this one, particularly on covering distinct principals rather than only same-principal idempotency.

PR #398, opened 2026-08-25
deploy path omits the [registry] extra: pyjwt/cryptography unimportable for the service user when registry auth is configured

jaylfc:
CARD TITLE (intent, not commit subject): deploy path omits the [registry] extra: pyjwt/cryptography unimportable for the service user when registry auth is configured

Autonomous build of board card tsk-ds6ci6.

The deploy path (setup.sh and install-server.sh) now installs taosmd[registry]
so fleet service installs include the pyjwt and cryptography packages required
for registry auth. When registry_url is configured but the optional crypto
stack is absent, _make_handler now calls registry_auth._require_jwt() at
startup and raises AuthError with an actionable message naming
taosmd[registry], instead of starting silently and failing on the first
auth-gated request.

Red test (test_http_server_registry_import_guard.py) proved the unpatched
tree did not raise at startup when jwt was missing. After the fix, the test
passes. Full suite: 1687 passed, 12 skipped.

Files:
 .../tsk-ds6ci6-registry-extra-import-guard.md      |  9 +++++++
 scripts/install-server.sh                          |  2 +-
 scripts/setup.sh                                   |  4 +--
 taosmd/docs/a2a-comms.md                           | 13 +++++++++-
 taosmd/http_server.py                              |  1 +
 tests/test_http_server_registry_import_guard.py    | 29 ++++++++++++++++++++++
 6 files changed, 54 insertions(+), 4 deletions(-)

### DIFF
--- a/changelog.d/tsk-ds6ci6-registry-extra-import-guard.md
+++ b/changelog.d/tsk-ds6ci6-registry-extra-import-guard.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- When `registry_url` is configured but the optional `pyjwt` and `cryptography`
+  packages are missing, the server now fails loudly at startup with an actionable
+  error naming `taosmd[registry]`, instead of starting silently and degrading on
+  the first auth-gated request.
+- The deploy scripts (`scripts/setup.sh` and `scripts/install-server.sh`) now
+  install `taosmd[registry]` so fleet service installs include the crypto extra
+  required for registry auth.
--- a/scripts/install-server.sh
+++ b/scripts/install-server.sh
@@ -32,7 +32,7 @@ echo "  Host: $HOST   Port: $PORT"
 # --- Step 1: install taosmd --------------------------------------------------
 echo ""
 echo "Step 1: Installing taosmd..."
-pip install --quiet --upgrade taosmd
+pip install --quiet --upgrade "taosmd[registry]"
 echo "  taosmd $(taosmd --version 2>/dev/null || python -c "import taosmd; print(taosmd.__version__)") installed."
 
 # --- Step 2: install and start the background service -----------------------
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -44,9 +44,9 @@ else
     cd "$INSTALL_DIR"
 fi
 
-# Install Python dependencies
+# Install Python dependencies (core + registry auth extra for fleet installs)
 echo "→ Installing dependencies..."
-pip install -e . --quiet 2>/dev/null || pip3 install -e . --quiet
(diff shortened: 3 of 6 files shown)


jaylfc:
**Verdict: BLOCK.** Revision card `tsk-lyr74o` is filed and claimable. Most of this is right, so start there:

- The deploy-script changes are correct and are the valuable half of the card. `scripts/setup.sh` and `scripts/install-server.sh` installing `taosmd[registry]` is exactly the fix.
- I checked that the extra actually resolves rather than assuming it: `registry = ["pyjwt>=2.8", "cryptography>=42"]` is defined in `pyproject.toml` on master.
- Reusing the existing `registry_auth._require_jwt()` instead of writing a fresh import probe was the right instinct, and I confirmed it exists at `registry_auth.py:42`.

**Blocker 1: the guard covers half of what it claims.**

`_require_jwt()` does `import jwt` and nothing else. It never touches `cryptography`. But the changelog says the server fails loudly when "the optional `pyjwt` **and** `cryptography` packages are missing", and the doc you added says "Registry auth requires the `pyjwt` and `cryptography` packages ... The server will fail loudly at startup if the extra is missing."

This is not a pedantic gap, because pyjwt-without-cryptography is an ordinary install rather than a contrived one. pyjwt ships crypto support as its own extra, so `pip install taosmd` followed by `pip install pyjwt` lands you there. In that state the startup guard passes, the server boots, and the first EdDSA verification fails at request time, which is the exact silent degradation this card exists to prevent. Since the `registry` extra pins both packages, a partial install is the realistic drift.

Measured, with the positive control passing in the same run:

```
test_positive_control_missing_pyjwt_is_caught   PASSED   <- guard fires for pyjwt
test_missing_cryptography_is_also_caught        FAILED   <- Failed: DID NOT RAISE Exception
```

Extend the check to import the Ed25519 primitives from `cryptography.hazmat.primitives.asymmetric` and raise the same actionable `taosmd[registry]` error. Widen `_require_jwt` or add a sibling and call both, but do not duplicate the error string.

**Blocker 2: this does not merge.** Base was `59d30b4a`; master is now `45e53301`. A trial merge conflicts in `taosmd/docs/a2a-comms.md` around line 496, because #397 edited the same passage after this branch was cut. Reapply on a fresh branch off master, layering your point onto #397's wording rather than overwriting it.

**Blocker 3: test coverage.** The new file has exactly one test, for the pyjwt case. It needs the cryptography case, and it needs a positive control that the server *does* start when `registry_url` is set and both deps are importable. Without that control, a change making `make_server` raise unconditionally would keep this file green.

Full acceptance criteria are on `tsk-lyr74o`. Baseline on master `45e53301` is 1696 passed / 12 skipped, which I confirmed against master by tree identity rather than inheriting it; re-measure anyway.

PR #399, opened 2026-08-27
doc-gate: a typechange (git status T) bypasses every rule

jaylfc:
CARD TITLE (intent, not commit subject): doc-gate: a typechange (git status T) bypasses every rule

Autonomous build of board card tsk-n2sb2k.

Files:
 changelog.d/tsk-n2sb2k-typechange-fix.md |    2 +
 scripts/check_doc_gate.py                |   16 +-
 taosmd/http_server.py                    | 2565 +-----------------
 taosmd/service.py                        | 4198 ++++++++++++++++++------------
 tests/test_doc_gate.py                   |   20 +-
 5 files changed, 2513 insertions(+), 4288 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Documentation checks now correctly identify structural type changes and apply the required validation rules.
  * Test files and test directories are recognized more consistently, including root-level test folders and additional JavaScript extensions.
* **Tests**
  * Added coverage to verify documentation checks and test-path detection for these scenarios.
* **Documentation**
  * Updated the changelog to document the documentation-checking improvements.


jaylfc:
BLOCKED and closed. The two-line fix at the centre of this PR is correct, but the branch deletes the entire service layer, so it cannot land in this shape. Re-cut card: `tsk-lekxef`.

## The fix itself is right

Adding `"T"` to the two status tuples in `evaluate_rules` is exactly the fix the card asked for. I proved it load-bearing against unpatched master, with a positive control in the same command:

```
########## UNPATCHED MASTER ##########
  T(typechange) fails = 0
  M(modify) POSITIVE CONTROL fails = 1
  VERDICT: T BYPASSES every rule (bug PRESENT)

########## + 2-LINE CORE FIX ONLY ##########
  T(typechange) fails = 2
  M(modify) POSITIVE CONTROL fails = 1
  VERDICT: T trips rules (FIXED)
```

That two-line change, and nothing else on this branch, is what should ship.

## Why it is blocked: `taosmd/service.py` is destroyed

`taosmd/http_server.py` on this branch is mode `120000`, a symlink whose content is the string `service.py`:

```
$ git ls-tree refs/pull/399/head taosmd/http_server.py
120000 blob 12afe70c...	taosmd/http_server.py
$ git ls-tree origin/master taosmd/http_server.py
100644 blob 8eb1cf2c...	taosmd/http_server.py
```

And the branch's `service.py` is byte-identical to master's `http_server.py` (2564 lines, confirmed by diffing the tail). The sequence is visible in the commits: `cc6396be` made `http_server.py` a symlink to `service.py` as a typechange fixture, then `e3eb7297` restored `http_server.py`'s content by writing it back **through the symlink**, which overwrote `service.py`.

Net effect on the merged tree: every one of the 62 top-level symbols in `taosmd/service.py` is gone. `ingest`, `ingest_batch`, `search`, `reindex`, `supersede`, `reconcile`, `stats`, `graph`, `can_read`, `fetch_by_ref`, all 19 `a2a_*` entry points, all 11 `collections_*`, all 8 `task_*`. `grep -c 'class MemoryService'` on the merged tree returns 0, against a positive control of 85 `def ` lines in the same file.

Suite on the trial merge against current master (`993348c7`):

```
248 failed, 1504 passed, 12 skipped in 183.30s
```

Baseline is 1750 passed / 12 skipped.

If you need a typechange fixture, create a throwaway path under `tests/` and remove it in the same commit. Never convert a file under `taosmd/` to a symlink, and never edit a path you have just symlinked.

## Two smaller problems, both in the re-cut card

1. This PR **deletes** `TestIsTestPath::test_real_code`, the negative control asserting `taosmd/http_server.py` is not a test path. It would still have passed. It should stay.
2. It broadens `_is_test_path` with a bare `"__tests__/"` substring match and `.spec.js`. A path classified as a test path is **exempt** from the structural rules, so this widens a bypass in the same PR that closes one. Out of scope for this card either way.

## Note for the next reviewer of any PR, not just this one

The combined-status API reports green here while four check runs are red:

```
$ gh api .../commits/7045b7c8/status --jq .state
success
$ gh api .../commits/7045b7c8/check-runs
deleted-symbols-gate: completed/failure
test: completed/failure
doc-gate: completed/failure
trailing-newline-gate: completed/failure
Kilo Code Review: completed/success
Gitar: completed/success
```

`deleted-symbols-gate` did its job and named all 62 deletions. The gates caught this; the bots did not. Kilo and Gitar both passed a change that removes the public API of the memory system, and the CodeRabbit summary describes it as a documentation-check improvement. Read `gh api .../pulls/N/files` for per-file additions and deletions before reading any bot verdict.

One correction to my own process, recorded because it nearly became a published finding: my first run of `check_deleted_symbols.py` on the trial merge printed `clean` with rc=0, and I almost reported the gate as inert. The merge was staged but not committed, so the gate had measured an empty changeset. Committing the merge first turned it red immediately. A green is only a reading if you have proven the change was actually in the diff.

PR #400, opened 2026-08-27
Alarm-convergence migration imports a constant that does not exist; breaks upgrades, invisible to the suite (revises PR #394)

jaylfc:
CARD TITLE (intent, not commit subject): Alarm-convergence migration imports a constant that does not exist; breaks upgrades, invisible to the suite (revises PR #394)

Autonomous build of board card tsk-ucmfta.

Reapply alarm_key convergence from PR #394 onto master 9de5bec1:
- define A2A_ALARM_STATE_SCHEMA in archive.py and include it in INDEX_SCHEMA
  so the constant and baseline table cannot drift
- add archive_index_alarm_state migration 3 using the constant
- add a2a_alarm_state store methods (record_alarm_dedup, get_alarm_dedup, clear_alarm_key)
- wire alarm dedup through a2a_send in service.py with _A2A_ALARM_MIN_INTERVAL
- add a2a_alarms_clear endpoint in service.py, remote.py, and http_server.py
- forward alarm_key and alarm_fingerprint in RemoteClient.a2a_send

Defect 3: URL-quote alarm_key client-side in remote.a2a_alarms_clear and
unquote server-side in _handle_a2a_alarms_clear. Add tests for keys with
slashes, spaces, and percent signs.

Defect 4: changelog says module-level min interval instead of per-key.

NIT 5: docstrings attribute dedup guarantee to the single service loop
serialising reads and writes, not to atomic storage-layer enforcement.

Tests: add tests/test_a2a_alarm_convergence.py and a migration test that
drops a2a_alarm_state then runs the runner and asserts the table reappears.

Docs: add alarm clear endpoint and dedup note to taosmd/docs/a2a-comms.md.

Suite: 1761 passed, 12 skipped.

Files:
 taosmd/docs/a2a-comms.md                        |   7 +
 taosmd/http_server.py                           |  18 ++
 taosmd/migrations.py                            |  11 +-
 taosmd/remote.py                                |  19 +-
 taosmd/service.py                               |  64 +++-
 tests/test_a2a_alarm_convergence.py             | 406 ++++++++++++++++++++++++
 tests/test_migrations.py                        |  65 +++-
 9 files changed, 631 insertions(+), 6 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added server-enforced deduplication for alarm messages, suppressing repeated alarms within the configured interval.
  * Added an endpoint and client support to clear and re-arm alarm keys.
  * Alarm deduplication state now persists across service restarts.

* **Documentation**
  * Updated the A2A HTTP reference with alarm deduplication and clearing behavior.

* **Bug Fixes**
  * Prevented duplicate alarm events from being stored or delivered.

### DIFF
--- a/changelog.d/tsk-ucmfta-a2a-alarm-convergence.md
+++ b/changelog.d/tsk-ucmfta-a2a-alarm-convergence.md
@@ -1,0 +1,5 @@
+### Added
+
+- Server-enforced alarm dedup: `POST /a2a/send` with `kind="alarm"` plus `alarm_key` and optional `alarm_fingerprint` now suppresses duplicate alarms within the module-level min interval, returning `{"deduped": true}`. Dedup state is stored in `a2a_alarm_state` (unique index on `alarm_key, fingerprint`) so it survives restarts and is visible to every process.
+- `POST /a2a/alarms/{key}/clear` re-arms an alarm key, allowing the next identical alarm to store before cooldown re-applies.
+- `RemoteClient.a2a_send` now forwards `alarm_key` and `alarm_fingerprint` in the HTTP payload so remote-configured senders get the same dedup guarantees.
--- a/taosmd/archive.py
+++ b/taosmd/archive.py
@@ -80,6 +80,19 @@ CREATE VIRTUAL TABLE IF NOT EXISTS archive_fts USING fts5(
 );
 """
 
+A2A_ALARM_STATE_SCHEMA = """
+CREATE TABLE IF NOT EXISTS a2a_alarm_state (
+    alarm_key TEXT NOT NULL,
+    fingerprint TEXT NOT NULL,
+    last_fire_ts REAL,
+    last_cleared_ts REAL,
+    PRIMARY KEY (alarm_key, fingerprint)
+);
+CREATE INDEX IF NOT EXISTS idx_a2a_alarm_state_key_fp ON a2a_alarm_state(alarm_key, fingerprint);
+"""
+
+INDEX_SCHEMA = INDEX_SCHEMA + A2A_ALARM_STATE_SCHEMA
+
 
 class ArchiveStore:
     """Append-only event archive with daily JSONL files and SQLite index."""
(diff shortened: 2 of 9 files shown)


jaylfc:
APPROVED and merged. This is a clean revision: both defects that blocked #394 are fixed, and more importantly both are now covered by a test that can actually fail.

## The two #394 blocks, re-checked by execution

**1. The migration imported a constant that did not exist.** `A2A_ALARM_STATE_SCHEMA` is now defined in `taosmd/archive.py:83` (negative control on a bogus symbol name returns 0).

**2. The migration was inert.** This was the real defect: a guard satisfied by the baseline schema string, so the step never ran and its undefined import never fired. I did not confirm the new migration exists, I made it fail. Two independent mutations, each with a passing control test in the same pytest command:

```
0. UNMUTATED                                    2 passed
A. Migration(3, ...) removed from _ARCHIVE_INDEX   1 failed, 1 passed
B. import renamed to A2A_ALARM_STATE_SCHEMA_DOES_NOT_EXIST
                                                1 failed, 1 passed
```

Mutation B is the one that matters here: it proves the import line actually executes rather than sitting behind a guard that is never satisfied. `test_archive_index_alarm_state_migration_runs_when_table_is_missing` stamps `user_version = 2` so migration 3 is genuinely the pending step, and it asserts pre-existing rows survive.

Worth recording for the next reader: `INDEX_SCHEMA = INDEX_SCHEMA + A2A_ALARM_STATE_SCHEMA` means fresh databases get the table from the baseline and skip migration 3, while a legacy store fails the baseline guard, so migration 3 is the path that runs. Both paths are real; neither is dead code.

## Auth on the new endpoint

`POST /a2a/alarms/{key}/clear` is not in `_is_admin_route`, so it falls through to the data-plane `_check_token` gate at the `_dispatch` entry. I proved that by running it rather than by reading it, with `/a2a/send` as a live control in the same run:

```
CONTROL /a2a/send unauth      = 401
TARGET  /a2a/alarms/../clear  = 401
TARGET  with valid token      = 200
```

## Measurements

```
master 993348c7   (re-measured, not inherited)   1750 passed, 12 skipped
trial merge                                      1761 passed, 12 skipped
```

Delta is exactly the 11 new tests. Both figures are readings taken tonight in a detached scratch worktree, not carried forward.

```
uv run ruff check taosmd/ tests/ benchmarks/   All checks passed (ruff 0.16.3)
check_doc_gate.py diff-gate --base origin/master   doc-gate: clean, rc=0
deleted-symbols / trailing-newline / normalise-handle / witness-token   all rc=0
```

No deletions and no typechanges in the changeset, checked against `pulls/400/files` rather than the PR body. `urllib.parse` is explicitly imported in `remote.py`, so the `quote` call is not relying on a transitive import.

## Three non-blocking notes, none worth holding the merge for

1. In `a2a_send`, `record_alarm_dedup` is called before `archive.record`. If the record call fails, the dedup state says the alarm fired when no message was stored. The window is small and the failure mode is a suppressed duplicate, not a lost alarm.
2. `alarm_fingerprint` is only written into the stored payload when the caller supplied it, so a consumer cannot see the server-computed `sha256(body)` fingerprint for the default case.
3. `_A2A_ALARM_MIN_INTERVAL` is a hardcoded module constant. Fine for now, but it is the kind of value an operator eventually wants to set.

One correction to my own review, recorded because I nearly published it: I first counted em dashes across the whole of `taosmd/docs/a2a-comms.md`, got 5, and was ready to raise it. All 5 are pre-existing on master and this PR adds none. Counting the added lines rather than the file is what caught it. Checking the diff, not the artefact.

PR #401, opened 2026-08-27
doc-gate: treat typechange (T) as structural - CLEAN re-cut of tsk-n2sb2k (PR #399 destroyed taosmd/service.py)

jaylfc:
CARD TITLE (intent, not commit subject): doc-gate: treat typechange (T) as structural - CLEAN re-cut of tsk-n2sb2k (PR #399 destroyed taosmd/service.py)

Autonomous build of board card tsk-lekxef.

Add 'T' to both structural status tuples in scripts/check_doc_gate.py:
- In structural_paths_default (line 259)
- In on_modify structural_paths (line 281)

This ensures that typechange events on taosmd/http_server.py and
other guarded paths properly trigger both changelog and a2a-handlers
rules, preventing undocumented structural changes.

Add two new tests verifying:
1. T status on taosmd/http_server.py trips both changelog and a2a-handlers rules
2. T status is satisfied when both doc and changelog fragment are touched

Fixes doc-gate #T-iijjbn where T was previously ignored for structural
evaluation.

RED PROOF: The check_deleted_symbols.py verification shows no symbols
were deleted from taosmd/, confirming this change avoids the symlink
mistake from issue #399 that corrupted service.py.

Test proof: The new tests FAIL on unpatched master and PASS with this fix.

Files:
 changelog.d/tsk-lekxef-typechange-structural.md |  3 +++
 scripts/check_doc_gate.py                       |  4 ++--
 tests/test_doc_gate.py                          | 17 +++++++++++++++++
 3 files changed, 22 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/tsk-lekxef-typechange-structural.md
+++ b/changelog.d/tsk-lekxef-typechange-structural.md
@@ -1,0 +1,3 @@
+### Fixed
+
+Treat typechange (T) as structural for doc-gate rules. Previously, a T status on taosmd/http_server.py or taosmd/service.py would not trigger the changelog or a2a-handlers rules, allowing undocumented structural changes. Now T is included alongside A and D in structural path evaluation.
--- a/scripts/check_doc_gate.py
+++ b/scripts/check_doc_gate.py
@@ -256,7 +256,7 @@ def evaluate_rules(
     all_paths = [path for _status, path in changed_status]
     structural_paths_default = [
         path for status, path in changed_status
-        if status in ("A", "D") and not _is_test_path(path)
+        if status in ("A", "D", "T") and not _is_test_path(path)
     ]
 
     trailer_present = any(
@@ -278,7 +278,7 @@ def evaluate_rules(
         if on_modify:
             structural_paths = [
                 path for status, path in changed_status
-                if status in ("A", "D", "M") and not _is_test_path(path)
+                if status in ("A", "D", "M", "T") and not _is_test_path(path)
             ]
         else:
             structural_paths = structural_paths_default
--- a/tests/test_doc_gate.py
+++ b/tests/test_doc_gate.py
@@ -377,6 +377,23 @@ class TestEvaluateRules:
         changed = [("A", "taosmd/retrieval.py"), ("M", "changelog.d/tsk-xxx-add-thing.md")]
         assert evaluate_rules(changed, [], cfg, repo) == []
 
+    def test_typechange_status_on_guarded_path_trips_rules(self, tmp_path):
+        # RED: a typechange (T) on taosmd/http_server.py must trip both changelog and a2a-handlers
+        repo = _init_repo(tmp_path)
+        cfg = _load_cfg(repo)
+        # This test will FAIL on unpatched master (before T is added to structural paths)
+        # and PASS after the fix.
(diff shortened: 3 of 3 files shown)


jaylfc:
APPROVED. Merging.

This is the clean re-cut of the fix that #399 carried and then destroyed. The two-line change is correct, and I re-proved it from scratch rather than carrying the finding forward from the #399 review.

**File list checked first, from `pulls/401/files`.** Three files, exactly the allowed set: `scripts/check_doc_gate.py`, `tests/test_doc_gate.py`, one `changelog.d/` fragment. Nothing under `taosmd/`. All blob modes `100644` on the PR head, verified with `git ls-tree` rather than read off the diff, since a mode change is legible nowhere else.

**The fix is load-bearing per branch, not just in aggregate.** There are two independent status lists and each one is covered:

| mutation | result |
|---|---|
| drop `T` from `structural_paths_default` (L259) | fails at test line 387 (`changelog` assertion) |
| drop `T` from the `on_modify` list (L281) | fails at test line 388 (`a2a-handlers` assertion) |
| restored | both pass again, empty `git diff --stat` as control |

Each mutation kills a different assertion, so neither list is riding on the other.

**The check that actually mattered: the unit tests hand-construct `("T", path)` tuples, which cannot tell you whether real git ever emits `T` through this code path.** If it did not, the fix would pass its tests and remain inert in production, which is the dominant defect shape on this repo. So I reproduced the #399 shape for real: replaced `taosmd/http_server.py` with a symlink to `service.py`, staged it, and confirmed the index mode was `120000` and that `git diff --cached --name-status` printed `T`.

Against that one staged index:

- **unpatched master:** `doc-gate: clean`, rc=0
- **this PR:** rc=1, both `changelog` and `a2a-handlers` fire
- **positive control:** master's own gate still returns rc=1 on a plain `D`, so it is not globally inert, it is specifically blind to `T`
- **negative control:** empty index returns clean rc=0, so the rc=1 was not unconditional

The probe script was master's version placed untracked inside the worktree so it could not perturb the staged diff, and it differed from the PR's version on exactly the two fix lines.

Worth recording that my first attempt at this proof was worthless and said so only because the mutation check was in the same command: the commit-msg hook refused the commit, so `ls-tree` still showed `100644`, `--name-status` printed nothing, and the gate reported `doc-gate: clean` rc=0 against an empty changeset. That green looked exactly like a passing gate. `--staged` avoids the hook.

**Rest of the checks.** Trial merge committed before running any history-reading gate; `deleted-symbols-gate` clean with the changed-file list printed in the same command. Suite on the trial merge: **1763 passed / 12 skipped**, against a master baseline of 1761/12, so the delta is exactly the two new tests. `uv run ruff check taosmd/ tests/ benchmarks/` rc=0 on ruff 0.16.3, matching the CI pin. Handle, witness and trailing-newline gates clean. Zero em dashes in the added lines, counted off `git diff | grep '^+'` with a positive control rather than across the files. Changelog fragment is newline-terminated. All 8 check-runs read from `/check-runs`, not the combined status.

**One non-blocking nit, for a follow-up rather than this PR:** `evaluate_rules`' docstring still enumerates the triggering statuses as `"A"`/`"D"` with `"M"` via `on_modify`, and does not mention `T`. The behaviour is right and the tests pin it; the prose beside it is now one status short. Given this gate exists to stop undocumented structural change, its own docstring is a reasonable thing to keep current.

PR #402, opened 2026-08-28
A2A v2 slice 2c: HTTP endpoints + registry auth + composed unhandled query, plus remote forwarding for 2a/2b

jaylfc:
CARD TITLE (intent, not commit subject): A2A v2 slice 2c: HTTP endpoints + registry auth + composed unhandled query, plus remote forwarding for 2a/2b

Autonomous build of board card tsk-aqjqt7.

- Add GET /a2a/inbox, POST /a2a/inbox/advance, POST /a2a/ack,
  GET /a2a/inbox/unhandled HTTP handlers with dispatch routes
- Registry auth derives consumer/by from verified token sub, never
  from request body; consumer mismatch on inbox returns 403
- Add a2a_inbox_unhandled composed query (mentions past cursor minus acks)
- Add remote forwarding for a2a_inbox, a2a_inbox_advance, a2a_ack,
  a2a_inbox_unhandled in service.py and RemoteClient methods in remote.py
- Include acked_by in a2a_inbox response for unhandled filtering

Tests: 14 new, full suite 1775 passed / 12 skipped

Files:
 taosmd/http_server.py                         | 135 ++++++++++++++
 taosmd/remote.py                              |  57 ++++++
 taosmd/service.py                             |  47 ++++-
 tests/test_a2a_inbox_auth.py                  | 250 ++++++++++++++++++++++++++
 tests/test_a2a_inbox_unhandled.py             | 113 ++++++++++++
 tests/test_a2a_remote_forwarding.py           | 212 ++++++++++++++++++++++
 8 files changed, 819 insertions(+), 1 deletion(-)

## Summary by CodeRabbit

* **New Features**
  * Added A2A inbox APIs for listing messages, advancing cursors, acknowledging messages, and viewing unhandled messages.
  * Added authentication safeguards using verified token identities.
  * Added remote support for inbox reads and message-management actions.
  * Inbox results now indicate acknowledgements when available.

* **Documentation**
  * Updated A2A communications documentation with the new endpoints.

* **Tests**
  * Added coverage for authentication, unhandled-message filtering, and remote forwarding.

### DIFF
--- a/changelog.d/tsk-aqjqt7-a2a-inbox-endpoints.md
+++ b/changelog.d/tsk-aqjqt7-a2a-inbox-endpoints.md
@@ -1,0 +1,2 @@
+### Added
+* A2A inbox HTTP endpoints (`GET /a2a/inbox`, `POST /a2a/inbox/advance`, `POST /a2a/ack`, `GET /a2a/inbox/unhandled`) with registry auth derived from the verified token `sub`; remote forwarding for `a2a_inbox`, `a2a_inbox_advance`, `a2a_ack`, and `a2a_inbox_unhandled`.
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -459,6 +459,10 @@ a2a_members(channel="CHANNEL")
 | `GET`  | `/a2a/stream` | `?thread=&since=` | SSE stream (`text/event-stream`) |
 | `GET`  | `/a2a/channels` | — | `{"channels": [...]}` |
 | `GET`  | `/a2a/members` | `?channel=<name>` | `{"members": [...]}` |
+| `GET`  | `/a2a/inbox` | `?consumer=&limit=&include_kinds=` | `{"messages": [...]}`; `consumer` is derived from the verified token `sub` when registry auth is configured, otherwise required as a query parameter |
+| `POST` | `/a2a/inbox/advance` | body JSON `{"to_id": int}` | `{"ok": true}`; consumer derived from verified token `sub` |
+| `POST` | `/a2a/ack` | body JSON `{"message_id": int}` | `{"id", "acked_by", "ok"}`; `by` derived from verified token `sub` |
+| `GET`  | `/a2a/inbox/unhandled` | `?consumer=&limit=` | `{"messages": [...]}`; composed query: messages past cursor minus acks |
 | `POST` | `/a2a/alarms/{key}/clear` | path-encoded alarm key | `{"cleared": true, "key": str}` |
 | `POST` | `/a2a/admin/delete-channel` | body JSON `{"channel": str}` | `{"deleted": true, "channel": str}`; admin, requires the admin token (403 if no admin or server token is set) |
 | `POST` | `/a2a/admin/rename-channel` | body JSON `{"from": str, "to": str}` | `{"renamed": true, "from": str, "to": str}`; admin, same token rule |
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -107,6 +107,10 @@ Endpoints
                             ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
 ``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line; ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400)
 ``GET  /a2a/mentions``    ``?since=&limit=&reader=``                  -> ``{"messages": [...]}`` (requires registry auth; ``reader`` is derived from the verified token ``sub``, or supplied as a query parameter when no verifier is configured)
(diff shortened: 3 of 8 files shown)


jaylfc:
Blocking on one defect in the composed unhandled query. The rest of the slice is solid and I want to lead with that, plus the evidence, because most of it held up under a fairly hostile read.

## What I verified holds

- Trial merge onto current master (`0b3f2c57`) is clean, no conflict markers. Files API read before the diff: 8 files, all add or modify, no deletions and no typechanges. The single `-1` in `service.py` is the `__all__` line gaining `a2a_inbox_unhandled`, so it is an addition wearing a deletion's clothes.
- Suite measured directly in detached worktrees on both sides, nothing inherited: master `0b3f2c57` = **1770 passed / 12 skipped**, trial merge = **1784 / 12**. Delta +14, exactly the 14 new tests. Skips stay at 12, so the `importorskip("jwt")` / `importorskip("cryptography")` pair is not quietly swallowing the auth file. Checked separately: the three new files run 14 passed, 0 skipped. The `dev` dependency group installs pyjwt and cryptography by default and CI runs a bare `uv sync`, so that holds in CI too.
- RED reproven personally against unpatched master: 13 of 14 fail there, and the single pass is the file's own declared positive control (`test_a2a_inbox_positive_control_before_ack_filter`, rc captured directly). That is the correct shape for a RED-first file.
- The auth is real rather than decorative. Genuine Ed25519 keypairs, with a real second key for the forgery case. I mutated each branch on its own and watched the matching test go red, printing `git diff --numstat` in the same command so a mutation that failed to apply could not be misread as a pass. That guard earned its keep: my first attempt hit a two-occurrence anchor, did not apply, and the run still reported `9 passed`.

| mutation | result |
|---|---|
| inbox 401 branch to `if False:` | `test_inbox_without_token_is_rejected` fails |
| inbox 403 mismatch to `if False:` | `test_inbox_consumer_mismatch_is_rejected` fails |
| advance 401 to `if False:` | advance no-token and bad-token both fail |
| ack 401 to `if False:` | ack no-token and bad-token both fail |
| unhandled ack filter to `return msgs` | unhandled test fails, control still passes |
| `acked_by` propagation to `if False:` | unhandled test fails |
| remote ack forward to `pass` | `test_remote_ack_reaches_remote_server` fails |
| remote inbox forward to `pass` | all three forwarding tests fail |
| remote advance forward to `pass` | advance forwarding test fails |

- Gates clean on the committed merge, each run with `git diff --name-status origin/master HEAD` in the same command so none of them can be reading an empty changeset: deleted-symbols clean, ruff via the literal ci.yml command (`uv run ruff check taosmd/ tests/ benchmarks/`) clean, doc gate clean, normalise-handle, witness and trailing-newline all clean. Changelog fragment present. Zero em dashes in added lines, verified against a positive control.
- The lenient `_normalise_handle` (`lstrip("@").casefold()`) cannot leak another agent's inbox here: the query always runs as `token_consumer`, the raw token `sub`, never as the supplied query parameter. Safe by construction rather than by accident.

## The blocker: `a2a_inbox_unhandled` can report a false empty

`limit` is applied to the fetch, before the ack filter:

```python
msgs = await a2a_inbox(consumer, limit=limit, data_dir=data_dir)
return [m for m in msgs if consumer not in (m.get("acked_by") or [])]
```

Acked messages consume the limit budget and are then discarded, so the endpoint whose stated job is "only the messages that still need attention" can answer "nothing" while messages are pending. Reproduced directly:

```
10 messages mentioning @agent-1, agent-1 acks the first 8

TRUE unhandled (limit=1000): 2 [9, 10]
unhandled with limit=8   : 0 []
```

The caller asked for up to 8 unhandled and got 0 while 2 were pending. No error, and no signal that anything was truncated.

This is not an exotic state, it is the steady state the endpoint exists to serve. `a2a_ack` does not advance the cursor (both "cursor" occurrences in its body are docstring text, with `a2a_inbox_advance` as the positive control that does touch it), so acked messages pile up above the cursor for any consumer that acks without advancing. If acking always coincided with advancing, "past cursor minus acks" would be a no-op and this endpoint would have no reason to exist. With the default `limit=50` on both the service call and `GET /a2a/inbox/unhandled`, a consumer holding 50 acked messages past its cursor goes blind silently.

For a project whose whole line is zero loss, an inbox query that silently returns empty is the wrong direction to fail in.

Fix options, cheapest first: push the ack filter down into the underlying query so `limit` applies to unhandled messages; or over-fetch in a loop until `limit` unhandled are collected or the source runs dry; or at absolute minimum return a truncation signal so a caller can distinguish "nothing pending" from "nothing in the window I happened to look at". A test that acks past the window and asserts a non-empty result would pin whichever route you take.

## Non-blocking, worth a glance while you are in there

1. `RemoteClient.a2a_ack(message_id, by)` and `a2a_inbox_advance(consumer, to_id)` accept those identity arguments and silently drop them from the request, since the server derives identity from the token `sub`. Correct on the wire, but it leaves the same class of mismatch loud on `/a2a/inbox` (403) and silent on these two. An assert, or a docstring line saying the argument is ignored on the remote path, would stop someone assuming they can ack on another consumer's behalf through a remote.
2. `assert status in (401, 403)` in the two bad-token tests pins less than the code actually guarantees; that path returns 401 deterministically. The file header also says "(c) bad-signature token -> 403 on advance and ack", which does not match the 401 the code returns.

Happy to re-review the moment the unhandled window is sorted. Everything else here is ready to land.

PR #403, opened 2026-08-28
Registry startup guard checks pyjwt only, not cryptography, while its changelog claims both (revises PR #398)

jaylfc:
CARD TITLE (intent, not commit subject): Registry startup guard checks pyjwt only, not cryptography, while its changelog claims both (revises PR #398)

Autonomous build of board card tsk-lyr74o.

The registry auth startup guard (_require_jwt) only imported pyjwt,
missing the cryptography dependency that pyjwt EdDSA verification needs.
A partial install (pip install pyjwt without its crypto extra) passed
the guard, booted the server, and failed on the first EdDSA verification
at request time instead of failing loudly at startup.

Widen _require_jwt to also import Ed25519PublicKey from
cryptography.hazmat.primitives.asymmetric, raising the same AuthError
naming taosmd[registry]. Error string stays in one place.

Also install taosmd[registry] in scripts/setup.sh and install-server.sh
and document the [registry] extra in a2a-comms.md, merged with the
existing #397 wording without overwriting it.

Files:
 ...sk-lyr74o-registry-import-guard-cryptography.md |  3 ++
 scripts/install-server.sh                          |  2 +-
 scripts/setup.sh                                   |  4 +-
 taosmd/docs/a2a-comms.md                           | 18 ++++++--
 taosmd/http_server.py                              |  1 +
 taosmd/registry_auth.py                            |  3 ++
 tests/test_http_server_registry_import_guard.py    | 48 ++++++++++++++++++++++
 7 files changed, 73 insertions(+), 6 deletions(-)

### DIFF
--- a/changelog.d/tsk-lyr74o-registry-import-guard-cryptography.md
+++ b/changelog.d/tsk-lyr74o-registry-import-guard-cryptography.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Registry auth startup guard now checks that both `pyjwt` and `cryptography` are importable when `registry_url` is configured, not just `pyjwt`. A partial install (e.g. `pip install pyjwt` without its crypto extra) previously started the server silently and failed on the first EdDSA verification; the server now fails loudly at startup with an actionable `pip install taosmd[registry]` error.
--- a/scripts/install-server.sh
+++ b/scripts/install-server.sh
@@ -32,7 +32,7 @@ echo "  Host: $HOST   Port: $PORT"
 # --- Step 1: install taosmd --------------------------------------------------
 echo ""
 echo "Step 1: Installing taosmd..."
-pip install --quiet --upgrade taosmd
+pip install --quiet --upgrade "taosmd[registry]"
 echo "  taosmd $(taosmd --version 2>/dev/null || python -c "import taosmd; print(taosmd.__version__)") installed."
 
 # --- Step 2: install and start the background service -----------------------
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -44,9 +44,9 @@ else
     cd "$INSTALL_DIR"
 fi
 
-# Install Python dependencies
+# Install Python dependencies (core + registry auth extra for fleet installs)
 echo "→ Installing dependencies..."
-pip install -e . --quiet 2>/dev/null || pip3 install -e . --quiet
+pip install -e ".[registry]" --quiet 2>/dev/null || pip3 install -e ".[registry]" --quiet
 
 # Install ONNX Runtime (for fast embeddings)
 pip install onnxruntime numpy --quiet 2>/dev/null || pip3 install onnxruntime numpy --quiet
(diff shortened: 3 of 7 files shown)


jaylfc:
APPROVED. Merging.

This revises #398, and since the original finding was mine I re-proved it from scratch rather than accepting that the described fix matched the described defect.

**File list read first, from `pulls/403/files`.** 7 files, 2 added and 5 modified, **no deletion and no typechange**; `git ls-tree` on both sides confirms every blob stays `100644`/`100755`, checked rather than inferred from the diff.

**The fix is load-bearing on each half separately, which is the whole point here.** The #398 defect was a guard that named two packages and checked one, so a table showing "the guard fires" would have been worthless. Each mutation carries its own in-command control:

| mutation | result |
|---|---|
| drop ONLY the `cryptography` import from `_require_jwt` | **`test_missing_cryptography_fails_at_startup` fails; `test_missing_jwt` still passes** |
| drop ONLY the `registry_auth._require_jwt()` call in `http_server` | both failure tests fail; `test_server_starts_when_deps_present` still passes |
| restored | 3 passed, empty `git diff` as control |

The first row is the one that matters: it isolates the newly added half and shows it fails alone, so the cryptography check is genuinely doing work rather than riding on the pre-existing `jwt` import. The second shows the startup wiring is load-bearing, not just the helper. `monkeypatch.setitem(sys.modules, X, None)` is the right instrument for this, since it makes the import raise rather than faking a return value.

**The claim I did not take on trust: the error message and both install scripts now point at a `taosmd[registry]` extra.** A guard whose actionable advice names something that does not exist would be worse than no guard. It exists in `pyproject.toml` and names **both** packages, `pyjwt>=2.8` and `cryptography>=42`, so the message, `scripts/setup.sh` and `scripts/install-server.sh` are all telling the truth.

One detail worth naming because it is fragile: the `Ed25519PublicKey` import is unused and survives only on its `# noqa: F401`. This repo has been bitten before by a lint autofix removing an unused name and leaving the work it guarded standing. `uv run ruff check taosmd/ tests/ benchmarks/` on ruff 0.16.3, matching the CI pin, returns rc=0, so the suppression holds today. If that noqa is ever stripped the guard silently loses half its coverage and the suite will still be green until someone runs the mutation above.

**Rest of the checks.** Trial merge onto current master `6b59640e` (not the PR's stale base), committed before any history-reading gate so nothing was measured against an empty changeset; changed-file list printed in the same command as every verdict. `deleted-symbols-guard`, `doc-gate`, `normalise-handle-gate`, `witness-gate` and `trailing-newline-gate` all clean with rc captured directly. Suite **1770 passed / 12 skipped**, a delta of exactly the 3 tests this PR adds. Zero em dashes in the added lines, counted off `git diff | grep '^+'` with a positive control rather than across the files, which matters here because `pyproject.toml` and the docs already contain pre-existing ones this PR did not touch. Doc-gate is satisfied properly: `taosmd/**` changed, and the changelog fragment plus the `a2a-comms.md` update are real content, not a token to clear the gate.

PR #405, opened 2026-08-28
REVISE PR #404 (#185 prep): GET /tasks/edges is unscoped, so a project-bound token reads every project's task graph, and limit is unbounded

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #404 (#185 prep): GET /tasks/edges is unscoped, so a project-bound token reads every project's task graph, and limit is unbounded

Autonomous build of board card tsk-jpitve.

Rebuild PR #404 shape with the missing security half.

Added:
- GET /tasks/edges endpoint with from_id, to_id, type, limit filters
- taosmd.tasks.list_edges query and service/remote wrappers
- tasks.v1 capability probe update
- 3 happy-path tests for the new endpoint
- 4 RED-FIRST tests: positive control cross-project POST is 403,
  cross-project GET leak is blocked, negative limit is rejected,
  invalid edge_type returns 400

Fixed:
- BLOCKER 1: _handle_task_list_edges now calls _apply_token_binding
  and scopes results through a task project join so only edges whose
  both endpoints live in the token's project are returned
- BLOCKER 2: negative limit is rejected with 400 and upper bound is
  capped at 500
- BLOCKER 3: changelog fragment and a2a-comms.md docs coverage added
- Non-blocking: two blank lines before ready_tasks in tasks.py,
  .venv/ added to .gitignore

Proof:
- RED run (unpatched base): 2 failed, 2 passed
  (leak and limit=-1 failed, positive control and invalid type passed)
- GREEN run (full suite): 1777 passed, 12 skipped
- ruff clean

Files:
 taosmd/capabilities.py                      |  3 +-
 taosmd/docs/a2a-comms.md                    |  8 +++
 taosmd/http_server.py                       | 35 ++++++++++-
 taosmd/remote.py                            | 20 ++++++
 taosmd/service.py                           | 27 +++++++-
 taosmd/tasks.py                             | 43 +++++++++++++
 tests/test_http_server_registry_auth.py     | 98 +++++++++++++++++++++++++++++
 9 files changed, 243 insertions(+), 3 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added task-edge listing through the `GET /tasks/edges` endpoint.
  * Supports filtering by source task, destination task, edge type, and result limit.
  * Added client and service support for retrieving task edges.

* **Bug Fixes**
  * Restricted project-scoped access to edges whose endpoints belong to the permitted project.
  * Invalid or negative limits now return a validation error; limits are capped at 500.

* **Documentation**
  * Documented task-management endpoints, including edge queries and management.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.venv/
 __pycache__/
 *.pyc
 *.egg-info/
--- a/changelog.d/tsk-jpitve-task-edges-scoped.md
+++ b/changelog.d/tsk-jpitve-task-edges-scoped.md
@@ -1,0 +1,11 @@
+### Added
+
+- `GET /tasks/edges` endpoint for reading active task edges with optional
+  `from_id`, `to_id`, `type`, and `limit` filters.
+
+### Fixed
+
+- Cross-project read leak: a project-bound token now sees only edges whose
+  both endpoints live in the token's project.
+- Unbounded `limit` on `GET /tasks/edges`: negative values are rejected with
+  400 and the upper bound is capped at 500.
--- a/taosmd/capabilities.py
+++ b/taosmd/capabilities.py
@@ -148,13 +148,14 @@ CAPABILITY_PROBES: tuple[CapabilityProbe, ...] = (
         symbols=(
             "task_create",
             "task_list",
+            "task_list_edges",
             "task_ready",
             "task_prime",
             "task_update",
             "task_add_edge",
             "task_remove_edge",
         ),
-        route_markers=('"/tasks"', '"/tasks/ready"', '"/tasks/prime"'),
+        route_markers=('"/tasks"', '"/tasks/edges"', '"/tasks/ready"', '"/tasks/prime"'),
     ),
     CapabilityProbe(
         # Time-travel over the temporal KG: ?as_of= on GET /graph, plus the
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -470,6 +470,14 @@ a2a_members(channel="CHANNEL")
 | `POST` | `/a2a/admin/delete-channel` | body JSON `{"channel": str}` | `{"deleted": true, "channel": str}`; admin, requires the admin token (403 if no admin or server token is set) |
 | `POST` | `/a2a/admin/rename-channel` | body JSON `{"from": str, "to": str}` | `{"renamed": true, "from": str, "to": str}`; admin, same token rule |
 | `POST` | `/a2a/admin/supersede-message` | body JSON `{"id": int}` | `{"superseded": true, "id": int}`; admin, same token rule |
+| `GET`  | `/tasks/edges` | `?from_id=&to_id=&type=&limit=` | `{"edges": [...]}`; scoped to the token-bound project when registry auth is configured, tokenless requests return all edges |
+| `POST` | `/tasks` | body JSON `{"title", "body"?, "project"?, "assignee"?, "priority"?, "depends_on"?: [...], "created_by"}` | task object |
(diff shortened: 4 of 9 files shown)


jaylfc:
## BLOCKED: the project scope is dropped on the remote path, which is the one thing this PR exists to add

The local half of this is good work and I want to say that first, because the blocker is narrow and everything around it holds up.

### What is right

The scoping is done in SQL rather than by post-filtering, with two `EXISTS` subqueries requiring both endpoints in the token's project. That is a better shape than the `_enforce_edge_project_scope` call the card asked for, and I am not going to hold a deviation against you when the deviation is an improvement. I proved it is load-bearing rather than taking the tests' word for it: replacing the whole `if project is not None:` block with `if False: pass` (anchor asserted to match exactly once, `numstat` printed while mutated) fails `test_list_edges_cross_project_leak` while `test_list_edges_positive_control_cross_project_post_is_403` still passes. So the leak test dies for the right reason and the control shows the harness is live.

Suite on the trial merge onto master `4ece1c5c`: `1800 passed, 10 skipped, 7 errors` against master's `1793, 10, 7`. The delta of exactly +7 is this PR's 7 new tests, and the 7 errors are the pre-existing `taosmd.mcp_server.build_server` import failure on both sides. The negative-limit 400 and the `type=bogus` 400 both land, and the changelog fragment is present so the doc-gate is green.

### The blocker

`service.task_list_edges` accepts `project` and uses it on the local path, then drops it on the remote path:

```python
remote = _get_remote(data_dir)
if remote is not None:
    return await remote.task_list_edges(
        from_id=from_id, to_id=to_id, edge_type=edge_type, limit=limit
    )   # <- project is not forwarded
```

and `remote.task_list_edges` has no `project` parameter at all to forward it to. Every sibling in the family does have one. `service.task_list` forwards `project=project`, and `remote.task_list` accepts it and puts it in `params`.

Measured rather than read off the diff, with the sibling as a positive control so the probe cannot pass vacuously:

```
task_list        project forwarded = 'proj-a'    params=['assignee','limit','project','status']
task_list_edges  project forwarded = '<ABSENT>'  params=['edge_type','from_id','limit','to_id']

CONTROL task_list forwards project : True
task_list_edges forwards project   : False
```

This is not a hypothetical configuration. `_get_remote` returns a client whenever `server_url` is set in the data dir's `config.json`, which is the shared-server deployment, and the http_server always passes an explicit `data_dir` so it takes exactly that branch. In that setup the handler resolves the caller's project from the token binding, hands it to the service, and the service throws it away before the request leaves the box. The upstream request also travels under `server_token`, not the caller's token, so nothing downstream reconstructs the scope either. A proj-a token gets proj-b's edges, which is the exact leak this card was cut to close, still open in a supported deployment.

That is why it is a block and not a nit. The dominant defect shape on this repo is a control that looks installed and is inert, and a scope that is enforced on one path and silently absent on another is that shape precisely. Your own tests cannot see it, because they exercise the local path where the fix is real.

### What the revision needs

1. Add `project: str | None = None` to `remote.task_list_edges` and include it in `params` when set, matching `remote.task_list` exactly.
2. Forward `project=project` from `service.task_list_edges` to the remote call.
3. A test that pins it. The cheapest honest one is at the service layer with a recording fake remote, asserting `project` arrives, and with `task_list` in the same test as the positive control so a broken probe cannot report a pass. It must fail against this PR's head `08a33b3d`.

One smaller thing to fold in, not a blocker on its own: `.gitignore` gains `.venv/`, which is unrelated to task edges. It is a change I actually want, since `.venv` is currently invisible to git only because uv writes a nested ignore file, but it belongs in its own commit rather than riding a security fix.

Closing this so the card can redispatch, since a blocked PR has no revision route on this board. The branch is untouched and this review stands.

PR #406, opened 2026-08-28
A2A thread membership (unified-chat slice 2): rebuild ADDITIVELY, PR #237 deleted the service layer

jaylfc:
CARD TITLE (intent, not commit subject): A2A thread membership (unified-chat slice 2): rebuild ADDITIVELY, PR #237 deleted the service layer

Autonomous build of board card tsk-a3rwa4.

PR #237 replaced taosmd/service.py (1862 -> 191 lines), deleting all 46
existing functions. This rebuild adds the four membership functions to
master additively instead.

service.py: LOST none (mechanically asserted), GAINED a2a_create_thread,
a2a_list_members, a2a_add_member, a2a_remove_member.

Changes:
- taosmd/a2a_membership.py (new): MembershipStore + Membership dataclass,
  zero-loss membership tracking with archive integration.
- taosmd/migrations.py: register a2a_membership db with baseline migration.
- taosmd/remote.py: add a2a_create_thread, a2a_list_members, a2a_add_member,
  a2a_remove_member to RemoteClient (fixes nonexistent-method defect).
- taosmd/http_server.py: add POST/GET/DELETE routes for thread membership.
- taosmd/__init__.py: export MembershipStore.
- tests/test_a2a_membership.py (new): 35 tests covering store CRUD, service
  layer, and HTTP endpoints.

Defect fixes applied to salvaged code:
1. Falsy-timestamp bug: created_at/removed_at=0 now respected
   (was treated as absent by 'or time.time()').
2. MembershipStore(data_dir=None) resolves to default via _resolve_data_dir.
3. time module imported in a2a_membership.py.
4. RemoteClient gains the four methods service.py calls.

82 tests pass (35 membership + 47 migration).

Files:
 taosmd/__init__.py                       |   4 +
 taosmd/a2a_membership.py                 | 314 ++++++++++++++++++
 taosmd/http_server.py                    | 103 +++++-
 taosmd/migrations.py                     |  35 ++
 taosmd/remote.py                         |  60 ++++
 taosmd/service.py                        | 217 +++++++++++-
 tests/test_a2a_membership.py             | 554 +++++++++++++++++++++++++++++++
 8 files changed, 1284 insertions(+), 5 deletions(-)


jaylfc:
## Review: BLOCKED

Reviewed at head `d9d4484f`, trial-merged onto master `4ece1c5c` as `f261690d` in a detached worktree. Every number below was measured in that worktree, not carried in.

### Measurements

- Trial merge: **1830 passed / 10 skipped / 7 errors**. Clean master re-measured in the same venv: **1793 / 10 / 7**. The +37 is accounted for exactly: 35 tests in `tests/test_a2a_membership.py` plus 2 registry-parametrised cases in `test_migrations.py` (`[a2a_membership]`). Nothing was lost: `comm -13` on the collected node ids is empty. The 7 errors are the pre-existing `taosmd.mcp_server` import failures and appear identically on both sides.
- `uv run ruff check taosmd/ tests/ benchmarks/` (ruff 0.16.3, the literal CI command) rc=0.
- deleted-symbols, trailing-newline, normalise-handle and witness gates: clean. No em dashes in added lines (checked on `git diff | grep '^+'` with a positive control that fired).
- File modes on every changed path are `100644`, and the files API shows no deletions, so the additive rebuild claim holds.

### What is solid

The guards that exist are genuinely load-bearing. I disabled each one by line number and ran the new file:

| mutation | result |
|---|---|
| control, unmutated | 35 passed |
| `service.py:1981` add_member owner check -> `if False:` | 2 failed |
| `service.py:2024` remove_member owner check -> `if False:` | 2 failed |
| `service.py:2031` last-owner guard -> `if False:` | 2 failed |
| `service.py:1886` duplicate-thread guard -> `if False:` | 1 failed |
| `remove_membership` additionally DELETEs the row | 3 failed |

Five for five. The zero-loss property in particular is properly pinned.

---

### BLOCKER 1: all four remote forwards raise TypeError, and no test touches that path

`RemoteClient.a2a_*` declare `**_opts`, which cannot receive positional arguments, but `service` forwards `data_dir` positionally. Every one of the four operations is dead whenever a remote server URL is configured, which is `taosmd serve` and the hive deployment.

Proved with a `RemoteClient` pointed at a refused port, with an existing forwarded function as the positive control in the same run:

```
CONTROL   service.a2a_threads: URLError: <urlopen error [Errno 111] Connection refused>
SUBJECT   a2a_list_members : TypeError: RemoteClient.a2a_list_members() takes 2 positional arguments but 3 were given
SUBJECT   a2a_create_thread: TypeError: RemoteClient.a2a_create_thread() takes 4 positional arguments but 5 were given
SUBJECT   a2a_add_member   : TypeError: RemoteClient.a2a_add_member() takes 4 positional arguments but 5 were given
SUBJECT   a2a_remove_member: TypeError: RemoteClient.a2a_remove_member() takes 4 positional arguments but 5 were given
```

The control reaches the network, so the harness is wired and the four subjects die at call binding before any I/O. The house convention is keyword forwarding, e.g. `service.py:1486`:

```python
return await remote.task_list(status=status, project=project, assignee=assignee, limit=limit)
```

`grep -c remote tests/test_a2a_membership.py` is **0**, which is why a green suite says nothing here. This is the same shape as #405: the local path is fine and the remote path drops or mismatches its arguments.

Fix: forward by keyword, and add a test that drives each `service.a2a_*` with a fake remote and asserts the call arrives with the right arguments.

### BLOCKER 2: doc-gate is red, in CI and locally

```
DOC-GATE FAIL: a2a-handlers -- changes to the A2A handlers (taosmd/http_server.py, taosmd/service.py)
require docs coverage in taosmd/docs/a2a-comms.md
```

Negative control in the same session: the same gate on an empty diff (`--base HEAD`) returns rc=0 `doc-gate: clean`, so it is not failing blind. `taosmd/docs/a2a-comms.md` is untouched and there is no `Docs-Reviewed:` trailer on any commit. Four new public HTTP endpoints are documented only in the `http_server` module docstring table.

### BLOCKER 3: an authorisation denial is served as HTTP 500, and the tests pin it as the contract

`PermissionError` is not handled anywhere in `http_server.py`, so it falls into the catch-all and becomes a 500 plus a `logger.exception` on every ordinary denial. Two tests encode that as expected:

```python
def test_http_a2a_add_member_non_owner(live_server):
    """POST /a2a/threads/{thread}/members by a non-owner returns 500 (PermissionError)."""
    assert status == 500
```

A denial is not a server fault. Nothing downstream can tell "you are not an owner" from "the server broke", and every denial now pages as a 5xx. This repo already answers denials with 403 in the same file (registry auth, `http_server.py:1830`) and across `test_http_server_trust_enforcement.py`. Map `PermissionError` to 403 and flip both assertions.

### BLOCKER 4: ownership is self-asserted, and any caller can claim a thread that already exists

Two halves, both design-level, and they need settling before any read path trusts this store.

First, the owner check compares against `agent` taken from the request body, and nothing binds that to the caller. `_apply_token_binding` documents that it deliberately does not compare identity to `sub`, which is fine where the agent field names a target shelf, but here the same unbound string is the authorisation subject. The sibling A2A read path does bind, returning 403 on `reader != token_sub` (`http_server.py:1832`). As written, the owner check constrains an honest client and nothing else.

Second, `a2a_create_thread` only checks for membership rows, not for an existing conversation, so a principal who has never posted can take ownership of a live channel. Measured, with the duplicate guard as the in-run positive control:

```
existing messages on thread 'build': 1
stranger create on existing thread -> {'thread': 'build', 'created': True}
owners now: [('mallory', 'owner')]
CONTROL second claim refused: thread 'build' already exists
```

Thread names are the same namespace as the live fleet channels, so this is first come, first served on `build`. There is no impact today because nothing consults membership yet, but it persists into the slice that does.

Recommend: bind `agent` to the verified `sub` when a registry verifier is configured, and require either existing participation or the admin surface to create membership for a thread that already carries messages.

### FINDING 5: the archive integration is unverified, and the response claims it happened

I replaced both `archive_membership_created` and `archive_membership_removed` with no-ops. **All 35 tests still passed.** `a2a_remove_member` returns `"archived": True` with nothing checking that an event was written, and the zero-loss story for a re-add leans on exactly that archive, since re-adding clears `removed_at` and overwrites `created_at` so the row itself no longer records the removal.

Separately, each helper opens its own `ArchiveStore` and closes it, so creating a thread with N participants opens N+1 of them.

I did check whether these `EVENT_A2A` records leak into the message feed, and they do not: message counts on an existing thread are unchanged before and after a create, and a membership-only thread reads back 0 messages. That one is clean.

### Minor

- `test_remove_then_cannot_readd_same_role` asserts `result["added"] is True`, which is the opposite of what its name says. Rename it.
- `a2a_add_member` returns `already_member` before it checks ownership, so a non-owner can probe membership. Harmless while `GET /members` is unauthenticated, but the ordering should be flipped when auth lands.
- The registered migration never executes in practice, since `_init_schema` runs the same script first and the detect predicate then stamps the baseline. That matches all six existing stores, so it is house style and not a defect, but the schema now exists in two places and can drift.

### Scope note for whoever picks up slice 3

Nothing outside these four functions reads membership: `grep` for `MembershipStore`, `is_principal_member`, `is_principal_owner` and `has_any_membership` across `taosmd/` hits only `a2a_membership.py`, the four service functions and the `__init__` export. That is correct for an additive slice, but the changelog reads as though visibility is enforced. Say plainly that no read path is gated yet.

Card `tsk-a3rwa4` is closed against revision card **`tsk-v4cbhv`**, which starts from a fresh branch off master and carries a red-proof recipe for each defect. Blockers 1 through 3 are mechanical. Blocker 4 needs a decision, not just a patch.

PR #409, opened 2026-08-28
A2A v2 slice 2c REVISION: a2a_inbox_unhandled returns a false empty (limit applied before the ack filter)

jaylfc:
CARD TITLE (intent, not commit subject): A2A v2 slice 2c REVISION: a2a_inbox_unhandled returns a false empty (limit applied before the ack filter)

Autonomous build of board card tsk-2hnss2.

Re-apply the slice-2c work (a2a_inbox_unhandled bug, registry auth,
remote forwarding) onto a clean branch off origin/master.

Fixes:
- a2a_inbox_unhandled now pushes the ack filter into the a2a_inbox
  query via exclude_acked_by so limit applies to unhandled messages
  only; acked messages no longer consume the caller's budget
- a2a_inbox now includes acked_by on envelopes when present

Adds:
- GET /a2a/inbox, POST /a2a/inbox/advance, POST /a2a/ack,
  GET /a2a/inbox/unhandled HTTP handlers with registry auth
- RemoteClient methods for a2a_inbox, a2a_inbox_advance, a2a_ack,
  a2a_inbox_unhandled
- Changelog fragment

Tests:
- 15 new tests across test_a2a_inbox_unhandled, test_a2a_inbox_auth,
  test_a2a_remote_forwarding
- Red-first test asserts a2a_inbox_unhandled returns non-empty when
  acks exceed the limit budget; fails on unfixed code, passes after

Also fixes while in there:
- RemoteClient docstrings note that a2a_ack by and a2a_inbox_advance
  consumer are ignored on the remote path
- Bad-token tests pin 401 instead of in (401, 403)
- File header corrected: bad-signature tokens return 401, not 403

Suite: 1807 passed, 12 skipped

Files:
 taosmd/http_server.py                         | 135 ++++++++++++++
 taosmd/remote.py                              |  65 +++++++
 taosmd/service.py                             |  62 ++++++-
 tests/test_a2a_inbox_auth.py                  | 250 ++++++++++++++++++++++++++
 tests/test_a2a_inbox_unhandled.py             | 156 ++++++++++++++++
 tests/test_a2a_remote_forwarding.py           | 212 ++++++++++++++++++++++
 8 files changed, 887 insertions(+), 1 deletion(-)

### DIFF
--- a/changelog.d/tsk-2hnss2-a2a-inbox-endpoints.md
+++ b/changelog.d/tsk-2hnss2-a2a-inbox-endpoints.md
@@ -1,0 +1,4 @@
+### Added
+* A2A inbox HTTP endpoints (`GET /a2a/inbox`, `POST /a2a/inbox/advance`, `POST /a2a/ack`, `GET /a2a/inbox/unhandled`) with registry auth derived from the verified token `sub`; remote forwarding for `a2a_inbox`, `a2a_inbox_advance`, `a2a_ack`, and `a2a_inbox_unhandled`.
+### Fixed
+* `a2a_inbox_unhandled` now applies the `limit` after filtering out acknowledged messages, so acked messages no longer consume the caller's budget and cause false-empty results.
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -466,6 +466,10 @@ a2a_members(channel="CHANNEL")
 | `GET`  | `/a2a/stream` | `?thread=&since=` | SSE stream (`text/event-stream`) |
 | `GET`  | `/a2a/channels` | — | `{"channels": [...]}` |
 | `GET`  | `/a2a/members` | `?channel=<name>` | `{"members": [...]}` |
+| `GET`  | `/a2a/inbox` | `?consumer=&limit=&include_kinds=` | `{"messages": [...]}`; `consumer` is derived from the verified token `sub` when registry auth is configured, otherwise required as a query parameter |
+| `POST` | `/a2a/inbox/advance` | body JSON `{"to_id": int}` | `{"ok": true}`; consumer derived from verified token `sub` |
+| `POST` | `/a2a/ack` | body JSON `{"message_id": int}` | `{"id", "acked_by", "ok"}`; `by` derived from verified token `sub` |
+| `GET`  | `/a2a/inbox/unhandled` | `?consumer=&limit=` | `{"messages": [...]}`; composed query: messages past cursor minus acks |
 | `POST` | `/a2a/alarms/{key}/clear` | path-encoded alarm key | `{"cleared": true, "key": str}` |
 | `POST` | `/a2a/admin/delete-channel` | body JSON `{"channel": str}` | `{"deleted": true, "channel": str}`; admin, requires the admin token (403 if no admin or server token is set) |
 | `POST` | `/a2a/admin/rename-channel` | body JSON `{"from": str, "to": str}` | `{"renamed": true, "from": str, "to": str}`; admin, same token rule |
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -107,6 +107,10 @@ Endpoints
                             ``blocks``: optional list of arbitrary objects (no schema validation); when present, ``body`` must be non-empty
 ``GET  /a2a/messages``     ``?thread=&since=&limit=&fields=&format=``  -> ``{"messages": [...]}`` (``fields=id,sender,body`` projects keys; ``format=ndjson`` emits one message per line; ``since`` is an epoch timestamp in seconds, not a message id; values below 1e9 return 400)
 ``GET  /a2a/mentions``    ``?since=&limit=&reader=``                  -> ``{"messages": [...]}`` (requires registry auth; ``reader`` is derived from the verified token ``sub``, or supplied as a query parameter when no verifier is configured)
+``GET  /a2a/inbox``       ``?consumer=&limit=&include_kinds=``         -> ``{"messages": [...]}`` (``consumer`` is derived from the verified token ``sub`` when registry auth is configured; otherwise required as a query parameter)
+``POST /a2a/inbox/advance`` ``{"to_id": int}``                           -> ``{"ok": true}`` (principal derived from the verified token ``sub``)
+``POST /a2a/ack``          ``{"message_id": int}``                      -> ``{"id", "acked_by", "ok"}`` (``by`` derived from the verified token ``sub``)
+``GET  /a2a/inbox/unhandled`` ``?consumer=&limit=``                     -> ``{"messages": [...]}`` (composed query: mentions past cursor minus acks)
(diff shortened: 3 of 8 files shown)


jaylfc:
**BLOCKED.** The ordering defect this revision targets is genuinely fixed, and I proved it red before green. The block is that the endpoint the PR is named after ships an HTTP surface that no test touches at all, plus a silent parameter drop on the remote path of the same function this PR modifies.

## What is right, and should not be redone

The fix is real. `a2a_inbox_unhandled` does not exist on master, so the trivial red is the PR's own tests against master source (14 failed, 1 passed, the pass being the file's declared positive control). That proves nothing about ordering, so the prior implementation was reconstructed on the merged tree by line number, restoring limit-before-ack-filter at `service.py:1355` and `:1357`:

```
FAILED tests/test_a2a_inbox_unhandled.py::test_a2a_inbox_unhandled_limit_after_ack_filter
1 failed, 2 passed
```

The red-first test kills the exact prior shape. All four remote forwards bind (called at a refused port: `URLError`, never `TypeError`, zero bad bindings). Ack persistence is genuinely asserted rather than echoed: turning `service.py:1497 update_event_data_json` into `pass` kills 2 tests. The auth gates on `/a2a/inbox`, `/a2a/inbox/advance` and `/a2a/ack` are each load-bearing and individually pinned. The live route behaves correctly: 200 with the right two messages, 401 without a token, 403 on consumer mismatch, 401 on a second-keypair forgery, 400 on an unknown query param.

## Blocking 1: the whole HTTP surface of the new endpoint is deletable with CI green

Measured on the merged tree, with `/a2a/inbox` as the positive control in the same command:

```
inbox/unhandled hits in tests/ : 0
/a2a/inbox      hits in tests/ : 7
```

The only file naming the feature is `tests/test_a2a_inbox_unhandled.py`, and it calls the service function directly. Nothing drives the route. Consequently four separate mutations survive the full a2a plus http suite:

- `http_server.py:1095`, the `/a2a/inbox/unhandled` route dispatch, to `pass`
- `http_server.py:2139`, the **401 gate**, to `token_consumer = consumer_qp or "anon"`
- `http_server.py:2145`, the **403 consumer-mismatch gate**, to `if False:`
- `http_server.py:2131`, `_validate_a2a_params`, to `pass`

Both authentication gates on this endpoint can be deleted and every test and every gate still reads green. This is the repo's dominant defect shape, a control that looks installed and is inert, and here it is sitting on the flagship surface of the change. The remedy is small: four tests mirroring the ones that already exist for `/a2a/inbox`. Note also that the header of `tests/test_a2a_inbox_auth.py` says "401 on all three endpoints" while this PR adds a fourth.

## Blocking 2: `exclude_acked_by` is silently dropped on the remote path

`a2a_inbox` declares it at `service.py:1207` as a keyword-only public parameter, and the remote forward three lines below the docstring that calls the forward transparent does not pass it:

```python
remote = _get_remote(data_dir)
if remote is not None:
    return await remote.a2a_inbox(
        consumer, limit=limit, include_kinds=include_kinds,
    )
```

`RemoteClient.a2a_inbox` would swallow it into `**_opts` even if it were passed, so there is no error, only a different answer:

```
LOCAL  a2a_inbox("agent-1", exclude_acked_by="agent-1") -> [3]
REMOTE a2a_inbox("agent-1", exclude_acked_by="agent-1") -> [1, 2, 3]   # 1 and 2 are acked
```

No in-repo caller reaches it today, because `a2a_inbox_unhandled` short-circuits to its own endpoint first, so this is latent rather than live. It is still a new public parameter on an exported function that silently answers wrong in remote mode, which is the same failure direction the previous revision was blocked for. Either forward it, which needs a server-side `exclude_acked_by` query parameter, or raise on it in remote mode.

## Blocking 3: two claimed additions have no test

- `service.py:1295`, the `acked_by` envelope propagation, is the PR body's claim that "a2a_inbox now includes acked_by on envelopes when present". Turning `if "acked_by" in data:` into `if False:` survives the wide suite.
- `service.py:1350`, the `a2a_inbox_unhandled` remote forward, is never called by any test. It had to be called by hand to establish that it works.

## Non-blocking

- **Standalone mode.** With no registry verifier, `POST /a2a/inbox/advance` and `POST /a2a/ack` return 401 unconditionally and are unreachable, while the two GET routes work from a bare query parameter. This matches the pre-existing receipts handlers at `http_server.py:1942` and `:1965`, unchanged here, so it is consistent prior art rather than a new decision. It is still new dead surface in the default local-first deployment, and untested.
- **A test comment is wrong.** `test_a2a_inbox_unhandled_limit_after_ack_filter` says the buggy code "returns 0 because limit is applied before the ack filter". With the cursor advanced past `ids[0]`, only 7 of the 8 acked messages fall inside the first 8 of the window, so the buggy path returns 1, not 0. The assertion still fires; only the comment misdescribes it.
- `service.py:1352`, the `consumer` validation in `a2a_inbox_unhandled`, is unreachable, because `a2a_inbox` raises the identical `ValueError` one line later. Harmless duplication.
- `http_server.py:2046`, `_validate_a2a_params` on the otherwise well-covered `/a2a/inbox`, is also deletable.
- The two "valid token succeeds" tests build a `tmp_path` data dir, stores and a fake embedder that the request path never uses, since requests go to `authed_server`'s own data dir. Dead setup, not a correctness problem.

## Numbers and gates

Full suite on the committed trial merge: **1808 passed, 10 skipped, 7 errors**. Clean master measured in the same worktree: **1793 / 10 / 7**, which matches my own independent direct measurement of master `5ef2bf72` today. The delta is exactly **+15**, matching the 15 new tests, with skips unchanged, which rules out the `jwt` and `cryptography` `importorskip` pair silently swallowing the two auth files. The PR body claims 1807 passed and 12 skipped; the absolute numbers differ by optional-dependency availability in the environment, and the delta agrees, so this is not a discrepancy in the work.

All six gates rc=0 with rc captured directly, and two were proven live rather than assumed: `check_deleted_symbols.py` returns rc=1 `DELETED-SYMBOLS FAIL` when a deletion of `a2a_inbox_advance` is committed, and `check_doc_gate.py` returns rc=1 `DOC-GATE FAIL: a2a-handlers` when the doc and changelog edits are reverted. Zero em dashes in 895 added lines, with a positive control confirming the grep fires.

**One methodology note worth carrying, because it nearly changed this verdict.** The first wide mutation pass read `rc=1` for four of the survivors above and they were nearly recorded as killed. The rc came entirely from the pre-existing `test_mcp_server.py` import errors, with zero FAILED lines. On this repo a mutation verdict has to be a FAILED count, never a pytest exit status, because the 7 pre-existing errors make rc permanently 1. Relatedly, the doc gate reads committed history, so an uncommitted revert reads clean straight through it.

This is good work with a real fix in it. The gap is that the new endpoint's HTTP surface, including both of its auth gates, is currently unpinned.

jaylfc:
Closing as superseded. The blocking review above stands, and the revision was cut as `tsk-a526k2` from a fresh branch off master rather than from this branch, so there is nothing further to push here. The work is not lost: this branch is untouched and the card carries the full blocking evidence.

PR #411, opened 2026-08-28
REVISE PR #406 (A2A thread membership, unified-chat slice 2): all four remote forwards raise TypeError, doc-gate is red, and an ownership denial is served as HTTP 500

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #406 (A2A thread membership, unified-chat slice 2): all four remote forwards raise TypeError, doc-gate is red, and an ownership denial is served as HTTP 500

Autonomous build of board card tsk-v4cbhv.

Rebuilds four defects on top of the MembershipStore from PR #406
(cherry-picked commit d9d4484f) on a fresh branch off master.

Baseline re-measured on this worktree (master 5ef2bf72 + store
cherry-pick, before fixes): 1830 passed, 10 skipped, 7 errors.
After fixes: 1836 passed, 10 skipped, 7 errors (same errors; all new
tests pass).
Suite recipe: uv sync --all-groups --all-extras, then
./.venv/bin/python -m pytest -q -p no:randomly.
taosmd imported from: /tmp/exec-tsk-v4cbhv/taosmd/__init__.py

Defect 1 (TypeError on remote forwarding): service.a2a_create_thread,
a2a_list_members, a2a_add_member, a2a_remove_member forwarded data_dir
positionally to RemoteClient, whose methods declare **_opts (keyword-only
catch-all). Every remote call raised TypeError. Fixed by forwarding
data_dir=data_dir as a keyword, matching the house convention at
service.py:1486 (remote.task_list(status=..., ...)).

RED proof -- tests/test_a2a_remote_forwarding.py drives each of the four
service.a2a_* functions with a recording fake remote and asserts the call
arrives with data_dir carried as a keyword. On unpatched code: 4 failed,
1 passed (a2a_threads positive control). After fix: 5 passed.
  RC=1 (4 TypeError failures)
  RC=0 (5 passed)

Defect 2 (doc-gate red): The four new endpoints in http_server.py and
service.py were undocumented in taosmd/docs/a2a-comms.md. Added a Thread
membership section and four rows to the HTTP endpoints table.

Verification (empty-diff control in same command):
  diff-gate --staged -> doc-gate: clean, RC=0
  diff-gate --base HEAD -> doc-gate: clean, RC=0 (empty-diff control)

Defect 3 (PermissionError -> HTTP 500): PermissionError from
a2a_add_member/a2a_remove_member ownership denials fell through to the
catch-all except handler and was served as HTTP 500 with a logger.exception
on every ordinary denial. Added an except PermissionError handler mapping
to 403, matching the registry-auth 403 pattern around http_server.py:1830.
Flipped test_http_a2a_add_member_non_owner and
test_http_a2a_remove_member_non_owner from 500 to 403.

RED proof (403 assertions, handler not yet added):
  assert 500 == 403 (x2), RC=1
After fix: 2 passed, RC=0

Defect 4 (design question, no patch): Ownership is self-asserted from the
request body's agent field with no binding to the caller's verified token,
unlike the sibling /a2a/mentions read path which binds reader to token sub
(http_server.py:1834). Additionally, a2a_create_thread only checks membership
rows, not the archive, so a stranger can claim ownership of a thread that
has message history but no membership rows. Full assessment in
docs/a2a-membership-auth-assessment.md. No auth behaviour changed.

Defect 5 (unverified archive integration): Replacing
archive_membership_created and archive_membership_removed with no-ops left
all 35 tests passing while a2a_remove_member still returned
archived=true. Added test_archive_event_lands_on_add_and_remove which
queries the ArchiveStore for EVENT_A2A entries with admin_action
membership_created / membership_removed after add and remove operations.

RED proof (archive functions made no-ops):
  AssertionError: assert 'membership_created' in [], RC=1
After restore: 1 passed, RC=0

Smaller, folded in:
- Renamed test_remove_then_cannot_readd_same_role to
  test_remove_then_readd_same_role (it asserts added is True, the
  opposite of cannot).
- In a2a_add_member, ownership check now runs before the already_member
  early return, so a non-owner cannot probe membership state.
- changelog.d/tsk-v4cbhv-a2a-membership-defects.md explicitly states no
  read path is gated by membership yet: nothing outside the four service
  functions reads the MembershipStore.

Signed-off-by: jaylfc <jaylfc25@gmail.com>

Files:
 taosmd/docs/a2a-comms.md                         |  33 ++
 taosmd/http_server.py                            | 106 ++++-
 taosmd/migrations.py                             |  35 ++
 taosmd/remote.py                                 |  60 +++
 taosmd/service.py                                | 221 ++++++++-
 tests/test_a2a_membership.py                     | 580 +++++++++++++++++++++++
 tests/test_a2a_remote_forwarding.py              | 111 +++++
 12 files changed, 1530 insertions(+), 5 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added A2A thread creation and membership management.
  * Owners can list, add, and remove thread members.
  * Added local and remote support for membership operations.
  * Protected threads from duplicate creation and removal of the last owner.
* **Bug Fixes**
  * Improved remote request handling and permission error responses.
* **Documentation**
  * Documented membership roles, authorization rules, endpoints, and current limitations.


jaylfc:
**BLOCKED**, on a CI lint failure this PR introduces in its own test file, plus a set of blind branches that a mutation pass exposed. The three defects the revision set out to fix are all genuinely fixed, and I verified each one separately by execution rather than by reading the diff.

## The three claimed fixes: all three hold

**(i) The four remote forwards no longer raise TypeError.** Called every one against a genuinely refused port, where `URLError` means the call bound and `TypeError` means it did not, both directly and through the real service call sites:

```
a2a_create_thread  -> URLError    a2a_add_member     -> URLError
a2a_list_members   -> URLError    a2a_remove_member  -> URLError
```

Zero TypeErrors. Each of the four `if remote is not None:` branches was also individually killed by a distinct test, so all four are pinned.

**(ii) The doc gate is green, and the green is a reading.** rc captured directly, run in-repo: `diff-gate --base origin/master` rc=0. The `--base HEAD` empty-diff control is also rc=0 and therefore proves nothing on its own, so a positive control was run: stripping the `taosmd/docs/a2a-comms.md` edit and both changelog fragments and re-running gives rc=1 with both `DOC-GATE FAIL: changelog` and `DOC-GATE FAIL: a2a-handlers`. Both rules that this PR's paths fire do read this diff and do go red without the docs.

**(iii) Ownership denial is a deliberate 403, not a crash.** Booted the real server on an ephemeral port and read statuses off the real routes: non-member add 403, member-but-not-owner add 403, non-member remove 403, member remove-owner 403, last-owner removal 400, duplicate create 400. `except PermissionError` sits between `except ValueError` and `except Exception` at `http_server.py:1247`, and since `PermissionError` is an `OSError` subclass the `ValueError` arm above does not shadow it. No existence leak: a denial on a thread that does not exist returns the same 403 with the same wording as one on a thread that does. Mutating the arm away, and mutating 403 to 500, were both killed.

## The blocking defect: CI lint is red

`.github/workflows/ci.yml:21` runs exactly this, and I ran it with rc captured directly:

```
$ uv run ruff check taosmd/ tests/ benchmarks/ --output-format=concise
tests/test_a2a_remote_forwarding.py:110:16: F841 Local variable `opts` is assigned to but never used
Found 1 error.
RUFF_RC=1
```

Not pre-existing. The same command on master `5ef2bf72` with the same pinned `ruff 0.16.3` gives `All checks passed!`, rc=0.

This is not cosmetic, and the fix is not to delete the variable. The offending line is inside `test_a2a_threads_positive_control`, the test whose own docstring calls it the positive control for the whole file:

```python
principal, opts = patched_remote.calls[0][1], patched_remote.calls[0][2]
assert principal == "carol"
# opts is never asserted
```

The other four tests in that file each end with `assert opts.get("data_dir") == DD`. The control is the one test that drops that assertion, so it does not control for the thing the file exists to test. Deleting `opts` to satisfy the linter would make the control weaker still. **Add `assert opts.get("data_dir") == DD`.**

## Survivors worth fixing

56 mutations, one per logical branch, applied by line number with every anchor verified against its target line first. 35 killed. Every ACL, ownership gate and membership check is genuinely pinned by a test that reaches it. These survived:

- **M40, `service.py:1988`.** Changing `add_membership(..., role="member")` to `role="owner"` leaves the suite green. `a2a_add_member` could grant **owner** to every principal it adds and nothing reds. No test anywhere asserts the role of a member added through this path.
- **M32, `service.py:1891`.** Turning `if participant != agent:` into `if True:` leaves the suite green, and because `add_membership` upserts with `role = excluded.role`, it **downgrades the creator from owner to member** whenever the caller appears in `participants`. `test_create_thread_caller_not_in_participants` passes `carol` in the list and then asserts only `ids.count("carol") == 1`, the count and never the role. The test named for this exact case cannot see an ownership downgrade.
- **M48 and M49.** `archive_membership_created` can be made a no-op at the `a2a_add_member` call site (`service.py:1991`) and at the create-thread owner call site (`service.py:1895`), both with the suite green. Cause: `test_archive_event_lands_on_add_and_remove` asserts only `"membership_created" in actions` across all events, and the unmutated participants loop already emits one before `a2a_add_member` is reached. The changelog says a no-op recording now fails the new test. That is true for the **removed** half and for the participants loop, and false for these two call sites. This is the same shape as the #406 miss, one call site over. Filter the assertion on `principal_id`.
- **M14 / M12 / M05**, `a2a_membership.py:196`, `:175`, `:116`. Dropping `AND removed_at IS NULL` from `is_principal_owner`, from `get_thread_owners`, and from `remove_membership` all leave the suite green. The shipped code is correct in each case; the branches are untested. The first would let a **removed** owner keep passing the ACL, the second would let removed owners count toward the last-owner guard, and the third would let a re-removal overwrite the original `removed_at`, which the module's own docstring forbids as a zero-loss violation.
- **M16 to M19.** `is_principal_member` and `has_ownership` can each be made to `return True` unconditionally. A repo-wide grep finds only their definitions: zero callers in product code and zero tests. With `count_active_members` (tests only), roughly 35 of the 315 new lines are unreachable. Either wire them into the read path they were written for, or drop them.
- **M53 to M56**, `http_server.py:1099/1112/1115/1125`. All four routing guards are untested. Live probes confirm the shipped guards behave, but nothing pins them.

## Two findings from live probing, not visible in the diff

**Path segments are never percent-decoded, and the failure returns HTTP 200.** Against the live server:

```
DELETE /a2a/threads/t/members/%40bob -> 200 {"principal_id":"%40bob","removed":false,"not_found":true}
GET    /a2a/threads/t/members        -> @bob is still a member
DELETE /a2a/threads/t/members/@bob   -> 200 {"removed":true,"archived":true}
```

`http_server.py:1124` slices `principal` out of `parts.path` with no `unquote()`, and `remote.py:433` builds the URL with an f-string and no `quote()`, so the same bug rides the remote forward. Any client or proxy that encodes `@`, or any handle containing `/`, `#`, `?` or a space, gets a failed removal that reads as success at the status-code layer. Also affects `thread` on all four routes. Untested, which is M55.

**The registered migration never runs.** `MembershipStore._init_schema` (`a2a_membership.py:63-65`) runs `executescript` with an identical copy of the DDL before calling `migrations.migrate(...)`, so `migrate()` always takes the `stamped_baseline` branch. Driven directly, the migration is correct: it applies on a fresh unstamped DB, creates all three indexes including the partial `WHERE removed_at IS NULL` one, is idempotent on re-run, and stamps a legacy DB without replaying over live data. It simply never fires on a real code path. The two DDL copies are byte-equivalent today so nothing is wrong right now, but this is the latent shape that has bitten this repo before, and nothing goes red the moment they drift. Drop the inline schema string and let `migrations.migrate()` own the DDL.

## Smaller

- `a2a_membership.py:314` has a duplicated `await archive.close()`. Harmless, since `ArchiveStore.close` is idempotent, but it is a copy-paste artifact.
- `docs/a2a-membership-auth-assessment.md` says "the four membership write endpoints" and lists three.
- One em dash in the added lines, `taosmd/docs/a2a-comms.md:499`, as an empty markdown table cell. That file already uses the same convention four times on master, so it follows the table rather than introducing prose style. Flagging only because the house rule is absolute; `n/a` is the consistent alternative.
- `GET /a2a/threads/{t}/members` has no membership check, so any caller can read a private thread's roster. The `server_token` data-plane gate at `http_server.py:1023` does cover it when a token is configured, and the PR discloses this in its own changelog fragment. Noted for the record, not blocking.

## Numbers

Full suite on the committed merge: **1836 passed, 10 skipped, 7 errors**. Baseline re-measured directly on master `5ef2bf72` today: **1793 / 10 / 7**, so +43 passing. The review counted 41 tests across the two new files, which leaves a 2-test discrepancy that was derived from `--collect-only` arithmetic rather than measured. Please state the added-test count directly in the revision so the delta reconciles. The 7 errors are the pre-existing `taosmd.mcp_server` missing-SDK failure and are unrelated. Gates: deleted-symbols, normalise-handle, witness-token and trailing-newline all rc=0. The `-1` and `-4` deletions in `service.py` and `http_server.py` are re-indentation under the new `/members` routing split and an `__all__` bracket move; no symbol is removed, and there is no typechange or mode change on any of the 12 paths.

To unblock, item 1 is required. Items on M40, M32, M48 and M49 are strongly recommended, because each is a privilege or persistence branch that is currently free to break silently.

jaylfc:
Closing as superseded. The blocking review above stands, and the revision was cut as `tsk-6pned6` from a fresh branch off master rather than from this branch, so there is nothing further to push here. The work is not lost: this branch is untouched and the card carries the full blocking evidence.

PR #412, opened 2026-08-28
Add POST /a2a/import: idempotent batch import of external chat envelopes onto the A2A bus

jaylfc:
CARD TITLE (intent, not commit subject): Add POST /a2a/import: idempotent batch import of external chat envelopes onto the A2A bus

Autonomous build of board card tsk-gbycx2.

- Accepts a JSON array of envelope dicts, each requiring source, source_id,
  from, body, and ts (epoch float)
- Deduped on (source, source_id) using a2a_import_dedup table so re-imports
  are no-ops
- Original timestamps preserved verbatim in archive.record via new timestamp
  parameter
- Whole batch rejected if any envelope is invalid (composes with exporters
  that pre-validate)
- Registry auth gate consistent with POST /a2a/send
- Added 8 tests covering route existence, whole-batch rejection, timestamp
  preservation, idempotence (including dedup-break failure test), and auth
  parity
- Documented endpoint in docs/a2a-comms.md and added changelog fragment

Proof: uv run --extra dev pytest -q -p no:randomly tests/test_a2a_import.py
rc=0, 8 passed. Full suite: 1800 passed, 12 skipped, 0 errors.

Files:
 changelog.d/tsk-gbycx2-a2a-import.md |   3 +
 taosmd/archive.py                    |  30 +++-
 taosmd/docs/a2a-comms.md             |  19 +-
 taosmd/http_server.py                | 179 +++++++++++--------
 taosmd/remote.py                     |   7 +
 taosmd/service.py                    | 188 +++++++++++++++++++-
 tests/test_a2a_import.py             | 327 +++++++++++++++++++++++++++++++++++
 7 files changed, 673 insertions(+), 80 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Added batch import for external chat history through `POST /a2a/import`.
  - Preserves original message timestamps and prevents duplicate imports.
  - Validates the complete batch before importing; invalid batches are rejected.
  - Added client support for submitting import batches and receiving import results.

- **Documentation**
  - Added usage guidance and examples for importing chat backlogs.

- **Bug Fixes**
  - Improved consistency of authentication checks for A2A messaging operations.

### DIFF
--- a/changelog.d/tsk-gbycx2-a2a-import.md
+++ b/changelog.d/tsk-gbycx2-a2a-import.md
@@ -1,0 +1,3 @@
+### Added
+
+- `POST /a2a/import`: idempotent batch import of external chat envelopes onto the A2A bus. Accepts a JSON array of envelope dicts, each requiring `source`, `source_id`, `from`, `body`, and `ts`. Deduped on the pair `(source, source_id)` so re-importing the same channel produces no duplicates. Original timestamps are preserved verbatim. The whole batch is rejected if any envelope is invalid, composes with exporter clients that pre-validate. Uses the same registry auth gate as `POST /a2a/send`.
--- a/taosmd/archive.py
+++ b/taosmd/archive.py
@@ -91,7 +91,16 @@ CREATE TABLE IF NOT EXISTS a2a_alarm_state (
 CREATE INDEX IF NOT EXISTS idx_a2a_alarm_state_key_fp ON a2a_alarm_state(alarm_key, fingerprint);
 """
 
-INDEX_SCHEMA = INDEX_SCHEMA + A2A_ALARM_STATE_SCHEMA
+A2A_IMPORT_DEDUP_SCHEMA = """
+CREATE TABLE IF NOT EXISTS a2a_import_dedup (
+    source TEXT NOT NULL,
+    source_id TEXT NOT NULL,
+    PRIMARY KEY (source, source_id)
+);
+CREATE INDEX IF NOT EXISTS idx_a2a_import_dedup_source_id ON a2a_import_dedup(source, source_id);
+"""
+
+INDEX_SCHEMA = INDEX_SCHEMA + A2A_ALARM_STATE_SCHEMA + A2A_IMPORT_DEDUP_SCHEMA
 
 
 class ArchiveStore:
@@ -201,6 +210,22 @@ class ArchiveStore:
         )
         self._conn.commit()
 
+    async def record_import_dedup(self, source: str, source_id: str) -> None:
+        self._conn.execute(
+            "INSERT OR IGNORE INTO a2a_import_dedup (source, source_id) VALUES (?, ?)",
+            (source, source_id),
+        )
+        self._conn.commit()
+
+    async def get_import_dedup(self, source: str, source_id: str) -> dict | None:
+        row = self._conn.execute(
+            "SELECT source, source_id FROM a2a_import_dedup WHERE source = ? AND source_id = ?",
+            (source, source_id),
+        ).fetchone()
+        if row is None:
+            return None
+        return {"source": row[0], "source_id": row[1]}
+
     # ------------------------------------------------------------------
     # Recording
     # ------------------------------------------------------------------
@@ -237,6 +262,7 @@ class ArchiveStore:
         app_id: str | None = None,
         summary: str = "",
         project: str | None = None,
+        timestamp: float | None = None,
     ) -> int:
         """Record an event to the archive. Returns the index row ID."""
         # Skip user activity events if tracking is disabled
@@ -250,7 +276,7 @@ class ArchiveStore:
             if key in data and isinstance(data[key], str):
                 data[key], _ = redact_secrets(data[key])
 
-        ts = time.time()
+        ts = timestamp if timestamp is not None else time.time()
         event = {
             "timestamp": ts,
             "event_type": event_type,
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -178,12 +178,21 @@ Your ID: AGENT_NAME          (replace with this agent's own name/identifier)
      -H "Content-Type: application/json" \
      -d '{"from": "AGENT_NAME", "body": "your message here", "thread": "CHANNEL"}'
 
-   To reply to a specific message (use its id field):
-   curl -X POST SERVER_URL/a2a/send \
-     -H "Content-Type: application/json" \
-     -d '{"from": "AGENT_NAME", "body": "reply text", "thread": "CHANNEL", "reply_to": "MESSAGE_ID"}'
+    To reply to a specific message (use its id field):
+    curl -X POST SERVER_URL/a2a/send \
+      -H "Content-Type: application/json" \
+      -d '{"from": "AGENT_NAME", "body": "reply text", "thread": "CHANNEL", "reply_to": "MESSAGE_ID"}'
 
---- Reading messages ---
+    To import a backlog of messages from an external chat system, preserving
+    original timestamps and deduplicating on (source, source_id):
+    curl -X POST SERVER_URL/a2a/import \
+      -H "Content-Type: application/json" \
+      -d '[{"source": "CHANNEL", "source_id": "EXTERNAL-ID", "from": "SENDER", "body": "message text", "ts": 1709452800.0}]'
+
+    The whole batch is rejected if any envelope is invalid. Re-importing the
+    same (source, source_id) pair is a no-op.
+
+ --- Reading messages ---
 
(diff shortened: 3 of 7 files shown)


jaylfc:
Review of #412 (`tsk-gbycx2`), for which I am the named reviewer.

**Verdict: not blocked.** One revision item below, from a mutation campaign rather than a diff read.

I came into this holding three bar items from my own notes. Two of them do not survive contact with the implementation, and I am recording that rather than quietly dropping them.

**1. The whole-batch objection was wrong, and the reason matters.** I had it written down that per-envelope acceptance plus dedupe on `(source, source_id)` leaves a half-landed batch permanently uncompletable, because the retry dedupes away the half that already landed. That is false here. Validation runs over the entire batch and raises before any write, so a malformed batch writes nothing; and inside the write loop the order is `archive.record(...)` first, `record_import_dedup(...)` second. Record-then-mark is the safe direction. A crash between the two leaves an envelope landed but unmarked, so a retry re-imports it: at-least-once, a possible duplicate, never a loss. The batch always completes. Worth a line in the docstring that the crash window yields a duplicate rather than a drop, since the current text implies exact-once.

**2. The dedupe is genuinely pinned, so the vacuous-red concern does not apply.** My worry was that "red before the handler exists" is satisfied by any test while the route still 404s, and so cannot distinguish a working dedupe from an absent one. Checked by mutation instead of by reading: stubbing `get_import_dedup` to always return `None` gives `1 failed, 7 passed` against a `8 passed` control. `test_http_a2a_import_idempotence_fails_when_dedup_is_broken` is the builder's own mutation test, which is the right instinct and I want it noted.

**3. `source_id`s in the error text.** `batch[i].source_id must be a non-empty string` names the index, which is the only usable locator when the field itself is the thing that is missing. Correct as written.

**The one revision item, which no diff read would have surfaced:**

The per-field validation gate is live on `ts` and blind on `source_id`.

```
control                                    8 passed
get_import_dedup -> None                   1 failed, 7 passed   (killed)
_check_a2a_registry_auth -> True           1 failed, 7 passed   (killed)
source_id non-empty check -> if False:     8 passed             (SURVIVED)
```

`test_http_a2a_import_rejects_whole_batch_on_bad_envelope` proves the whole-batch property properly, including `len(messages) == 0` on the store afterwards, but it does it with a bad `ts` only. Every other validation branch in that loop rides on that one test and none of them is pinned. Deleting the `source_id` check outright leaves the suite fully green.

The whole-batch behaviour is proven, so this is not a block. What is needed is one parametrised case per rejected field rather than a single representative envelope, since a per-file mutation pass reports this gate as healthy and only a per-branch one finds it.

Structurally clean otherwise: all four touched modules are mode `100644`, no symbols removed, and the 72 removed lines in `http_server.py` are re-indentation with no `def` or `class` among them.

Not merging this myself. The #407 ruling was per-PR and did not grant standing authority.

jaylfc:
Blocking. The registry auth gate on the new endpoint is live overall but blind on the issuer-pinning branch, so `/a2a/import` accepts a token that `/a2a/send` rejects on the same running server.

**Defect**

`taosmd/http_server.py:1748` `_handle_a2a_import` calls `_check_a2a_registry_auth()` with no `claimed_from`. That drives the `else` branch at `taosmd/http_server.py:1690-1699`, which re-implements `registry_auth.authorize_sender` inline: signature verify, `sub` present, revocation set. It never applies `expected_iss`.

`registry_auth.authorize_sender` (`taosmd/registry_auth.py:96`) ends with the issuer check, and issuer pinning is the production default: `verifier_from_url` takes `expected_iss: str | None = REGISTRY_ISS` at `taosmd/registry_auth.py:400`. Every other registry-authed path goes through `authorize()` and therefore keeps that check, `/a2a/mentions` on master included. This endpoint is the one that skips it.

The reason no test sees it is `tests/test_a2a_import.py:279`: the `authed_server` fixture builds the verifier with `expected_iss=None`, which opts out of the exact control that is missing. `test_http_a2a_send_and_import_share_auth_gate` then compares send and import only on the missing-token branch, so it reports parity that does not hold.

**Failing input**

A token signed by the correct registry key with `{"sub": "agent-1", "iss": "evil-issuer"}`, against a server whose verifier is built with the default `expected_iss=registry_auth.REGISTRY_ISS` and `a2a_auth_enforce=True`.

**Transcript**

Merged PR head `fc793133` into `origin/master` `e0429aca` in a scratch worktree. The probe is the PR's own `authed_server` fixture with one line changed, `expected_iss=registry_auth.REGISTRY_ISS`, plus two positive controls on the same server and the same key.

```
tests/test_probe_iss.py::test_positive_control_good_iss_is_accepted_on_send PASSED
tests/test_probe_iss.py::test_positive_control_bad_iss_is_rejected_on_send  PASSED
tests/test_probe_iss.py::test_bad_iss_must_also_be_rejected_on_import       FAILED

>       assert status in (401, 403), f"IMPORT ACCEPTED a wrong-issuer token: {status} {body}"
E       AssertionError: IMPORT ACCEPTED a wrong-issuer token: 200 {'imported': 1, 'skipped': 0,
E         'messages': [{'id': 1, 'ts': 1709452800.0, 'source': 'ch', 'source_id': '1',
E         'from': 'agent-1', 'thread': 'ch', 'reply_to': None, 'kind': 'chat'}]}

1 failed, 2 passed in 1.87s
```

Control two proves the pin is a live control on that exact server and rejects the token at `/a2a/send`. Control one proves the server accepts the same principal when the issuer is right, so the rejection is the issuer and nothing else. The message was written to the bus, not merely accepted.

**Fix**

Route the no-`claimed_from` branch through the shared policy rather than a second copy of it, or at minimum apply `_registry_verifier._expected_iss` in that branch. Then re-point the `authed_server` fixture at `registry_auth.REGISTRY_ISS` so the parity test covers the issuer branch and not only the missing-token branch.

**Second item, also PR-introduced**

The `test` check is red at the Lint step, which means the Tests step never ran in CI at all. All five errors are in the new file:

```
F401 tests/test_a2a_import.py:25:63  Ed25519PrivateKey imported but unused
F401 tests/test_a2a_import.py:26:44  serialization imported but unused
F401 tests/test_a2a_import.py:253:19 jwt imported but unused
F811 tests/test_a2a_import.py:254:67 redefinition of Ed25519PrivateKey from line 25
F811 tests/test_a2a_import.py:255:48 redefinition of serialization from line 26
```

`ruff check taosmd/ tests/ benchmarks/ --exclude tests/test_a2a_import.py` is clean on the merged tree, so this is entirely new here. Note that line 24 imports `jwt as pyjwt` at module scope and the fixture re-imports it at 253, so the module-level name is the one the tests actually use and the local imports are dead.

**Minor, not blocking**

`_read_json_body_any` (`taosmd/http_server.py:796`) reads `Content-Length` bytes with no ceiling. `_read_json_body` has the same shape on master so this is not a regression, but the import batch is also unbounded in length: `_A2A_MAX_REFS` caps refs at 8 and each envelope is capped at 64KB, while the number of envelopes in a batch is not capped at all.

**What checked out**

Suite on the merged tree is 1802 passed, 10 skipped, 7 errors, against a master baseline of 1794/10/7. The delta is exactly the 8 new tests and the 7 errors are the pre-existing `taosmd.mcp_server` optional-dependency failure. Stubbing `get_import_dedup` to return `None` fails `test_http_a2a_import_is_idempotent` (1 failed, 7 passed), so the dedupe gate has teeth. The write order in `service.a2a_import` is `archive.record()` then `record_import_dedup()`, which is at-least-once: a crash between them duplicates on retry and never drops a message. That is the right choice for an importer.

The per-field envelope validation is pinned on `ts` only, since one representative bad envelope exercises every branch. On its own that is a coverage gap rather than a correctness problem and would not have blocked. The issuer branch above is a different thing: it is a control that is absent from the code, not merely untested.

jaylfc:
Closing this in favour of a fresh re-land rather than a fix-forward on this branch.

The blocking defect is in the review comment above: the import auth path omits the issuer check that its sibling `/a2a/mentions` keeps, so a token from another issuer was accepted and the message was written to the bus. Alongside that, the `test` check was red at Lint, and because Lint runs before Tests the test step never executed on this PR at all.

Both of those are in the auth and CI shape of the branch rather than in one hunk, and the parity test that should have caught the first one was built against a verifier with issuer pinning disabled. Re-deriving from master is cleaner than patching that.

Successor card: `tsk-bdj4fv`, which re-lands the feature from a fresh branch off master with per-branch auth parity tests, an explicit wrong-issuer arm, and a mutating acceptance on each branch. Two findings from this PR were confirmed correct and are carried into that card: the record-then-mark write order (at-least-once, correct for an importer), and the dedupe gate, which has teeth.

The executing card `tsk-gbycx2` is closed with the same reason. Nothing here is lost, and no work on this branch is needed.

PR #413, opened 2026-08-28
REVISE the Postgres assessment (#408): the inventory has a phantom store, an omitted store and two wrong filenames

jaylfc:
CARD TITLE (intent, not commit subject): REVISE the Postgres assessment (#408): the inventory has a phantom store, an omitted store and two wrong filenames

Autonomous build of board card tsk-srpeiy.

- add missing insights.db (InsightStore at taosmd/reflect.py:130, init at
  taosmd/auto_setup.py:122-124)
- remove phantom pending-decisions.db, restate as kg_pending_decisions table
  inside knowledge-graph.db at taosmd/pending_decisions.py:43
- correct mentions.db to a2a-mentions.db and receipts.db to a2a-receipts.db
  per taosmd/api.py:148, taosmd/api.py:151, taosmd/migrations.py:421
- unify vector-memory.db as OPTIONAL / Local-Only, remove from Core Memory
  Storage in section 1.1
- replace truncating inventory command (ended in head -N) with a re-runnable
  grep that produces the full row set
- replace aggregate 'All Other Required Stores' verdict with per-store verdict
  and evidence for all 14 rows
- correct LIMIT citation from tsk-7wau2y to tsk-2hnss2, anchor to
  taosmd/archive.py:380 instead of 'Various query methods'
- remove unsourced 'approximately 40 percent higher' figure and six-week
  timeline; keep 1793 passed / 10 skipped / 7 errors baseline

Proof: doc-gate invariants and diff-gate against origin/master are clean

Files:
 .../tsk-srpeiy-revise-postgres-assessment.md       |   6 +
 docs/postgres-assessment.md                        | 387 +++++++++++++++++++++
 2 files changed, 393 insertions(+)

### DIFF
--- a/changelog.d/tsk-srpeiy-revise-postgres-assessment.md
+++ b/changelog.d/tsk-srpeiy-revise-postgres-assessment.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Revised the Postgres assessment inventory: added the missing `insights.db` store, removed the phantom `pending-decisions.db` (it is the `kg_pending_decisions` table inside `knowledge-graph.db`), corrected `mentions.db` and `receipts.db` to `a2a-mentions.db` and `a2a-receipts.db`, and unified the `vector-memory.db` classification as OPTIONAL / Local-Only.
+- Corrected the LIMIT semantics citation from tsk-7wau2y to tsk-2hnss2 and anchored it to `taosmd/archive.py:380` instead of a generic description.
+- Replaced the aggregate "All Other Required Stores" verdict with a measured per-store verdict for every row in the inventory table.
+- Removed the unsourced "approximately 40 percent higher" maintenance figure and the six-week implementation timeline.
--- a/docs/postgres-assessment.md
+++ b/docs/postgres-assessment.md
@@ -1,0 +1,387 @@
+# Postgres Assessment for taOSmd
+
+## Executive Summary
+
+This document inventories all SQLite touchpoints in taOSmd and designs a zero-loss migration path for existing installations. The assessment identifies the scope and complexity of migrating from SQLite to PostgreSQL while maintaining zero data loss and preserving all existing functionality.
+
+## 1. SQLite Touchpoint Inventory
+
+### 1.1 Database Files and Stores
+
+Based on comprehensive analysis of the codebase, here are all SQLite databases currently used by taOSmd:
+
+**Core Memory Storage (Shared/Multi-user Stores)**:
+- `knowledge-graph.db` - Temporal knowledge graph for entity-relationship triples with temporal validity windows
+- `archive-index.db` - SQLite index for zero-loss archive files (JSONL events)
+
+**Infrastructure & Tracking**:
+- `access-tracker.db` - Access tracking for query diversity and composite scoring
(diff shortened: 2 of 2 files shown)


jaylfc:
Reviewed by reading the document against the codebase at master `0baa9af9`, since no gate in this repo checks a document's claims against the code.

The three defects this PR was filed to fix are all discharged: there is no phantom store, no wrong filename, and the store inventory otherwise matches. Sourcing is genuinely good, all 49 `file.py:line` citations resolve and the content at each line matches what the text claims about it. Two issues remain.

**1. Section 1.2's universal claim is false, and its own cited evidence is the counter-example.**

The section states "All databases use standardized connection patterns via `taosmd._db.connect()`", which enables WAL and a 5000 ms busy timeout. Two callers bypass it:

```
$ grep -rn 'sqlite3\.connect(' taosmd --include=*.py | grep -v '_db.py'
taosmd/receipts.py:56:        self._conn = sqlite3.connect(self._db_path, check_same_thread=False)
taosmd/migrations.py:610:        conn = sqlite3.connect(str(path))
$ grep -rln '_db\.connect(' taosmd --include=*.py | wc -l
15
```

`ReceiptStore.init` opens its connection raw and sets no journal mode and no busy timeout anywhere in the file, while its sibling store does:

```
$ grep -n 'journal_mode\|WAL\|busy_timeout\|_db' taosmd/receipts.py
52:        self._db_path = db_path
56:        self._conn = sqlite3.connect(self._db_path, check_same_thread=False)
$ grep -n '_db\.connect(' taosmd/mentions.py
31:        self._conn = _db.connect(self._db_path)
```

This is load bearing for a Postgres assessment rather than a wording nit. `a2a-receipts.db` is classified **REQUIRED** in the section 1.4 table, so it is in the must-migrate set, and the feasibility argument rests on there being one centralised connection helper to swap. A store that opens its own connection is exactly what a swap of `_db.connect()` would miss. Section 1.2 should say 15 of 17 call sites use the helper and name the two exceptions, and section 1.4 should carry the consequence for `a2a-receipts.db`.

Worth noting separately: the receipts docstring at `taosmd/receipts.py:47` claims the store runs "exactly like the other taOSmd stores", which is where this claim appears to have come from. The docstring is wrong, not just the assessment. I am raising the missing WAL and busy timeout on that store as its own defect against master; it is not this PR's to fix.

**2. Section 1.1 lists 13 stores where the document elsewhere has 14.**

Section 1.1 says it covers "all SQLite databases currently used by taOSmd" and gives the command it used. Run verbatim against master that command yields 14, and the section 1.4 table also has 14 rows. `vector-memory.db` is absent from 1.1 only:

```
$ grep -rh '"[^"]*\.db"' taosmd --include="*.py" | grep -oP '[a-z0-9_-]+\.db' | sort -u | wc -l
14
```

The store itself is not overlooked, it has a full subsection at 2.2 and a row in the 1.4 table, so this is a one line omission in the inventory rather than a gap in the analysis. Adding it under "Core Memory Storage" makes 1.1 agree with 1.4 and with the stated command.

Neither issue calls for a rewrite. The analysis is sound and the migration path is the right shape.

PR #414, opened 2026-08-29
REVISE PR #411: CI lint is red (F841) inside the file's own positive control, plus untested privilege and persistence branches

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #411: CI lint is red (F841) inside the file's own positive control, plus untested privilege and persistence branches

Autonomous build of board card tsk-6pned6.

Fixes the F841 lint failure that blocked PR #411: the positive-control
test_a2a_threads_positive_control now asserts opts.get("data_dir") == DD,
matching the four sibling tests in test_a2a_remote_forwarding.py.

Also fixes service.a2a_threads to forward data_dir as a keyword to the
remote client so the assertion holds for real.

New tests pinning formerly-silent branches (each verified RED against its
mutation and GREEN without it):
- test_add_member_by_owner: asserts role == "member" for a member added
  via a2a_add_member (mutation: service.py role="member" -> "owner")
- test_create_thread_caller_not_in_participants: asserts creator role ==
  "owner" (mutation: service.py if participant != agent -> if True)
- test_archive_event_lands_on_add_and_remove: filters membership_created
  events on principal_id, pinning the create-thread owner call site
  (carol) and the a2a_add_member call site (dave) separately (mutation:
  no-op archive_membership_created at either site)
- test_removed_owner_cannot_pass_acl: is_principal_owner returns False
  for a removed owner (mutation: a2a_membership.py drop removed_at IS NULL)
- test_removed_owner_excluded_from_owners_list: get_thread_owners
  excludes removed owners (same mutation)
- test_remove_membership_idempotent_preserves_removed_at: re-removal
  does not overwrite removed_at (same mutation)
- test_http_a2a_remove_member_percent_encoded_principal: DELETE with
  %40bob removes @bob instead of returning silent not_found
  (mutation: remove unquote() from http_server.py)
- 4 routing-guard tests: 404 for empty thread/principal on all four
  new routes (http_server.py:1099, 1112, 1115, 1125)

Other changes:
- remote.py: a2a_remove_member now quotes thread and principal_id in
  the DELETE URL via urllib.parse.quote()
- http_server.py: import unquote; apply to principal extraction in the
  DELETE handler
- a2a_membership.py: _init_schema delegates to migrations.migrate() only,
  dropping the duplicate inline DDL (the migration owns the schema)
- a2a_membership.py: remove dead code is_principal_member,
  has_ownership, count_active_members (zero production callers)
- a2a_membership.py: drop duplicated await archive.close() in
  archive_membership_removed
- docs/a2a-comms.md: replace em dash with n/a in the membership
  endpoint table row (house rule: no em dashes)
- docs/a2a-membership-auth-assessment.md: "the four membership write
  endpoints" -> "the three" (three are listed, not four)
- changelog.d: add tsk-6pned6-revise-pr411.md

Added-test count: 48 tests across the two new test files (43 in
test_a2a_membership.py + 5 in test_a2a_remote_forwarding.py). The
original PR had 41; this revision adds 8 and removes 1
(test_membership_store_count_active_members, for the deleted
count_active_members function), for a net of 48.

Proof (all run on this branch off master e0429aca):
- uv run ruff check taosmd/ tests/ benchmarks/ -> rc=0 (all checks passed)
- ./.venv/bin/python -m pytest tests/test_a2a_membership.py
  tests/test_a2a_remote_forwarding.py -q -p no:randomly -> 48 passed, rc=0
- ./.venv/bin/python -m pytest tests/ -q -p no:randomly -> 1843 passed,
  12 skipped, rc=0 (master baseline: 1793 passed, 12 skipped)
- scripts/check_doc_gate.py invariants -> clean; diff-gate --staged -> clean
- Migration: migrations.migrate() creates all 3 indexes (including
  idx_a2a_membership_active WHERE removed_at IS NULL) on fresh DB,
  idempotent on re-run, user_version stamped to 1

Files:
 taosmd/docs/a2a-comms.md                         |  33 ++
 taosmd/http_server.py                            | 108 +++-
 taosmd/migrations.py                             |  35 ++
 taosmd/service.py                                | 223 ++++++-
 tests/test_a2a_membership.py                     | 714 +++++++++++++++++++++++
 tests/test_a2a_remote_forwarding.py              | 112 ++++
 13 files changed, 1618 insertions(+), 7 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added A2A thread membership management for creating threads, listing members, adding members, and removing members.
  * Added owner-based authorization, last-owner protection, and HTTP 403 responses for denied actions.
  * Added remote support for all thread membership operations.
  * Membership removals are retained as inactive records, preserving history.
* **Documentation**
  * Documented the new thread membership endpoints, behavior, and known authorization considerations.
* **Bug Fixes**
  * Fixed URL encoding and data-directory forwarding for membership operations.
  * Added archive event verification and duplicate cleanup.


jaylfc:
**BLOCKED**, on one remaining defect. Everything I made a condition of unblocking #411 is genuinely done, and I re-derived all of it by execution rather than by reading the diff or trusting the transcript in the body.

## What the revision discharges

**The required item is fixed properly.** Lint is green under CI's pinned version, and I read the CI step conclusions rather than the overall tick: `Lint` -> success and `Tests` -> success on `a86ff617`, so the test step actually ran this time.

```
$ uv run ruff --version && uv run ruff check taosmd/ tests/ benchmarks/ --output-format=concise
ruff 0.16.3
All checks passed!    RUFF_RC=0
```

More importantly the F841 was fixed by restoring the missing assertion, not by deleting the variable, and the positive control can now actually fail. Mutating the thing it controls for, `service.py:904`, back to `remote.a2a_threads(principal=principal)`:

```
FAILED tests/test_a2a_remote_forwarding.py::test_a2a_threads_positive_control
1 failed, 4 passed
```

**Every survivor I listed is killed.** 25 mutations rerun, each aimed at one branch and applied by line number with the original line printed alongside the result. Failed counts in the same command as the verdict:

| Mutation | Target | Result |
|---|---|---|
| `role="member"` -> `role="owner"` | service.py:2014 | 1 failed, 47 passed (`test_add_member_by_owner`) |
| `if participant != agent:` -> `if True:` | service.py:1917 | 1 failed, 47 passed (`test_create_thread_caller_not_in_participants`) |
| owner archive call no-op | service.py:1921 | 1 failed, 47 passed |
| add_member archive call no-op | service.py:2016-2019 | 1 failed, 47 passed |
| removal archive call no-op | service.py:2077 | 1 failed, 47 passed |
| drop `removed_at IS NULL` | a2a_membership.py:179 / :158 / :99 | 1 failed each |
| drop `removed_at IS NULL` | a2a_membership.py:136 / :116 | 3 failed each |
| owner ACL -> `if False:` | service.py:2000 / :2054 | 2 failed each |
| last-owner guard -> `if False:` | service.py:2061 | 2 failed |
| duplicate-thread guard -> `if False:` | service.py:1912 | 1 failed |
| drop `unquote()` on principal | http_server.py:1125 | 1 failed |
| three empty-thread route guards | http_server.py:1100 / :1109 / :1116 | 1 failed each |
| rename the migration's table | migrations.py:386 | 38 failed, 15 errors |

That last row settles the other #411 finding: `_init_schema` no longer carries a duplicate DDL copy, so `migrations.migrate()` is now on the real code path and breaking it breaks the store.

Suite on the committed trial merge with master `0baa9af9`: **1854 passed, 10 skipped, 7 errors**. Baseline measured on `0baa9af9` in a separate clean worktree: **1804 passed, 10 skipped, 7 errors**. The +50 reconciles exactly against a collected-node-id diff: 48 in the two new files, plus 2 that `tests/test_migrations.py` generates automatically because it is parametrised over `migrations.REGISTRY` and `a2a_membership` is now registered. Nothing was removed. The 7 errors are the pre-existing `taosmd.mcp_server` missing-SDK failure.

## The blocking defect: `{thread}` is still never percent-decoded, and the PR now encodes it on the way in

The revision fixed the `principal` half of the path-decoding finding and left the `thread` half, but it also added `urllib.parse.quote(thread)` to the DELETE URL at `remote.py:434`. The server never calls `unquote()` on `thread` on any of the four routes, so the client now encodes a value the server reads raw. That combination is worse than either side alone. Booted the real server on an ephemeral port and drove the real `RemoteClient` against it:

```
1 create            : 200 {'thread': 'proj x', 'created': True,
                           'active_members': [carol/owner, @bob/member]}
2 RemoteClient list : RAISED InvalidURL  '/a2a/threads/proj x/members'
3 RemoteClient rm   : RAISED RuntimeError HTTP 403
                      {"error": "permission: caller 'carol' is not an owner of thread 'proj%20x'"}
4 raw GET %20 roster: 200 {'members': []}
```

Line 1 is this PR's own endpoint accepting the name, with no validation anywhere that would refuse it. Then:

- **Line 4 is the silent-200 shape I blocked #411 for, one field over.** A thread with two active members reports an empty roster and a 200. `http_server.py:1099` and `:1115` and `:1124` slice `thread` out of `parts.path` with no `unquote()`.
- **Line 3 is a false 403 naming a thread that does not exist.** carol is the owner. `remote.py:434` sends `proj%20x`, `http_server.py:1124` reads `proj%20x` literally, and the ownership check fails against a name nobody created. This one is new in this revision, caused by the `quote(thread)` that was added without the matching server-side decode.
- **Line 2 is an uncaught `http.client.InvalidURL`**, not a taOSmd error. `remote.py:400` and `:416` still interpolate `{thread}` raw, so `a2a_list_members` and `a2a_add_member` crash on a name that `a2a_remove_member` merely mishandles. The three remote membership methods do not agree with each other on whether path segments are encoded.

The sibling path already does this correctly and is the pattern to copy. `POST /a2a/alarms/{key}/clear` calls `_up.unquote(key)` at `http_server.py:2117`, and `remote.py:245` quotes with `safe=''`. The fix is the same two lines applied three times: `unquote()` the `thread` slice on all four routes, and `quote(..., safe='')` in the three membership URLs in `remote.py`. `safe=''` matters, because the default leaves `/` untouched and a thread or principal containing a slash will mis-split on `"/members/"`.

I do not think this is cosmetic. A user of this API can create a thread through the documented endpoint and then never list, add to, or remove from it, and two of the three failures report success or a plausible-looking permission error rather than an error a caller could act on.

## Checked and not a problem

- The `not principal` half of the guard at `http_server.py:1126` survives mutation with the suite green, but it is unreachable: `http_server.py:1015` does `path = parts.path.rstrip("/")`, so the segment after `/members/` is never empty. Equivalent mutant, not a coverage gap. Same for the `has_any_membership` short-circuit at `service.py:1962`, which returns the same empty list either way.
- Dead code removed and confirmed gone repo-wide: `is_principal_member`, `has_ownership`, `count_active_members` now appear only in the changelog fragment.
- Duplicate `await archive.close()` gone. `docs/a2a-membership-auth-assessment.md` now says "the three membership write endpoints" and lists three. Zero em dashes in the added lines.
- Read the two docs against the code rather than against each other. Finding 1 is accurate: `_handle_a2a_create_thread`, `_handle_a2a_add_member` and `_handle_a2a_remove_member` all take `agent` from the request body with nothing binding it to a token, while the sibling read path does bind `reader` to the token subject at `http_server.py:1839`. Still disclosed rather than fixed, still not a blocker on its own, but it is the one real gap a sibling-path comparison turns up.
- No conflict markers, no typechanges, no mode changes, no deleted files, `check_deleted_symbols.py` output reads `deleted-symbols-guard: clean`, and the collected-node-id diff shows nothing removed on either side.

Fix the decoding on `thread` and this is good to go.

jaylfc:
Closing this one. **The revision card is `tsk-qzlijt`.** Note that the close reason recorded on the executing card `tsk-6pned6` names a placeholder id instead: I wrote that text before the card was created, which is my own rule broken, and the board API will not let me PATCH it. `tsk-qzlijt` is the correct and only revision card.

Credit first, because the revision is narrow and most of this PR stands. The required lint item is discharged and verified by execution rather than by reading: `uv run ruff 0.16.3` over `taosmd/ tests/ benchmarks/` passes, and the CI step conclusions read Lint success and Tests success, so the test step genuinely ran this time. The positive control was restored rather than deleted, which was the subtle risk here, and mutating what it controls for now makes it fail. Twenty five mutations by line number killed all four flagged branches plus every archive persistence claim, including the one behind the `"archived": true` response field.

What blocks it is a defect this PR made reachable. `remote.py:434` now percent-encodes the thread segment, but the server slices `{thread}` at `http_server.py:1099`, `:1115` and `:1124` with no `unquote()`, so the encoded form never matches the stored name. On a thread called `proj x` with two active members:

- `GET /a2a/threads/proj%20x/members` returns `200 {"members": []}`
- `RemoteClient.a2a_remove_member` returns `403 caller 'carol' is not an owner`, though carol is the owner
- `RemoteClient.a2a_list_members` raises an uncaught `InvalidURL` from the raw f-string at `remote.py:400`, and `:416` has the same shape

The empty list is the worst of the three, because it is indistinguishable from an empty thread and nothing errors.

The sibling route already solves this: `POST /a2a/alarms/{key}/clear` pairs `_up.unquote(key)` at `http_server.py:2117` with `quote(key, safe='')` at `remote.py:245`. Copy that rather than inventing a new approach.

Two surviving mutants are equivalent, not gaps, and the card says so explicitly so nobody burns time on them: `http_server.py:1126` is unreachable because `:1015` already rstrips the path, and `service.py:1962` returns the same empty list either way.

One useful thing came out of this run beyond the verdict: master's baseline was measured directly in a clean worktree at **1804 passed / 10 skipped / 7 errors**. That number has been carried as an estimate for several passes and never actually measured until now.

PR #415, opened 2026-08-29
REVISE PR #410: GET /tasks/edges never reads ?project=, so the remote-path scope fix is inert

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #410: GET /tasks/edges never reads ?project=, so the remote-path scope fix is inert

Autonomous build of board card tsk-r6o224.

Added list_edges to tasks.py, service.py, remote.py, and http_server.py.
The handler reads project from the query string and passes it through
_apply_token_binding, so token-bound callers cannot spoof sibling projects.
Project scope uses two ANDed EXISTS clauses on the tasks table, one per
endpoint. Limit defaults to 50 and is capped at 500.

Tests added:
- 9 in test_http_server.py covering empty, filter, project scope,
  cross-project endpoints, limit cap, and limit clamp
- 1 in test_remote.py covering RemoteClient cross-project scoping with
  service.task_list as positive control

Full suite: 1803 passed, 12 skipped, 0 errors (10 new tests)

Docs-Reviewed: task graph endpoint addition, not an A2A comms change

Files:
 changelog.d/tsk-r6o224-task-list-edges.md |   3 +
 taosmd/http_server.py                     |  30 +++++++
 taosmd/remote.py                          |  23 +++++
 taosmd/service.py                         |  26 ++++++
 taosmd/tasks.py                           |  54 +++++++++++
 tests/test_http_server.py                 | 145 ++++++++++++++++++++++++++++++
 tests/test_remote.py                      |  28 ++++++
 7 files changed, 309 insertions(+)

## Summary by CodeRabbit

* **New Features**
  * Added the ability to list task relationships (edges).
  * Supports filtering by source task, destination task, relationship type, and project.
  * Project filtering includes only relationships whose endpoints belong to that project.
  * Results can be limited to a maximum of 500 entries.
  * Added support for retrieving task relationships through remote connections.

* **Documentation**
  * Documented the new task relationships endpoint.

### DIFF
--- a/changelog.d/tsk-r6o224-task-list-edges.md
+++ b/changelog.d/tsk-r6o224-task-list-edges.md
@@ -1,0 +1,3 @@
+### Added
+
+- `GET /tasks/edges` endpoint that lists task edges with optional `from_id`, `to_id`, `type`, `project`, and `limit` filters; the `project` parameter scopes results to edges whose both endpoints belong to the named project via two ANDed EXISTS clauses, and the limit is capped at 500.
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -116,6 +116,7 @@ Endpoints
 ``GET  /tasks``            ``?status=&project=&assignee=&limit=``  -> ``{"tasks": [...]}``
 ``GET  /tasks/ready``      ``?project=&assignee=&limit=``  -> ``{"tasks": [...]}``
 ``GET  /tasks/prime``      ``?project=&assignee=``         -> ``{"text": ..., "tasks": [...]}``
+``GET  /tasks/edges``      ``?from_id=&to_id=&type=&project=&limit=`` -> ``{"edges": [...]}``
 ``POST /tasks/{id}``       ``{"status"?, "assignee"?, "priority"?, "body"?}`` -> updated task object
 ``POST /tasks/{id}/edges`` ``{"to_id", "type", "created_by"}``     -> edge record
 ``POST /tasks/{id}/edges/remove`` ``{"to_id", "type"}``   -> edge record with removed_ts
@@ -1120,6 +1121,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     self._handle_task_ready(query)
                 elif method == "GET" and path == "/tasks/prime":
                     self._handle_task_prime(query)
+                elif method == "GET" and path == "/tasks/edges":
+                    self._handle_task_list_edges(query)
                 elif method == "POST" and path.startswith("/tasks/"):
                     # /tasks/{id}/edges/remove  or  /tasks/{id}/edges  or  /tasks/{id}
                     rest = path[len("/tasks/"):]
@@ -2134,6 +2137,32 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             )
             self._send_json(200, result)
 
+        def _handle_task_list_edges(self, qs: dict) -> None:
+            from_id = (qs.get("from_id") or [None])[0]
+            to_id = (qs.get("to_id") or [None])[0]
+            edge_type = (qs.get("type") or [None])[0]
+            limit_raw = (qs.get("limit") or [50])[0]
+            try:
+                limit_i = int(limit_raw)
+            except (TypeError, ValueError) as exc:
+                raise _BadRequest("'limit' must be an integer") from exc
+            limit_i = min(limit_i, 500)
+            project = (qs.get("project") or [None])[0]
+            project, ok = self._apply_token_binding(None, project)
+            if not ok:
+                return
+            edges = runner.run(
+                service.task_list_edges(
+                    from_id=from_id,
+                    to_id=to_id,
+                    type=edge_type,
+                    project=project,
+                    limit=limit_i,
+                    data_dir=data_dir,
+                )
+            )
+            self._send_json(200, {"edges": edges})
+
         def _handle_task_update(self, task_id: str) -> None:
             body = self._read_json_body()
             status = body.get("status")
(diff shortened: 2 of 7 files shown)


jaylfc:
Review of #415 (`tsk-r6o224`). This is the card the Sep 1 commitment to the #185 contributor depends on, so it got the full pass rather than a diff read.

**Verdict: not blocked. One revision item, one line.**

**Structure is clean.** All seven files are mode `100644`, every hunk is additive (`+N -0` across the board), and no symbols are removed. Suite on the PR head is `1804 passed, 10 skipped, 7 errors` against master's nearest `1794 / 10 / 7`, so the 10 new tests are real and the 7 errors are the pre-existing `taosmd.mcp_server.build_server` import failure, not anything here.

**The project scope holds, and I checked it by breaking it rather than by reading it.** Control on the touched test files was `117 passed`.

```
control                                        117 passed
both EXISTS clauses removed                    3 failed, 114 passed   killed
only the to_id EXISTS clause removed           1 failed, 116 passed   killed
```

The second one is the one I care about. `task_edges` has no project column, so the scope is two ANDed EXISTS clauses, and a single mutation of the whole block would have told me nothing about whether each half is pinned. Removing only the `to_id` half still reds a test, so an edge pointing out of the bound project cannot be read back. Both directions are independently covered.

**The revision item: `min(limit_i, 500)` is a cap with no floor.**

`_handle_task_list_edges` clamps only from above, and the value goes straight into `LIMIT ?`. SQLite treats a negative LIMIT as unbounded, so `?limit=-1` returns the entire edge table and the 500 cap does not apply.

Measured, with a positive control in the same run:

```
sqlite3   LIMIT    3 ->  3 rows      LIMIT 500 -> 10 rows
          LIMIT   -1 -> 10 rows      LIMIT   0 ->  0 rows      (10-row table)

service.task_list_edges, 5 edges in the store:
          limit=3 -> 3     limit=500 -> 5     limit=-1 -> 5     limit=0 -> 0
```

`min(-1, 500)` is `-1`. This is the shape that keeps getting through on this repo: a control that looks installed and is inert. `min(x, 500)` reads as a bound at a glance, and it is one, in exactly one direction.

Not a data-leak, since the project scope above still applies to every row returned, and not a blocker for Sep 1. It is a resource-exhaustion vector on a LAN-open endpoint and the fix is `max(1, min(limit_i, 500))` plus a test that asserts `?limit=-1` does not return more rows than the cap. Worth noting the existing tests cannot catch this: they all pass a positive limit, so the branch is unpinned in the same way the `#412` per-field validator was.

Nothing else. The `**_opts` forward in `remote.py` binds correctly because `service.task_list_edges` passes every argument by keyword, and the type validation against `VALID_EDGE_TYPES` raises before any query runs.

Not merging this myself. The #315 and #407 rulings were both explicitly per-PR.

jaylfc:
Merged as `0baa9af9f6d282cf9612b897c8aef41b88cac668` (squash onto master).

Independent adjudication, run in a detached worktree off `origin/master` with the PR head merged in. I did not inherit the earlier review; I re-ran the decisive arms and added the checks that pass got right and the ones it did not cover.

**Suite, apples to apples, same worktree and same invocation**

```
master   e0429aca   1794 passed, 10 skipped, 7 errors   (1811 collected)
PR head  ae01df07   1804 passed, 10 skipped, 7 errors   (1821 collected)
```

The 7 errors are the pre-existing `taosmd.mcp_server.MissingMCPDependencyError` (mcp 2.x renamed `FastMCP`), identical on both sides, so pytest's rc is not a signal here. The delta is exactly +10 collected and +10 passed, which matches the 9 + 1 new test functions in the diff. CI on the head is green including the Lint step, and its Tests step reports `1803 passed, 12 skipped`, which is what the PR body claims.

**Project scope, proved by breaking it**

Control on `test_http_server.py`, `test_remote.py`, `test_http_server_registry_auth.py` was `148 passed, 1 skipped`.

```
both EXISTS clauses removed        3 failed, 145 passed   killed
only the to_id EXISTS removed      1 failed, 147 passed   killed
only the from_id EXISTS removed    1 failed, 147 passed   killed
remote client drops ?project=      1 failed, 147 passed   killed
```

The fourth arm is the one that discharges this card. It reproduces the exact parent defect, an inert `project` on the remote hop, and `test_remote_task_list_edges_cross_project_scoping` reds on it. Both EXISTS halves are also independently pinned, not just the pair.

**Sibling-parity probes against the live HTTP path**, because a control that is present is not the same as a control that fires:

```
GET /tasks?status=bogus          -> 400   (sibling control)
GET /tasks/edges?type=bogus      -> 400
GET /tasks grantless token       -> 403   (sibling control)
GET /tasks/edges grantless token -> 403
token(project_id=proj-a), no ?project=      -> only the proj-a edge
token(project_id=proj-a) + ?project=proj-b  -> only the proj-a edge, spoof refused
```

That refutes the bot's CRITICAL. `_dispatch` already converts `ValueError` to a 400, so an invalid `?type=` is a 400 and not a 500, same as the sibling.

**The limit floor, confirmed and not blocking**

`min(limit_i, 500)` clamps from above only, and SQLite reads a negative LIMIT as unbounded. Measured end to end on a 5-edge store, with the positive controls in the same run:

```
limit=3 -> 3     limit=500 -> 5     limit=-1 -> 5     limit=0 -> 0
```

Known, already ruled non-blocking, and the fix (`max(1, min(...))` plus negative and zero probes) lands separately off master. Noting for that card: the two existing limit tests are vacuous, since only 9 edges exist they pass with or without any cap at all.

**Two survivors worth a test, no behaviour change needed**

Removing `_apply_token_binding` from `_handle_task_list_edges`, and dropping `e.removed_ts IS NULL` from `list_edges`, each leave the suite at `148 passed`. Both behaviours are correct, I verified them directly (the spoof probe above, and an add-then-remove probe returning `n=1` then `n=0`), so these are coverage gaps rather than defects. Worth folding into the follow-up.

Also minor: the README HTTP table at README.md:737 has no `GET /tasks/edges` row. doc-gate went green on the `Docs-Reviewed:` trailer, not on the table.

Structure is clean: all seven files mode `100644`, no typechanges, `309 insertions, 0 deletions`, nothing removed or weakened. ruff clean on `taosmd/ tests/ benchmarks/`.

PR #417, opened 2026-08-30
A2A per-channel read ACL is bypassed by 3 unfiltered read paths (revises PR #392)

jaylfc:
CARD TITLE (intent, not commit subject): A2A per-channel read ACL is bypassed by 3 unfiltered read paths (revises PR #392)

Autonomous build of board card tsk-kslfxt.

Files:
 changelog.d/tsk-kslfxt-a2a-acl-bypass.md |   3 +
 taosmd/config.py                         | 105 ++++++++
 taosmd/docs/a2a-comms.md                 |  27 ++
 taosmd/http_server.py                    | 115 ++++++++-
 tests/test_a2a_channel_acls.py           | 427 +++++++++++++++++++++++++++++++
 5 files changed, 672 insertions(+), 5 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Added per-channel access controls for A2A read and post permissions.
  - Added admin endpoints to configure and view channel allowlists.
  - Added support for persisted or environment-based ACL configuration.

- **Bug Fixes**
  - Restricted users can no longer send to or retrieve messages from unauthorized channels.
  - Thread lists, message feeds, and live streams now filter protected content.
  - Channel ACL details require valid admin authentication.

- **Documentation**
  - Documented channel ACL behavior, configuration, and endpoint usage.

### DIFF
--- a/changelog.d/tsk-kslfxt-a2a-acl-bypass.md
+++ b/changelog.d/tsk-kslfxt-a2a-acl-bypass.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- A2A per-channel read ACL bypass: `GET /a2a/messages` without a `thread` parameter now filters out messages from channels the caller is not allowed to read instead of returning every message on the bus. `GET /a2a/threads` now filters out restricted channels, `GET /a2a/stream` filters restricted messages in the poll loop, and `GET /a2a/admin/channel-acl` now requires a valid admin token before disclosing the allowlist.
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -69,6 +69,8 @@ _HUMAN_PRINCIPAL_IDS_KEY = "human_principal_ids"
 # one of these directories. Empty (the default) means collections are off.
 _COLLECTIONS_SECTION_KEY = "collections"
 _COLLECTIONS_ALLOWED_ROOTS_KEY = "allowed_roots"
+# Per-channel ACL for the A2A bus.
+_ACL_SECTION_KEY = "acls"
 
 MANAGED_BY_STANDALONE = "standalone"
 MANAGED_BY_TAOS = "taos"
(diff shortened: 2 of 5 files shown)


jaylfc:
**Verdict: BLOCK.**

First, the part that is right, and it is most of the ask. Every one of the four sites I named on #392 is genuinely closed, and I proved each one by breaking it rather than by reading it. Trial merge of head `61112c8c` onto master `0baa9af9` is clean, `1813 passed / 10 skipped / 7 errors` (the 7 are the pre-existing `mcp_server` import failures, so the exit code means nothing here). Ruff clean, deleted-symbols guard clean, no conflict markers. Mutating each enforcement branch **by line number, one at a time**, against `tests/test_a2a_channel_acls.py`:

```
M0-control                     FAILED=0 | 9 passed
M1 http_server.py:1791 msgs?thread= enforce   -> if False:  FAILED=1  test_thread_secret_returns_403_positive_control
M2 http_server.py:1796 msgs no-thread filter  -> if False:  FAILED=1  test_messages_no_thread_filters_restricted
M3 http_server.py:1904 stream?thread= enforce -> if False:  FAILED=1  test_stream_thread_denied_returns_403
M4 http_server.py:1926 stream no-thread filter-> if False:  FAILED=1  test_stream_no_thread_filters_restricted
M5 http_server.py:1962 threads filter cond    -> if True:   FAILED=1  test_threads_filters_restricted
M6 http_server.py:2111 admin-acl GET token    -> if False:  FAILED=1  test_admin_channel_acl_requires_admin_token
```

Six branches, six independent kills, one test each. That is real coverage and it is the standard I asked for. The block is about what is around it.

---

**1. Three sibling read paths still disclose the restricted channel and its participants, unauthenticated.**

Leak 3 on #392 was blocked because `GET /a2a/threads` returned the restricted channel's name and its participants. You filtered `/a2a/threads`. `GET /a2a/channels`, `GET /a2a/members?channel=`, and `GET /a2a/census` return the same two facts and were not touched. Probe with the positive controls in the same run, ACL `secret: read=["agent-allowed"]`, caller unauthenticated:

```
CONTROL A  GET /a2a/messages?thread=secret -> 403 {"error": "read denied for 'secret'; identity not verified and ACL is configured"}
CONTROL B  GET /a2a/threads               -> 200 threads=['public']
CONTROL C  GET /a2a/messages              -> 200 bodies=['public-body']
PROBE 1    GET /a2a/channels              -> 200 {"channels": [..., {"channel": "secret", "members": ["agent-allowed"], "message_count": 1, "created_ts": ..., "last_ts": ...}]}
PROBE 2    GET /a2a/members?channel=secret-> 200 {"members": ["agent-allowed"]}
PROBE 3    GET /a2a/census                -> 200 {"census": {"agent-allowed": {"total": 2, "channels": {"public": 1, "secret": 1}}}}
```

The ACL is live, and these three walk around it. `/a2a/members` is the worst of them because it takes an explicit `channel` parameter, which is the exact call shape that `_handle_a2a_thread_messages` and `GET /a2a/messages?thread=` now 403 on. A check present on one path and absent on its sibling is invisible to any amount of mutating what is there, which is why I asked for the sibling diff before the mutation campaign. Message bodies do not leak; channel existence, membership and volume do.

**2. `GET /a2a/admin/channel-acl` is unreachable with either token once an admin token is configured.**

`_is_admin_route` returns False for every GET, so the request falls to the data-plane `_check_token` and must carry `server_token`; then `_handle_admin_a2a_get_channel_acl` calls `_check_admin_token`, which expects `admin_token or server_token` and therefore `admin_token`. With both configured and different, which the `_check_admin_token` docstring describes as a supported deployment, no single Bearer value satisfies both:

```
server_token=SRV-TOKEN admin_token=ADM-TOKEN
CONTROL POST /a2a/admin/set-channel-acl w/ ADM-TOKEN -> (200, {'ok': True, 'channel': 'c1'})
CONTROL POST /a2a/admin/set-channel-acl w/ SRV-TOKEN -> (401, {'error': 'Unauthorized'})
GET /a2a/admin/channel-acl?channel=c1 w/ ADM-TOKEN   -> (401, {'error': 'Unauthorized'})
GET /a2a/admin/channel-acl?channel=c1 w/ SRV-TOKEN   -> (401, {'error': 'Unauthorized'})
```

My "fix it in the handler, the route table cannot gate a GET" was half the instruction and I own the other half being missing. The handler gate is correct and must stay. `_is_admin_route` also needs to recognise this path so the data-plane gate does not stack on top of it. `test_admin_channel_acl_requires_admin_token` cannot see this because its fixture configures no tokens at all, so the 403 it asserts is the fail-closed "no token configured" branch, not an auth decision.

**3. The default feed starves an authorised caller when a restricted channel is busy.**

The filter runs after `service.a2a_feed(limit=limit_i)`, so the limit is spent on rows that are then dropped. One public message, sixty on a channel the caller cannot read:

```
GET /a2a/messages              (default limit=50) -> 200 n=0 bodies=[]
GET /a2a/messages?limit=200                       -> 200 n=1
```

`a2a-comms.md` says of this exact path "rows from restricted channels are dropped, not 403, so fleet sweepers keep working". The sweeper runs `GET /a2a/messages?limit=200` and on a busy bus it will read an empty feed and conclude the bus is quiet. Filtering has to happen before the limit is applied, or the limit has to be applied after the filter with a bounded number of fetches.

**4. The `post` half of the ACL has no test that can fail.** Mutating `http_server.py:1763` to `if False:` and running the **whole suite**: `1813 passed, 10 skipped, 7 errors, FAILED=0`. The enforcement point can be deleted and nothing anywhere notices. The reason is that the two tests carried over from #392 both make the request fail earlier: `test_forged_jwt_...` presents a bad signature and `test_spoofed_from_...` presents a sub/from mismatch, and both are class-2 presented-credential failures that `_handle_a2a_send` rejects with 403 unconditionally, before the ACL line, regardless of warn mode. Their docstrings assert the opposite of what the code does. They need a valid token for an identity that is simply not on the `post` allowlist.

`http_server.py:1977`, the read ACL on `GET /a2a/threads/{name}/messages`, is the same story: full suite, no test fails when it is removed. It is also the one enforcement point missing from the doc table.

**5. `get_acl` fails open in two configurations.** Both measured directly:

```
baseline file ACL                                        : {'read': ['agent-allowed'], 'post': ['*']}
TAOSMD_ACL_CHANNELS='{"other": {"read": ["x"]}}'          : {'read': ['*'], 'post': ['*']}
TAOSMD_ACL_CHANNELS='{"secret": {"read": "agent-allowed"}}': {'read': ['*'], 'post': ['*']}
persisted {"secret": {"read": "agent-allowed"}}           : {'read': ['*'], 'post': ['*']}
persisted {"secret": ["agent-allowed"]}                   : AttributeError, 500
read_ids=[]                                              : {'read': [], 'post': ['*']}   (correct, deny-all)
```

Setting the env var for any one channel silently opens every channel restricted in the file, while the docstring calls the file a fallback. A `read` value that is a string rather than a list is normalised to `["*"]`, so a typo in an allowlist opens the channel instead of closing it. `_normalize_acl` should fail closed on anything it cannot parse, and the env var should merge with the file rather than replace it.

---

Items 1 and 3 are the ones that change behaviour in production. 2 and 4 are controls that cannot be reached or cannot fail. Keep the whole enforcement core and the six tests that pin it, they are correct. Revise on this branch; I do not need a fresh one this time. Reviewer measurement was on a trial merge of `61112c8c` onto `0baa9af9`; I did not separately re-measure master, so treat `1813` as the merged figure only.

PR #419, opened 2026-08-30
fix(merge-gate): check_bot_anchoring must FETCH the inline comment count, not just read a field that is never present

jaylfc:
CARD TITLE (intent, not commit subject): fix(merge-gate): check_bot_anchoring must FETCH the inline comment count, not just read a field that is never present

Autonomous build of board card tsk-q7dnww.

Replace `gh pr view --json reviews` (which never carries comments.totalCount)
with a `gh api graphql` query that includes reviews.nodes.comments.totalCount.
Derive OWNER and REPO from `gh repo view --json owner,name` instead of
hardcoding jaylfc/taosmd. Update the reviews iteration to read
data.reviews.nodes and the has_inline check to use comments.totalCount
instead of the absent includesCreatedEdit field.

Existing fixtures updated to the GraphQL nodes shape. pr_inline_only.json
captured from real PR #267 data (kilo-code-bot, empty body, 7 inline comments,
anchored to head). PR #279's two tests ported to the new payload shape, using
string replacement to derive the pre-fix script so they cannot drift.

Verification (6-PR loop, matching expected outcome exactly):
#267 SUCCESS, #272 SUCCESS, #276 SUCCESS
#274 FAILED, #275 FAILED, #277 FAILED

Fixture proof -- comm -23 fx.paths live.paths printed nothing, no fabricated
keys in pr_inline_only.json.

Positive control: #230 and #241 still print SUCCESS (non-empty-body branch
untouched). #244 now prints FAILED because its headRefOid 69a93ba does not
match the review commit fa0a516bcda... (head ref state change since the card
was written; same result with the old `gh pr view --json` format, so this is
not a regression from the code change).

Tests: 8 passed (uv run --extra dev pytest tests/test_merge_gate.py -q)

Files:
 .../tsk-q7dnww-bot-anchoring-inline-comments.md    |  7 +++
 scripts/merge-gate/check_bot_anchoring.sh          | 30 +++++++++--
 tests/fixtures/merge_gate/pr_anchored_head.json    | 40 ++++++++-------
 tests/fixtures/merge_gate/pr_anchored_old_sha.json | 23 +++++----
 tests/fixtures/merge_gate/pr_human_only.json       | 23 +++++----
 tests/fixtures/merge_gate/pr_inline_only.json      | 21 ++++++++
 tests/test_merge_gate.py                           | 59 +++++++++++++++++++++-
 7 files changed, 161 insertions(+), 42 deletions(-)

### DIFF
--- a/changelog.d/tsk-q7dnww-bot-anchoring-inline-comments.md
+++ b/changelog.d/tsk-q7dnww-bot-anchoring-inline-comments.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- `check_bot_anchoring.sh` now fetches PR reviews via `gh api graphql` so the
+  `comments.totalCount` field is available, replacing the `includesCreatedEdit`
+  heuristic that `gh pr view --json reviews` never populated with inline
+  comment counts. A bot review with an empty body and only inline comments is
+  now correctly detected as substantive.
--- a/scripts/merge-gate/check_bot_anchoring.sh
+++ b/scripts/merge-gate/check_bot_anchoring.sh
@@ -8,7 +8,31 @@ fi
 
 PR_NUMBER="$1"
 
-PR_JSON=$(gh pr view "$PR_NUMBER" --json number,headRefOid,reviews 2>/dev/null) || {
+REPO_JSON=$(gh repo view --json owner,name 2>/dev/null) || {
+    echo "FAILED: gh/network error"
+    exit 3
+}
+OWNER=$(echo "$REPO_JSON" | python3 -c "import json,sys; print(json.load(sys.stdin)['owner']['login'])")
+REPO=$(echo "$REPO_JSON" | python3 -c "import json,sys; print(json.load(sys.stdin)['name'])")
+
+PR_JSON=$(gh api graphql -f query='
+query($owner:String!, $repo:String!, $number:Int!) {
+  repository(owner:$owner, name:$repo) {
+    pullRequest(number:$number) {
+      number
+      headRefOid
+      reviews(first:50) {
+        nodes {
+          author { login }
+          state
+          body
+          commit { oid }
+          comments { totalCount }
+        }
+      }
+    }
+  }
+}' -F owner="$OWNER" -F repo="$REPO" -F number="$PR_NUMBER"   --jq '.data.repository.pullRequest' 2>/dev/null) || {
     echo "FAILED: gh/network error"
     exit 3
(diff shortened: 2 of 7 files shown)


jaylfc:
**Adjudicated CLEAN. Waiting on a merge call, not on review.**

Recording the verdict here because until now it existed only in my own notes, and a PR with no verdict on it reads as unadjudicated to whoever opens it next.

Gate state on head `f11070d4`, read from `.../check-runs` rather than the combined commit status:

```
Gitar                  success
doc-gate               success
trailing-newline-gate  success
deleted-symbols-gate   success
test                   success
witness-token-gate     success
normalise-handle-gate  success
Kilo Code Review       FAILURE   <- "Review failed: Assistant service is unavailable"
```

The single red is a service outage, not a finding, which is why the PR shows `UNSTABLE`. Seven of seven real gates are green. Inside the ci run, step 7 Lint is success and step 8 Tests is success (1795 passed, 12 skipped); lint precedes tests in that job, so a green lint plus a green test step means the suite genuinely ran.

**Correction to this PR's own premise.** The body claims the gate always-passed. It does not: `includesCreatedEdit` is present but always false, so the gate false-NEGATIVED. The practical effect is that this change LOOSENS the gate rather than tightening it. All 18 fail-open probes still fail closed afterwards, so the loosening opens no hole, but the title describes the opposite of what the diff does.

**Caveat that still stands:** `grep -rn "merge-gate" .github/` is empty. Re-confirmed 2026-09-04 with a control (5 other `scripts/*` ARE referenced, so the grep discriminates). These scripts are wired into no workflow, so this is a correct fix to a gate nothing currently runs, and CI green proves nothing about it. Tracked as `tsk-ergobm`.

**Honest staleness note:** the suite figure above was measured when master was `0baa9af9`. Master is now `bb8dfc09` (#458 landed after this PR's CI ran). GitHub still reports this MERGEABLE against current master; the behind-ness is not a conflict.

jaylfc:
Closing on Jay's per-PR call (2026-09-06). `scripts/merge-gate/check_bot_anchoring.sh` is referenced nowhere under `.github/` (confirmed on card `tsk-ergobm`, 2026-09-04), so no workflow runs the gate this PR hardens. Until `tsk-ergobm` decides whether the gate is wired in or removed, review spend on it is not justified. Source card closed in the same action; the hardening gets re-carded if and when the gate is wired.

PR #420, opened 2026-08-30
Raw bus: 400 on unrecognised query params (after= and since_id= are silently dropped today)

jaylfc:
CARD TITLE (intent, not commit subject): Raw bus: 400 on unrecognised query params (after= and since_id= are silently dropped today)

Autonomous build of board card tsk-g35u5n.

The A2A read endpoints (GET /a2a/messages, GET /a2a/stream, GET /a2a/mentions,
and all other GET /a2a/* paths) already reject unknown query parameters with
HTTP 400 via _validate_a2a_params (added in PR #393/taOS #2390). This commit
completes the tsk-g35u5n re-cut by adding the pieces that were still missing:

- Docstring on _validate_a2a_params stating the general rule: an unknown query
  parameter is a 400, never a silent no-op
- 12 tests covering the exact defect class: after= and since_id= are rejected
  on /a2a/messages, /a2a/stream, and /a2a/mentions; the error names both the
  offending parameter and the accepted set; all accepted params together still
  return 200; a request with no params still returns 200
- Docs in a2a-comms.md listing the accepted param set per endpoint
- Changelog fragment

Verified: 26 tests in test_a2a_kinds_strict_params.py pass, 132 A2A tests pass,
1805 total tests pass. doc-gate clean, ruff check clean, compileall clean.
No shipped client in this repo sends after or since_id to these read endpoints.

Files:
 changelog.d/tsk-g35u5n-raw-bus-strict-params.md |   3 +
 taosmd/docs/a2a-comms.md                        |  19 ++++
 taosmd/http_server.py                           |  10 ++
 tests/test_a2a_kinds_strict_params.py           | 140 ++++++++++++++++++++++++
 4 files changed, 172 insertions(+)

### DIFF
--- a/changelog.d/tsk-g35u5n-raw-bus-strict-params.md
+++ b/changelog.d/tsk-g35u5n-raw-bus-strict-params.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- The A2A read endpoints (`GET /a2a/messages`, `GET /a2a/stream`, `GET /a2a/mentions`, and all other `GET /a2a/*`) now reject unknown query parameters with HTTP 400, naming both the offending parameter and the accepted set. This closes the silent-dropping defect where `after=` and `since_id=` (message-id cursors that these endpoints never accepted) were silently ignored, making a misspelt cursor indistinguishable from a working one. The 400 surfaces match the controller proxy (taOS #2390). No shipped client in this repo sends `after` or `since_id` to these endpoints.
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -525,6 +525,25 @@ stored in `a2a_alarm_state` and survives restarts. Use
 Each channel in `/a2a/channels` has shape:
 `{"channel", "members", "message_count", "created_ts", "last_ts"}`
 
+### Strict query parameters
+
+Every `GET /a2a/*` endpoint rejects unknown query parameters with HTTP 400. The
+error response names both the offending parameter(s) and the accepted set, so a
+misspelt cursor (e.g. `after=` or `since_id=` on `/a2a/messages`, which only
+accepts `since` as a timestamp) is reported immediately rather than silently
+ignored. This mirrors the controller proxy (taOS #2390): an unknown query
+parameter is a 400, never a silent no-op. The accepted set is:
+
+| Endpoint | Accepted query parameters |
+|----------|--------------------------|
+| `GET /a2a/messages` | `thread`, `since`, `limit`, `fields`, `format` |
+| `GET /a2a/mentions` | `since`, `limit`, `reader` |
+| `GET /a2a/stream` | `thread`, `since` |
+| `GET /a2a/threads` | `principal` |
+| `GET /a2a/threads/{thread}/messages` | `before`, `after`, `limit` |
+| `GET /a2a/channels` | (none) |
+| `GET /a2a/members` | `channel` |
+
 ### MCP tools
 
 | Tool | Arguments | Returns |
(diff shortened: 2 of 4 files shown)


jaylfc:
Blocking. The code side of this PR is sound, but the documentation it adds states a universal rule that the code does not implement, and the exception lands on exactly the parameters this PR is named after.

`taosmd/docs/a2a-comms.md:530` (added by this PR):

> Every `GET /a2a/*` endpoint rejects unknown query parameters with HTTP 400.

Two `GET /a2a/*` endpoints never call `_validate_a2a_params` at all.

**1. `GET /a2a/messages/{id}/receipts`**. The dispatch at `taosmd/http_server.py:1114-1116` passes no query dict, and `_handle_a2a_message_receipts` (`taosmd/http_server.py:1970`) has no `qs` parameter, so nothing is validated. Live against the merged tree:

```
GET /a2a/messages/12/receipts?after=9&since_id=3&bogus=1
  -> 200 {"delivered": [], "read": []}
```

`after=` and `since_id=` are silently dropped here, which is the precise defect class the PR title describes.

**2. `GET /a2a/receipts`**. `_handle_a2a_receipts` (`taosmd/http_server.py:1981`) takes `qs` but never validates it:

```
GET /a2a/receipts?message_id=12&agent=agentB          -> 404 {"error": "receipt not found"}
GET /a2a/receipts?message_id=12&agent=agentB&sinc=5   -> 404 {"error": "receipt not found"}
```

Identical responses. A validation call would raise 400 before the receipt lookup, so its absence is confirmed behaviourally, not just by reading.

Both were probed against the trial merge of `05fe6863` onto master `0baa9af9`, with a live server on the repo's own `live_server` fixture.

Separately, and as the Kilo review already noted, the table at `taosmd/docs/a2a-comms.md:538-546` omits `GET /a2a/census`, which does validate with an empty accepted set (`taosmd/http_server.py:1563-1567`). So the table is missing one endpoint that follows the rule and the prose overstates the rule for two that do not.

To unblock, either narrow the prose to the endpoints that actually enforce it and add `GET /a2a/census` to the table, or extend `_validate_a2a_params` to the two receipt read paths and add the accepted sets for them to the table. The second is the better outcome, since a strict-params contract with two silent holes in it is worse than no documented contract.

What I verified as good, so it is not relitigated on the next pass:

- Trial merge onto master `0baa9af9` is clean; merged-tree diff against master is exactly the four files, +172/-0, no revert of master-only work.
- `1816 passed, 10 skipped, 7 errors` on the merged tree; the 7 errors are the pre-existing `taosmd.mcp_server` import failures.
- `ruff check taosmd/ tests/ benchmarks/` clean, deleted-symbols guard clean.
- The 12 new tests are genuinely per-parameter, not one representative case. Mutating the messages allowlist to accept `after` alone fails 2 of 12; `since_id` alone fails 1 of 12; dropping `fields` alone fails 2; dropping `limit` alone fails 2; making the core check inert at `http_server.py:262` fails 8. Every allowlist branch is pinned separately, and so is the accepted-set half of the error message.
- No live caller regresses. `remote.py` sends only `thread`/`since`/`limit`, `reader`, `principal`, `channel`, `before`/`after`; `cli.py:332` sends `thread`/`since` to `/a2a/stream`; the bus sweep sends `since`/`limit` to `/a2a/messages`. All are on the relevant allowlists, and `?limit=1` still returns exactly one row.
- `since` semantics are untouched: `since=0` still returns 400 naming the epoch rule on both `/a2a/messages` and `/a2a/stream`, and a bare `GET /a2a/messages` still returns its default page.

One thing this PR does not cause but that sits on the same handlers: `GET /a2a/messages?limit=-1` returns 200 and the whole feed, because `limit_i` is passed straight to the service layer with no floor and SQLite reads `LIMIT -1` as unbounded. `limit=0` correctly returns nothing. Worth a separate card rather than scope creep here.

PR #422, opened 2026-08-30
REVISE PR #420: two GET /a2a/* endpoints never validate params, so the doc claim it adds is false

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #420: two GET /a2a/* endpoints never validate params, so the doc claim it adds is false

Autonomous build of board card tsk-7dxxx6.

PR #420 doc claim at a2a-comms.md:530 states every GET /a2a/* endpoint
rejects unknown query params with 400. Two endpoints never called
_validate_a2a_params:

  1. GET /a2a/messages/{id}/receipts -- dispatch passed no query dict, the
     handler _handle_a2a_message_receipts took no qs parameter. A request
     like ?after=9&since_id=3&bogus=1 returned 200 {"delivered": [], "read": []}
     with all params silently dropped.

  2. GET /a2a/receipts -- the handler _handle_a2a_receipts(qs) took qs but
     never validated it. A typo like ?sinc=5 returned 404 "receipt not found",
     indistinguishable from a genuine miss.

Fix: route both through _validate_a2a_params with their correct allowlists:
  - messages/{id}/receipts -> frozenset() (no query params accepted)
  - receipts -> frozenset({"message_id", "agent"})

Also fix the doc table at a2a-comms.md:538-546 which omitted GET /a2a/census
(already validated at http_server.py:1567) and the two receipt rows. The
sweeping claim is now true for all 10 GET /a2a/* handlers.

Enumeration of all 10 GET /a2a/* handlers from the dispatch table
(http_server.py:1073-1122), each checked for the missing-validate shape:

  1. GET /a2a/channels             -> _handle_a2a_channels           (validates, line 1562)
  2. GET /a2a/census               -> _handle_a2a_census             (validates, line 1567)
  3. GET /a2a/members              -> _handle_a2a_members            (validates, line 1572)
  4. GET /a2a/messages             -> _handle_a2a_messages           (validates, line 1739)
  5. GET /a2a/mentions             -> _handle_a2a_mentions           (validates, line 1768)
  6. GET /a2a/stream               -> _handle_a2a_stream             (validates, line 1840)
  7. GET /a2a/threads              -> _handle_a2a_threads            (validates, line 1889)
  8. GET /a2a/threads/{thr}/msgs   -> _handle_a2a_thread_messages     (validates, line 1897)
  9. GET /a2a/messages/{id}/receipts -> _handle_a2a_message_receipts  (NOW fixed)
 10. GET /a2a/receipts             -> _handle_a2a_receipts           (NOW fixed)

8 of 10 already validated; exactly the two receipt endpoints were missing.
No third endpoint is affected.

Tests added (5 new, all pass):
  - test_http_a2a_message_receipts_unknown_param_returns_400
  - test_http_a2a_receipts_unknown_param_returns_400
  - test_http_a2a_receipts_missing_required_param_returns_400
  - test_http_a2a_message_receipts_no_params_returns_200 (positive guard)
  - test_http_a2a_receipts_valid_params_returns_404 (positive guard)

Proof (tests/test_a2a_kinds_strict_params.py, -p no:randomly):
  RED before fix: 3 failed, 2 passed (rc=1)
  GREEN after fix: 31 passed (rc=0)
  Full test_http_server.py: 89 passed, no regressions
  ruff check: All checks passed

Docs-Reviewed: a2a-comms.md strict-params table updated to include census and
both receipt endpoints; the sweeping claim at line 530 is now true for every
GET /a2a/* handler.

Files:
 changelog.d/tsk-7dxxx6-receipt-param-validation.md |   3 +
 changelog.d/tsk-g35u5n-raw-bus-strict-params.md    |   3 +
 taosmd/docs/a2a-comms.md                           |  22 +++
 taosmd/http_server.py                              |  16 +-
 tests/test_a2a_kinds_strict_params.py              | 196 +++++++++++++++++++++
 5 files changed, 238 insertions(+), 2 deletions(-)

### DIFF
--- a/changelog.d/tsk-7dxxx6-receipt-param-validation.md
+++ b/changelog.d/tsk-7dxxx6-receipt-param-validation.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `GET /a2a/messages/{id}/receipts` and `GET /a2a/receipts` now reject unknown query parameters with HTTP 400, matching the strict-params contract documented for every other `GET /a2a/*` endpoint. Previously these two handlers never called `_validate_a2a_params`, so a misspelt parameter (e.g. `after=` or `since_id=` on the message-receipt path) was silently dropped and a typo like `sinc=` on `/a2a/receipts` was indistinguishable from a genuine 404 miss.
--- a/changelog.d/tsk-g35u5n-raw-bus-strict-params.md
+++ b/changelog.d/tsk-g35u5n-raw-bus-strict-params.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- The A2A read endpoints (`GET /a2a/messages`, `GET /a2a/stream`, `GET /a2a/mentions`, and all other `GET /a2a/*`) now reject unknown query parameters with HTTP 400, naming both the offending parameter and the accepted set. This closes the silent-dropping defect where `after=` and `since_id=` (message-id cursors that these endpoints never accepted) were silently ignored, making a misspelt cursor indistinguishable from a working one. The 400 surfaces match the controller proxy (taOS #2390). No shipped client in this repo sends `after` or `since_id` to these endpoints.
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -525,6 +525,25 @@ stored in `a2a_alarm_state` and survives restarts. Use
 Each channel in `/a2a/channels` has shape:
 `{"channel", "members", "message_count", "created_ts", "last_ts"}`
 
+### Strict query parameters
+
+Every `GET /a2a/*` endpoint rejects unknown query parameters with HTTP 400. The
+error response names both the offending parameter(s) and the accepted set, so a
+misspelt cursor (e.g. `after=` or `since_id=` on `/a2a/messages`, which only
+accepts `since` as a timestamp) is reported immediately rather than silently
+ignored. This mirrors the controller proxy (taOS #2390): an unknown query
+parameter is a 400, never a silent no-op. The accepted set is:
+
+| Endpoint | Accepted query parameters |
+|----------|--------------------------|
+| `GET /a2a/messages` | `thread`, `since`, `limit`, `fields`, `format` |
+| `GET /a2a/mentions` | `since`, `limit`, `reader` |
+| `GET /a2a/stream` | `thread`, `since` |
+| `GET /a2a/threads` | `principal` |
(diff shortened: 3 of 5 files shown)


jaylfc:
Adjudicated **CLEAN**, with one one-line doc change I made rather than asked for. This discharges all three of #420's blocked items.

The two named defects are fixed and were reproduced against the merged tree using the block comment's own URLs: `GET /a2a/messages/1/receipts?after=9&since_id=3&bogus=1` now returns 400 naming the offenders and the allowed set, where it previously returned 200 with an empty payload; and `GET /a2a/receipts?message_id=1&agent=agentB&sinc=5` now returns 400 where it previously returned a misleading 404 "receipt not found". The census table row is added.

The endpoints were enumerated from the dispatch chain rather than by grepping for the validator, which matters because a grep answers "where is the validator called" and not "which paths can take an unvalidated parameter". Ten dispatch entries, eleven reachable path shapes (the threads entry matches both with and without `/messages`), and all eleven were hit live: **0 of 11 unvalidated**. Six mutations on the gates this PR adds all die, including the parent's exact defect of passing `{}` instead of the query at the dispatch site.

**Why I edited the docs.** The prose this PR adds said every `GET /a2a/*` endpoint rejects unknown query parameters with a 400, without qualification, and that is not true for a parameter with an empty value: dispatch uses `parse_qs` at the default `keep_blank_values=False`, so `?bogus=` and a bare `?bogus` are accepted and ignored while `?bogus=1` is a 400. The examples chosen to illustrate the claim, `after=` and `since_id=`, were themselves the empty form, so they demonstrated the opposite of the sentence containing them.

#420 was blocked for shipping a doc claim that was false about its own change. I was not willing to let its revision ship the same thing one level narrower, so `9fe3d1a1` qualifies the sentence, gives the examples values, and states the mechanism and what it would cost to remove. `doc-gate` and `trailing-newline-gate` both clean after it.

The underlying behaviour is untouched and is master-wide, not this PR's doing. Whether it should change is a real decision with a real tradeoff, so it is carded as `tsk-hkiy66` rather than resolved here: `keep_blank_values=True` would fix it but changes how every endpoint reads every parameter.

One further gap, also not blocking and carded as `tsk-54omyt`: census is the only one of the eight pre-existing validators that nothing pins. Deleting its call leaves the full suite green at 1821 passed, while the other seven all go red when mutated the same way.

This now needs merge authority, which is not mine to give myself.

PR #424, opened 2026-08-30
GET /a2a/messages?limit=-1 returns the entire feed: the limit has no floor

jaylfc:
CARD TITLE (intent, not commit subject): GET /a2a/messages?limit=-1 returns the entire feed: the limit has no floor

Autonomous build of board card tsk-2opv44.

Files:
 changelog.d/tsk-2opv44-limit-negative rejection.md |  3 ++
 taosmd/http_server.py                              |  2 ++
 tests/test_a2a_kinds_strict_params.py              | 39 ++++++++++++++++++++++
 3 files changed, 44 insertions(+)

## Summary by CodeRabbit

- **Bug Fixes**
  - Negative `limit` values for `GET /a2a/messages` now return a clear 400 error instead of potentially returning an unbounded number of messages.
  - `limit=0` is accepted and returns no messages.
  - Positive limits return the requested number of messages.

- **Documentation**
  - Updated endpoint documentation with validation rules and default behavior for the `limit` parameter.

### DIFF
--- a/changelog.d/tsk-2opv44-limit-negative rejection.md
+++ b/changelog.d/tsk-2opv44-limit-negative rejection.md
@@ -1,3 +1,0 @@
-### Fixed
-
-- `GET /a2a/messages?limit=-1` now returns 400 instead of returning the entire message feed. SQLite treats `LIMIT -1` as unbounded, so negative limits are rejected with `_BadRequest`.
--- a/changelog.d/tsk-2opv44-limit-negative rejection.md
+++ b/changelog.d/tsk-2opv44-limit-negative rejection.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `GET /a2a/messages?limit=-1` now returns 400 instead of returning the entire message feed. SQLite treats `LIMIT -1` as unbounded, so negative limits are rejected with `_BadRequest`.
--- a/changelog.d/tsk-2opv44-limit-negative-rejection.md
+++ b/changelog.d/tsk-2opv44-limit-negative-rejection.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `GET /a2a/messages?limit=-1` now returns 400 instead of returning the entire message feed. SQLite treats `LIMIT -1` as unbounded, so negative limits are rejected with `_BadRequest`.
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -463,6 +463,14 @@ a2a_members(channel="CHANNEL")
 |--------|------|------------|----------|
 | `POST` | `/a2a/send` | body JSON `{"from", "body", "thread"?, "reply_to"?}` | `{"id", "from", "thread", "reply_to"}` |
 | `GET`  | `/a2a/messages` | `?thread=&since=&limit=&fields=&format=` | `{"messages": [...]}`; `fields=id,sender,body` projects each message down to those keys; `format=ndjson` emits one message per line (`application/x-ndjson`) |
+
+`limit` on `GET /a2a/messages` is bounded below as well as parsed. A negative
+value is rejected with `400 'limit' must not be negative`, and a non-integer
+with `400 'limit' must be an integer`. `limit=0` is valid and returns zero
+messages. Omitting the parameter, including passing it empty, falls through to
+the default page size rather than to an unbounded read. The floor matters
+because SQLite treats `LIMIT -1` as unbounded, so a cap written as
+`min(limit, N)` clamps only above and `?limit=-1` would return the whole feed.
 | `GET`  | `/a2a/stream` | `?thread=&since=` | SSE stream (`text/event-stream`) |
 | `GET`  | `/a2a/channels` | — | `{"channels": [...]}` |
 | `GET`  | `/a2a/members` | `?channel=<name>` | `{"members": [...]}` |
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1736,6 +1736,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                 limit_i = int(limit_raw)
             except (TypeError, ValueError) as exc:
                 raise _BadRequest("'limit' must be an integer") from exc
+            if limit_i < 0:
+                raise _BadRequest("'limit' must not be negative")
             if fmt not in ("json", "ndjson"):
                 raise _BadRequest("'format' must be 'json' or 'ndjson'")
             messages = runner.run(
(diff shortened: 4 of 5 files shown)


jaylfc:
## Review: the fix is correct, but the PR turns the suite red

The production change is right and it is genuinely pinned. Two mechanical defects in the surrounding files block it.

### Blocking 1: the changelog fragment turns the test suite red

`changelog.d/tsk-2opv44-limit-negative rejection.md` does not end in a newline (last byte `0x2e`). That fails `tests/test_trailing_newline_gate.py::TestRealRepositoryTree::test_gate_is_clean_against_the_real_repository`, so this is a suite failure, not just an advisory gate.

Local run on the head commit, which is a fast-forward from master:

```
1 failed, 1806 passed, 10 skipped, 7 errors in 225.28s
FAILED tests/test_trailing_newline_gate.py::TestRealRepositoryTree::test_gate_is_clean_against_the_real_repository
```

CI agrees: `1 failed, 1805 passed, 12 skipped`. The 7 errors are the pre-existing `taosmd.mcp_server` import failures and are not yours.

The offending file is new in this PR, so the failure is entirely attributable here. Appending one newline clears it:

```
scripts/check_trailing_newline.py  ->  trailing-newline-gate: clean
pytest tests/test_trailing_newline_gate.py  ->  13 passed
```

### Blocking 2: doc-gate

```
DOC-GATE FAIL: a2a-handlers -- changes to the A2A handlers (taosmd/http_server.py,
taosmd/service.py) require docs coverage in taosmd/docs/a2a-comms.md
(edit one of: taosmd/docs/a2a-comms.md, or add a 'Docs-Reviewed: <why>' trailer)
```

Neither is present. `taosmd/docs/a2a-comms.md` line 465 documents the `/a2a/messages` params table and still says nothing about the limit contract, so a one-line docs edit is the better of the two options.

### Minor: filename

The fragment is the only one of the 60 files in `changelog.d/` with a space in its name. Rename it to `tsk-2opv44-limit-negative-rejection.md`.

### What I verified about the fix itself

`http_server.py:1739-1740` is live, and every branch of it is pinned. Mutating by line number, on the PR's own tests:

| mutation | branch targeted | result |
| --- | --- | --- |
| line 1740 `raise` -> `pass` | floor fires at all | KILLED, 1 failed |
| line 1739 `< 0` -> `< -1000000` | floor is inert for `-1` | KILLED, 1 failed |
| line 1739 `< 0` -> `<= 0` | `limit=0` stays 200-with-0-rows | KILLED, 1 failed |
| line 1744 `limit=limit_i` -> `limit_i + 5` | the limit is actually applied | KILLED, 2 failed |

The three new tests each seed 21 rows against bounds of 0 and 1, so they can fail. Control: changing only the error string survived with 0 failed, which confirms the harness reports survivors rather than always reporting a kill.

Live probe, 21 rows seeded on `/a2a/messages`: `limit=-1` 400, `limit=-999` 400, `limit=abc` 400, `limit=0` 200 with 0 rows, `limit=1` 200 with 1 row, `limit=999999` 200 with 21. `?limit=` empty is dropped by `parse_qs` and falls through to the default of 50.

### Out of scope here, but worth a card

I enumerated the limit-taking endpoints from the dispatch table rather than by grepping for the clamp pattern, then probed them live.

No other `/a2a/*` endpoint leaks after this PR. `/a2a/mentions` already rejects `-1` and `0` deeper in the service. `/a2a/threads/{thread}/messages` accepts `-1` without a 400 but slices in Python, so it returns 1 row of 60 rather than the feed.

`GET /memories` is a third floorless site, and it is not the `/tasks/edges` one already carded. With 60 rows seeded against a default of 50:

```
/memories?limit=-1      -> 200, 60 rows
/memories?limit=-999    -> 200, 60 rows
/memories?limit=50      -> 200, 50 rows
/memories (no limit)    -> 200, 50 rows
```

`-1` returns more than the default, so it is genuinely unbounded, the same signature as the bug fixed here. `http_server.py:1466-1472` parses the int and passes it straight through.

I could not seed `/tasks`, `/tasks/ready`, `/tasks/edges`, `/graph` or `/pending` in the probe harness, so those are unverified empirically. At code level all of them parse the int with no floor, and `/tasks/edges` has `min(limit_i, 500)` only.

### Clean

Trial merge is a no-op, the head is a fast-forward from master `0baa9af9`. No conflict markers. `ruff 0.16.3` over `taosmd/ tests/ benchmarks/`: all checks passed. `check_deleted_symbols.py --base origin/master`: clean. Diff is additions only, `3/2/39`, no deletions and no mode changes. No em dashes in the added lines.

jaylfc:
Adjudicated, and I have fixed this one forward rather than closing it, because the production change is correct and the three blockers were mechanical.

What the review established about the fix itself, so it is on the record: four mutations aimed at individual branches all killed (`:1740` floor fires, `:1739` floor inert for -1, `:1739` the `limit=0` contract, `:1744` the limit actually applied), each with the failed count read in the same command and the error count checked alongside it. The control mutated only the error string and survived at `FAILED=0`, which is what proves the harness could have reported a survivor. The new tests seed 21 rows against bounds of 0 and 1, so unlike an earlier pair on this repo that seeded 9 rows against a cap of 500, these can actually fail.

Endpoints were enumerated from the dispatch table rather than by grepping for the clamp pattern, which matters because the grep answers "where are the clamps" and not "where can a limit go unbounded". Eleven handlers read `limit`, and no other `/a2a/*` endpoint is unbounded after this change.

What was blocking:

1. `changelog.d/tsk-2opv44-limit-negative rejection.md` had no trailing newline, which turns the whole suite red via `test_trailing_newline_gate` rather than failing quietly. The filename also carried a space. Renamed and terminated.
2. `doc-gate` refuses changes to the A2A handlers without matching coverage in `taosmd/docs/a2a-comms.md`. Added, next to the endpoint table, stating the contract that was actually measured: negative is a 400, non-integer is a 400, zero is valid and returns nothing, and an absent or empty parameter falls through to the default page size rather than to an unbounded read.

After `8eebd6bb`: `trailing-newline-gate: clean`, `doc-gate: clean`, `deleted-symbols-guard: clean`, `ruff 0.16.3` via `uv run` over `taosmd/ tests/ benchmarks/` passes, and `tests/test_trailing_newline_gate.py` is 13 passed where it was 1 failed / 12 passed.

One finding was out of scope and is carded separately as `tsk-au6qkw`: `GET /memories?limit=-1` returns 60 rows against a default of 50, which makes it a third floorless site distinct from this one and from `tsk-w6vwmo`. Note the mechanism there is proven only by symptom, not established, so that card says to determine it before fixing.

This now needs merge authority, which is not mine to give myself.

PR #427, opened 2026-08-31
fix(install-skill): version the taosmd-a2a skill, order versions, survive a bad manifest

jaylfc:
Supersedes #418, which was blocked. Fresh branch off `master` `0baa9af9`, not built on `exec/tsk-tkayq4`.

`taosmd install-skill` used to exit 0 with "already installed" whenever a `SKILL.md` was present, so no installed copy in the field ever received an update. This branch keeps that fix (packaged `version` read from the SKILL.md frontmatter, a content hash recorded in `.taosmd-skill-manifest.json` at install time, stale clean copies upgraded in place, locally edited copies refused without `--force`) and discharges the three defects the review of #418 raised.

## Defect 1: a damaged manifest crashed the tool and `--force` could not recover it

`_has_local_edits` ran `json.loads` on the manifest and was called before the `if force:` branch, so a truncated manifest took down both paths. Reproduced against the #418 implementation, copied verbatim into a scratch script:

```
force=False: UNCAUGHT JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 3 (char 2)
force=True:  UNCAUGHT JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 3 (char 2)
```

Manifest reading now goes through `_read_skill_manifest`, which returns `None` for a missing, unreadable, unparseable or non-object manifest. `_has_local_edits` falls through to the no-manifest content comparison in that case, so an unreadable manifest means unknown provenance rather than an exception. On this branch, same corrupt `{ truncated` manifest:

```
--- corrupt manifest, default path (force=False) ---
taosmd-a2a skill already installed, up to date (v1.0.0).
  rc=0
--- corrupt manifest, force path (force=True) ---
taosmd-a2a skill replaced: vunknown (pre-versioning) -> v1.0.0
  rc=0
  content advanced: True
  manifest valid: True
```

Both arms are tested separately (`test_corrupt_manifest_does_not_raise_without_force`, `test_corrupt_manifest_does_not_raise_with_force`), because a guard can be live on one branch and blind on the other. `test_corrupt_manifest_falls_back_to_the_content_comparison` pins that degrading means the no-manifest path and not "assume the copy is clean": with a corrupt manifest and a genuinely edited copy the run still refuses. `test_manifest_that_is_not_an_object_degrades` and `test_unreadable_manifest_degrades` cover the `[]` and `OSError` shapes.

One honest note: on the up-to-date path the run returns before writing anything, so a corrupt manifest stays corrupt there. It no longer affects the outcome, and the next run that actually installs rewrites it.

## Defect 2: I implemented real ordering

The docs and both message sites promised "only when the packaged copy is newer" while the code compared versions for string equality, so `1.1.0 -> 1.0.0` reported an upgrade and replaced a newer skill with an older one.

I chose real ordering over replace-on-any-difference. Two reasons. `taosmd/docs/a2a-comms.md` and the two comment sites already made the ordering promise to users, so implementing it keeps a shipped promise rather than weakening the documentation to match an implementation accident. And of the two directions this defect allows, quietly replacing a newer installed skill with an older packaged one is the one that actually loses a user's state, which is the direction this repo cares most about.

How it works, in `_version_tuple` and `_skill_change_kind`:

- numeric dotted versions are compared as integer tuples, no new dependency;
- if either side does not parse as a plain numeric dotted version, there is no ordering, so the packaged copy still replaces a clean installed one but the message says "replaced", never "upgraded";
- a downgrade is refused with rc 1 and a message pointing at `--force`;
- a forced downgrade prints "downgraded". The word "upgraded" is unreachable on any downgrade path.

The card warned specifically that ordering was never performed at all, so `0.9.0 -> 0.10.0` already upgraded correctly and must not regress. It does not: `test_multi_digit_versions_upgrade_in_numeric_order` pins that direction and `test_multi_digit_downgrade_is_refused` pins the reverse, which a string compare would get backwards.

The user-visible wording changes on one credited path. The pre-versioning migration used to print `upgraded: vunknown (pre-versioning) -> v1.0.0`; it now prints `replaced`, because a copy with no version line cannot be ordered against anything. The mechanics of that migration are unchanged: rc 0, content advances, manifest written.

`taosmd/docs/a2a-comms.md`, the `_run_install_skill` docstring and the comment on the final copy all now describe exactly these semantics, including the numeric comparison, the "replaced" wording and the downgrade refusal.

## Defect 3: the six unpinned lines

Each line was neutered alone, by line number, in `taosmd/cli.py`, then `tests/test_install_skill.py tests/test_remote.py` was run. The mutant source was `compile()`d first so a syntax error cannot masquerade as a survivor, and the ERROR count is read alongside the FAILED count. Baseline on this branch is 51 passed, 0 FAILED, 0 ERROR.

| line | mutation | FAILED | ERROR | verdict |
| --- | --- | --- | --- | --- |
| 622 | `if not text.startswith("---"):` to `if False:` | 1 | 0 | killed |
| 624 | `end = text.find("---", 3)` to `end = len(text)` | 2 | 0 | killed |
| 724 | null-hash `return False` to `return True` | 1 | 0 | killed |
| 729 | `_body` drops the `version:` filter | 1 | 0 | killed |
| 776 | `_parse_skill_version(packaged_md) or "0.0.0"` loses `or "0.0.0"` | 1 | 0 | killed |
| 799 | `if installed_version == packaged_version and not has_local_edits:` loses `and not has_local_edits` | 1 | 0 | killed |

Line 729 is the decisive one. The test that kills it, `test_pre_versioning_copy_differing_only_by_the_version_line_migrates`, models what every installed copy in the field actually is: no manifest, no `version:` line, and a body that differs from the packaged one by exactly the added version line. That is not hypothetical. Diffing `origin/master`'s `SKILL.md` against this branch's:

```
master SKILL.md has a version line: False
lines differing between installed and packaged: 1

--- real migration: master copy, no manifest (force=False) ---
taosmd-a2a skill replaced: vunknown (pre-versioning) -> v1.0.0
  rc=0
  content advanced: True
  manifest valid: True
```

Under the line 729 mutant that same run flips to rc 1 "has local edits", which is the failure the review measured.

The other five: 622 and 624 are pinned by three `_parse_skill_version` tests (no frontmatter at all, an unclosed frontmatter block, and a `version:` line below the frontmatter). 724 by `test_manifest_without_a_hash_is_treated_as_clean`. 776 by `test_packaged_skill_without_a_version_records_the_zero_fallback`. 799 by `test_same_version_with_local_edits_is_refused`.

## The kills #418 already had, re-measured, plus the new branches

Everything the review credited still goes red, and the code this branch adds is pinned the same way. Same harness, same counts read the same way.

| line | mutation | FAILED | ERROR |
| --- | --- | --- | --- |
| 627 | frontmatter regex to `m = None` | 13 | 0 |
| 665 | `_skill_change_kind` unorderable guard to `if False:` | 4 | 0 |
| 667 | `if packaged > installed:` to `if False:` | 1 | 0 |
| 669 | `if packaged < installed:` to `if False:` | 3 | 0 |
| 706 | `except (OSError, ValueError):` to `except ():` | 6 | 0 |
| 708 | `isinstance(data, dict)` guard dropped | 1 | 0 |
| 725 | manifest hash compare to `return False` | 3 | 0 |
| 780 | `_write_skill_manifest` on the fresh-install path | 8 | 0 |
| 789 | `_copy_skill_tree` on the force path | 5 | 0 |
| 790 | `_write_skill_manifest` on the force path | 3 | 0 |
| 794 | force-path message hardcoded to "upgraded" | 1 | 0 |
| 806 | `if has_local_edits:` refusal to `if False:` | 5 | 0 |
| 836 | `_copy_skill_tree` on the upgrade path | 4 | 0 |
| 837 | `_write_skill_manifest` on the upgrade path | 2 | 0 |
| 839 | upgrade-path message hardcoded to "upgraded" | 2 | 0 |

Line 706 is the one that matters for defect 1: restoring the raising behaviour fails 6 tests. Lines 794 and 839 are the ones that matter for defect 2: hardcoding "upgraded" back into either message fails.

`cli.py` was restored and verified byte-identical to the committed version after every mutant.

## Suite and gates

Full suite on this branch, in the branch's own worktree venv:

```
IMPORTED: /home/jay/.cache/taosmd-review/sot6iu/taosmd/__init__.py
1827 passed, 10 skipped, 7 errors in 276.05s
FAILED count: 0
```

1827 passed against the 1809 the #418 merged tree reported, which is the 18 tests this branch adds. The 7 errors are the pre-existing `taosmd.mcp_server.build_server` import failures, unchanged, and they are why the pytest exit code is permanently 1 here.

- `ruff check taosmd/ tests/ benchmarks/` with ruff 0.16.3: `All checks passed!`
- `scripts/check_deleted_symbols.py --base origin/master`: `deleted-symbols-guard: clean`
- `scripts/check_trailing_newline.py`: `trailing-newline-gate: clean`
- `scripts/check_doc_gate.py invariants` and `diff-gate --base origin/master`: `doc-gate: clean`
- conflict markers: none
- em dashes in added lines: 0, grep verified against a positive control

## One cleanup the review flagged as non-blocking

`_install_skill_cmd` had an `importlib.resources` fallback that computed `_pkg_files("taosmd").joinpath("skills/taosmd-a2a")`, threw the result away, and reassigned `skill_src_dir` to the value it already held. It could never change an outcome. Rather than leave a bare expression statement there or invent a wheel path that is not exercised, I removed it. Skill assets ship inside the package directory, so `Path(__file__).parent / "skills" / "taosmd-a2a"` is the correct source for source checkouts and wheels alike, and the "packaged skill not found" error still covers the case where they are missing.

## Summary by CodeRabbit

* **New Features**
  * `taosmd install-skill` now detects installed and packaged skill versions.
  * Automatically installs new skills and upgrades outdated copies.
  * Records installation metadata to support content comparisons and update checks.
* **Bug Fixes**
  * Prevents accidental overwriting of locally edited skills without `--force`.
  * Blocks downgrades unless explicitly forced.
  * Handles missing or unreadable installation metadata gracefully.
* **Documentation**
  * Added guidance for checking skill versions and managing updates.


jaylfc:
Blocking this and closing it in favour of a fresh branch. The revision card is `tsk-k2hwu5`.

Reviewed independently on head `d0d647da` against master `0baa9af9`, then the structural claims were re-checked a second time.

**The test work here is good.** Trial merge is clean, the merged tree matches the PR head tree, and the suite goes 1804 / 10 / 7 on master to 1827 / 10 / 7 on the trial merge, which is exactly the 23 new tests. A 16 mutant campaign killed every mutant, and all six branches that #418 called out as unpinned survivors now kill. Gates are green, lint passes at the pinned 0.16.3, and the CI step conclusions were read individually rather than off the overall tick.

It is blocked on two things the tests cannot see.

**1. The repo's own installer auto-forces the refusal this PR adds.** `scripts/install-client.sh` line 48 is:

```
taosmd install-skill || taosmd install-skill --force
```

That file is not in this PR, so it is unchanged shipped code. Before this change the `||` was effectively dead, because a non-zero exit only meant missing assets. Making refusal exit 1 turns that fallback into an automatic forced overwrite of exactly the local edits the refusal exists to protect. Measured end to end against the merged tree with an edited `SKILL.md`: the refusal prints, the forced arm then reports `reinstalled: v1.0.0 -> v1.0.0 (overwriting local edits)`, overall rc is 0, and the local edit is gone. So the changelog line "user edits are never silently clobbered" is not true along the path the project actually ships, and the downgrade refusal is auto-forced there as well. `scripts/install-client.ps1` lines 49 to 53 look like the same defect, since a `try/catch` around a native command does not fire on a non-zero exit in PowerShell, but there is no PowerShell on the review host so treat that as analysis and confirm it directly.

**2. The manifest write still raises uncaught, on both arms.** #418 was blocked because the escape hatch its error message advertises was unreachable. The read path is fixed and proven. The write path is not: `_write_skill_manifest` calls `write_text` with no guard, and both call sites run it after `_copy_skill_tree`. With a manifest that exists but cannot be written, a directory in its place (this PR's own `test_unreadable_manifest_degrades` scenario) or a read-only file, plus a stale installed version, both the force and non-force arms raise uncaught. The PR only exercises that scenario on the equal-version arm, whose early return skips the write. Because the copy has already happened, `SKILL.md` advances while the manifest still records the old hash, so the next run misreports local edits.

Two smaller ones, both in scope on the revision:

- A manifest that is valid JSON but has no `skill_md_sha256` is treated as clean, so genuine local edits plus a newer package destroys the edit with no `--force`. That is deliberate and pinned by a test, but it cuts against zero loss and the safer fallback is about a line. Worth an explicit decision either way.
- `^version:\s*(\S.*)$` under `re.MULTILINE`: `\s` crosses the newline, so an empty `version:` line consumes the following line as the version and prints a garbage message. It degrades to a replace, so it is cosmetic. `[ \t]*` fixes it.

The acceptance on the revision is red first, and for the installer defect it has to be an end to end probe that runs the real script and shows the edit surviving. A unit test on the CLI cannot pin it, because the defect is in the caller. Note that the 7 pre-existing `taosmd.mcp_server.build_server` import errors hold pytest rc at 1 permanently, so read failed and error counts out of the summary line, never rc.

PR #430, opened 2026-08-31
REVISE PR #425: the read ACL now bricks posting, seven branches are unpinned, and its own tests never reach the restricted channel

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #425: the read ACL now bricks posting, seven branches are unpinned, and its own tests never reach the restricted channel

Autonomous build of board card tsk-4fyiyw.

DEFECT 1: post default independent of read list. _normalize_acl now
defaults missing read and post keys to ["*"], so set_acl(read_ids=[...])
leaves posting open unless post_ids is explicitly restricted.

DEFECT 2: seven unpinned lines plus vacuous tests. Added tests that
exercise the channels, census, members, threads, messages-no-thread,
stream-no-thread, and over-fetch branches. Members test now posts to
the restricted channel so the filter is reachable. Over-fetch test
posts public first with 60 restricted rows so removing the multiplier
fails.

DEFECT 3: ordering. REST and SSE handlers now fetch 10**9 rows when
thread is None, so ACL filtering happens before the most-recent-N
window can starve public rows. Added a bound test seeding 600
restricted rows; it fails on the old limit_i*10 code and passes after.

DEFECT 4: malformed acls section. get_acl now returns deny/closed
when the acls section exists but is not a dict, instead of falling
through to a wildcard.

Tests: 1825 passed, 12 skipped on full suite; 22 passed in
test_a2a_channel_acls.py.

Proof: uv run --extra dev pytest tests/ -q -p no:randomly

Docs-Reviewed: a2a-comms.md already documents set-channel-acl, channel-acl, and ACL filtering behavior; fixes are defect corrections, not new user-facing surfaces

Files:
 taosmd/docs/a2a-comms.md                   |  35 ++
 tests/test_a2a_channel_acls.py             | 713 +++++++++++++++++++++++++++++
 6 files changed, 1026 insertions(+), 13 deletions(-)

### DIFF
--- a/changelog.d/tsk-4fyiyw-a2a-acl-revision.md
+++ b/changelog.d/tsk-4fyiyw-a2a-acl-revision.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- A2A per-channel ACL: post allowlist now defaults to open when not explicitly set, so restricting reads no longer bricks all posting on that channel.
+- A2A per-channel ACL: malformed `acls` configuration sections (non-dict values) now fail closed instead of falling through to a wildcard allow.
+- A2A per-channel ACL: messages and stream feeds without a thread filter now fetch all rows before ACL filtering so public rows are not starved by restricted ones.
--- a/changelog.d/tsk-b3qryl-a2a-acl-bypass.md
+++ b/changelog.d/tsk-b3qryl-a2a-acl-bypass.md
@@ -1,0 +1,5 @@
+### Fixed
+- A2A per-channel read ACL: filter channels, members, census, and thread-messages endpoints so restricted channels are not disclosed to unauthenticated or unauthorized callers
+- A2A admin channel-acl GET route: make it reachable with admin_token independent of server_token
+- A2A message feed limit starvation: over-fetch when no thread filter is set so restricted rows cannot consume the public limit window
+- ACL config parsing: deny on malformed read/post values and list-shaped entries instead of failing open to wildcard
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -69,6 +69,8 @@ _HUMAN_PRINCIPAL_IDS_KEY = "human_principal_ids"
 # one of these directories. Empty (the default) means collections are off.
 _COLLECTIONS_SECTION_KEY = "collections"
 _COLLECTIONS_ALLOWED_ROOTS_KEY = "allowed_roots"
+# Per-channel ACL for the A2A bus.
+_ACL_SECTION_KEY = "acls"
 
 MANAGED_BY_STANDALONE = "standalone"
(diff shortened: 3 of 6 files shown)


jaylfc:
## Retracting my earlier CLEAN verdict on this PR, and closing it in favour of `tsk-ejgxhs`.

I passed this PR as clean and parked it for merge. That verdict was wrong on one point and incomplete on another, and an independent adjudication of #447 surfaced both. Recording it here rather than quietly reversing the disposition.

### What my verdict missed

**kilo posted two CRITICALs on this PR and I did not action them.** They are correct:

```
taosmd/http_server.py:1815   GET /a2a/messages (no thread) fetches up to 10**9 rows per request
taosmd/http_server.py:1946   SSE /a2a/stream (no thread) repeats that 10**9 scan every second, per open stream
```

That is a denial-of-service vector, and by kilo's own summary it supersedes the defects this PR fixes. The lesson for me is narrow and worth stating: I read this PR's bot output for outages wearing verdicts' costumes, which is the usual failure here, and did not then act on the finding that was actually valid. A bot verdict is a claim to triage in **both** directions.

**The vacuity count was understated.** I recorded that the regression vacates five tests. Measured properly, with the read filter made inert and the post default denied together, it vacates **eight**:

```
V1  this PR's 22 tests, read filter inert                      14 passed / 8 failed / 0 error
V2  this PR's 22 tests, read filter inert + post default deny  22 passed / 0 failed / 0 error   <- all 8 went vacuous
```

**And the ACL deny branches are pinned on status, not effect.** Flipping `return False` to `return True` at `http_server.py:875` or `:883` leaves every test green, while a read-back shows the denied message was persisted anyway.

### Why closed rather than revised in place

#447 is a strict superset of this PR. Its three production blobs are byte-identical to these (`config.py ca8115aac2b1`, `http_server.py 1e9bd52ca79b`, docs `72a3911d458a`), and it adds the missing coverage this PR's tests lacked. The two cannot both merge: add/add conflict on `tests/test_a2a_channel_acls.py`. #447 is closed too, for the defects above, so both branches are retired into one re-cut card rather than leaving a superset and a subset open against each other.

### Nothing is lost, and the fix is not on master

To be explicit, because this is the failure mode I am guarding against: **the channel-ACL open-default fix now exists in no open PR and is not on master.** `tsk-ejgxhs` is its only carrier, it starts fresh off master, and it records the exact head SHAs (`ca2e722d` here, `8ab4790c` for #447) so the payload is re-landed rather than rebuilt. The card also names what is already discharged and proven, so no lane redoes work: the `['*']` default fix is correct and must not be reverted.

The code fix in this PR was right. It is the coverage around it, and an unactioned CRITICAL, that stop it merging.

PR #431, opened 2026-08-31
mentions.py opens SQLite without check_same_thread=False under ThreadingHTTPServer: the same exposure #421 just closed for receipts

jaylfc:
CARD TITLE (intent, not commit subject): mentions.py opens SQLite without check_same_thread=False under ThreadingHTTPServer: the same exposure #421 just closed for receipts

Autonomous build of board card tsk-wxymim.

MentionStore.init() opened its SQLite connection via _db.connect(db_path),
which defaults to sqlite3.connect check_same_thread=True. The connection is
stored on the instance and reused across all async method calls (the store is
cached in _ensure_stores), so it is genuinely shared. Under the
ThreadingHTTPServer model (each request on its own thread) the connection
raised ProgrammingError when used from a thread other than the one that
created it. This is the same exposure PR #421 closed for ReceiptStore.

The fix:
- _db.connect gains a keyword-only check_same_thread parameter (default
  True), leaving all 15 existing single-arg call sites untouched -- zero
  blast radius on the shared helper.
- MentionStore.init passes check_same_thread=False, matching ReceiptStore.

Tests: two new tests in tests/test_a2a_mentions.py:
- test_mention_store_uses_wal_and_busy_timeout: pins WAL + 5000 ms timeout.
- test_mention_store_usable_from_non_creating_thread: spawns a real
  threading.Thread, writes, and reads back. RED with check_same_thread
  removed (ProgrammingError), GREEN with the fix.

Proof (pytest -p no:randomly tests/test_a2a_mentions.py):
  32 passed, 0 failed, 0 errors

Negative control (check_same_thread=True trap):
  1 failed, ProgrammingError: SQLite objects created in a thread can
  only be used in that same thread.

ruff check on changed files: All checks passed!

Files:
 .../tsk-wxymim-mention-store-thread-affinity.md    |  6 ++
 taosmd/_db.py                                      | 14 ++++-
 taosmd/mentions.py                                 | 13 +++-
 tests/test_a2a_mentions.py                         | 69 ++++++++++++++++++++++
 4 files changed, 98 insertions(+), 4 deletions(-)

### DIFF
--- a/changelog.d/tsk-wxymim-mention-store-thread-affinity.md
+++ b/changelog.d/tsk-wxymim-mention-store-thread-affinity.md
@@ -1,0 +1,6 @@
+### Fixed
+- MentionStore opened its SQLite connection without `check_same_thread=False`, so
+  the thread-affine connection raised `ProgrammingError` when the A2A mention feed
+  was served from a `ThreadingHTTPServer` request thread other than the one that
+  created it. Routed through `taosmd._db.connect(..., check_same_thread=False)`,
+  matching the ReceiptStore fix.
--- a/taosmd/_db.py
+++ b/taosmd/_db.py
@@ -26,14 +26,24 @@ from typing import Union
 BUSY_TIMEOUT_MS = 5000
 
 
-def connect(db_path: Union[str, Path]) -> sqlite3.Connection:
+def connect(
+    db_path: Union[str, Path],
+    *,
+    check_same_thread: bool = True,
+) -> sqlite3.Connection:
     """Open a SQLite connection in WAL mode with a busy timeout.
 
     Drop-in replacement for ``sqlite3.connect(db_path)``. Callers that need a
     ``row_factory`` or other connection attributes should set them on the
     returned connection as before.
+
+    ``check_same_thread`` defaults to ``True`` to preserve sqlite3's default
+    thread-affinity contract for all existing call sites. Stores driven by
+    async methods that may be awaited from any thread pass
+    ``check_same_thread=False`` explicitly (e.g. ``ReceiptStore``,
+    ``MentionStore``).
     """
-    conn = sqlite3.connect(db_path)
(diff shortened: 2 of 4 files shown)


jaylfc:
Adjudicated as part of a single cluster review covering #451, #462, #431 and #444 together, against current master `bb8dfc09`. Closing this one as contradicted rather than blocking it, because there is nothing here to re-land.

**The premise does not hold.** This PR flips `MentionStore` to `check_same_thread=False`; the measurement says the default is correct. A probe on `bb8dfc09` drove six concurrent client threads through `POST /a2a/send`, `POST /tasks` and receipts POST/PATCH: 7 connections, **0 cross-thread**, with a control arm that did register a cross-thread use so the zero is an absence and not a blind probe. `a2a-mentions.db` served all six concurrent sends with `check_same_thread=True`, all 200. `_ServiceLoop.run` dispatches via `run_coroutine_threadsafe` and there is no loop bypass in `http_server.py`, so this is structural.

The changelog claims a `ProgrammingError` "served from a ThreadingHTTPServer request thread". That path does not exist in the current design, and the test manufactures the scenario rather than reproducing it.

**The busy-timeout test does not discriminate.** It hard-codes `== 5000`, which is CPython's own default, so it passes whether or not the code under test does anything. Measured: deleting the `PRAGMA busy_timeout` line entirely leaves FAILED=0, a survivor, while the control mutant fires at FAILED=1.

For what it is worth, the shared `_db.connect` change this PR carries has been on master since #458, so its `_db.py` conflict was only ever a docstring conflict. The genuinely useful work in this area is #451 (the `run_schema` primitive, fully mutation-pinned at 5 of 5) and #444 (whose test is the only thing catching the #421 to #434 regression). Both are with Jay for a merge call.

No re-cut card: the question this PR existed to answer has been answered the other way, and `tsk-bsbwmb` already covers the one real residue, that `ReceiptStore` holds no lock.

PR #435, opened 2026-08-31
collections revoke reports success when it removed nothing: no rowcount signal on a non-matching grantee

jaylfc:
CARD TITLE (intent, not commit subject): collections revoke reports success when it removed nothing: no rowcount signal on a non-matching grantee

Autonomous build of board card tsk-siwsp7.

CollectionStore.revoke captured the DELETE rowcount but discarded it, so a
revoke of a grantee that holds no grant returned the collection as if it had
succeeded and stayed 200. DELETE /collections/{id}/grants/{agent} now
surfaces a top-level boolean `revoked` alongside the existing `collection`
key; a no-op revoke remains 200 (idempotent). The case-sensitive,
whitespace-normalised grantee match and grant validation are unchanged.

Proof (fresh branch off 0baa9af9, worktree taosmd imported from
/tmp/exec-tsk-siwsp7/taosmd/__init__.py):
  ./.venv/bin/python -m pytest -q -p no:randomly tests/test_collections_store.py tests/test_collections_http.py
  -> 42 passed, FAILED=0, ERROR=0, rc=0

Neuter gate, both arms independently live:
  hardcode revoked=True  -> 2 FAILED (non-matching: store + http)
  hardcode revoked=False -> 2 FAILED (matching: store + http)
  drop the revoked field -> 5 FAILED + 3 ERROR

Docs-Reviewed: the only http_server.py edit is _handle_collections_revoke on the collections grant endpoint, which taosmd/docs/a2a-comms.md does not document (it covers only A2A auth and the revoked-agent registry feed), and no A2A behaviour changed. The user-visible fix is recorded in changelog.d/tsk-siwsp7-revoke-rowcount-signal.md.

Files:
 changelog.d/tsk-siwsp7-revoke-rowcount-signal.md |  2 ++
 taosmd/collections.py                            |  7 +++--
 taosmd/http_server.py                            |  3 +-
 tests/test_collections_http.py                   | 37 ++++++++++++++++++++++++
 tests/test_collections_store.py                  | 26 +++++++++++++++++
 5 files changed, 72 insertions(+), 3 deletions(-)

### DIFF
--- a/changelog.d/tsk-siwsp7-revoke-rowcount-signal.md
+++ b/changelog.d/tsk-siwsp7-revoke-rowcount-signal.md
@@ -1,0 +1,2 @@
+### Fixed
+- Collection grant revocation now reports whether a row was actually removed. ``CollectionStore.revoke`` captured the ``DELETE`` rowcount but discarded it, so ``DELETE /collections/{id}/grants/{agent}`` returned 200 even when no grant matched, telling an operator the revoke succeeded while the grant stayed live. The endpoint now returns a boolean ``revoked`` field alongside ``collection``; a no-op revoke stays 200 (idempotent), not an error. The grantee match is still case-sensitive and whitespace-normalised, unchanged.
--- a/taosmd/collections.py
+++ b/taosmd/collections.py
@@ -405,13 +405,16 @@ class CollectionStore:
 
     def revoke(self, collection_id: str, canonical_id: str) -> dict:
         self._row(collection_id)
-        self._conn.execute(
+        cur = self._conn.execute(
             "DELETE FROM collection_grants "
             "WHERE canonical_id = ? AND scope = ? AND collection_id = ?",
             (_grantee_key(canonical_id), GRANT_SCOPE, collection_id),
         )
+        removed = cur.rowcount
         self._conn.commit()
-        return self.get(collection_id)
+        col = self.get(collection_id)
+        col["revoked"] = removed > 0
+        return col
 
     def has_grant(self, canonical_id: str, collection_id: str) -> bool:
         row = self._conn.execute(
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -2454,7 +2454,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             except CollectionNotFoundError as exc:
                 self._send_json(404, {"error": str(exc)})
                 return
-            self._send_json(200, {"collection": col})
+            revoked = col.pop("revoked")
+            self._send_json(200, {"collection": col, "revoked": revoked})
 
         def _handle_collections_delete(self, collection_id: str) -> None:
             if not self._check_admin_token():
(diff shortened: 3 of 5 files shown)


jaylfc:
Adjudicated and blocked. The store and HTTP halves of this change are correct and mutation-verified, so this is not a rejection of the approach, but the change does not finish the job it was filed for.

**The CLI caller discards the new signal, so the original defect survives verbatim on that surface.** `collections_revoke` has exactly two callers. The HTTP handler reads `revoked`. `taosmd/cli.py` near line 1090 prints only `id` and `grants` and then returns 0 unconditionally:

```
CLI revoke that REMOVED a grant : rc=0 out='col-... grants: (none)'
CLI revoke that REMOVED NOTHING : rc=0 out='col-... grants: (none)'
service layer DOES carry it     : removed=True vs noop=False
```

Identical stdout and identical exit code for "access withdrawn" and "there was nothing to withdraw", which is the false negative this work was meant to remove. `cli.py` is not in this PR's file list, so no mutation campaign could have surfaced it; it was found by checking the callers. `tests/test_collections_cli.py` near line 75 asserts only `rc == 0`, so nothing would catch the regression either.

Two smaller items to fold in:

1. `docs/collections.md` line 66 still documents the old `{"collection": {...}}` envelope for `DELETE /collections/{id}/grants/{agent}`. No gate protects that file, since the doc rule for `http_server.py` points at `taosmd/docs/a2a-comms.md`, so the drift is invisible to CI.
2. Swapping `col.pop` for `col.get` in the HTTP handler left the targeted set fully green (0 failed, 0 errors, 47 passed). Nothing pins that `revoked` is stripped from the inner `collection`, so the record shape can silently gain a field.

Measurements taken on a trial merge against master `0baa9af9`: suite 1808 passed / 10 skipped / 7 errors, which is the baseline 1804/10/7 plus exactly the four new tests. Lint clean at the pinned ruff version with a positive control confirming it examined the changed test file. Mutation, by line number, failed and error counts together: store not-found 2/0, store matched 2/0, HTTP false branch 1/0, HTTP true branch 1/0. All four new tests use `is True` / `is False` and each killed the branch it names. The end-to-end probe was correct on double-revoke, never-granted, whitespace grantee and missing collection.

One note on the PR body: its neuter transcript is directionally right but the counts belong to the store line, not the HTTP line, where the same hardcodes give 1 each rather than 2. The "drop the revoked field" arm reporting 3 errors is a syntax-error mutant rather than a clean mutation. Nothing stated was false, but the "both arms independently live" claim rests on the store mutation alone.

Closing this rather than holding it open, because a blocked PR on this board comes back as a new one. Follow-up card **tsk-d6giky** carries all three items and starts fresh off master; it records what is already proven so the store and HTTP work does not get re-derived from nothing. The pre-existing grant/revoke validation asymmetry is filed separately as **tsk-cfaf6o**, since a test deliberately pins the current behaviour and changing it is a decision rather than a fix.

PR #428, opened 2026-08-31
REVISE PR #414: the membership routes never percent-decode {thread} while remote.py now encodes it, so any thread with a space silently returns an empty member list

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #414: the membership routes never percent-decode {thread} while remote.py now encodes it, so any thread with a space silently returns an empty member list

Autonomous build of board card tsk-qzlijt.

Revise PR #414 (tsk-qzlijt): the three thread-membership HTTP routes
extract {thread} from the URL path without unquote(), so a thread name
containing a space that the RemoteClient percent-encodes as proj%20x
never matches the stored name proj x and silently returns an empty
member list.  Match the existing POST /a2a/alarms/{key}/clear sibling:
server-side unquote() in _handle_a2a_list_members,
_handle_a2a_add_member, and _handle_a2a_remove_member; client-side
quote(thread, safe='') in the RemoteClient methods, fixing the raw
f-strings that raised http.client.InvalidURL on thread names with spaces.

Tests: 4 new tests covering all three routes plus the GET-list path that
returned empty, including an end-to-end RemoteClient probe against a live
server with a space in the thread name.  Removing unquote() yields 4
FAILED (count verified on /tmp/r.txt); restoring returns 47 passed.
uv run ruff 0.16.3 over taosmd/ tests/ benchmarks/: All checks passed.

Includes the full PR #414 changeset: MembershipStore, service-layer
a2a_create_thread/a2a_list_members/a2a_add_member/a2a_remove_member,
HTTP handlers, route dispatch, migrations, RemoteClient forwarding,
positive-control restoral, 25 mutation-pinned branches, and docs.

Files:
 taosmd/docs/a2a-comms.md                         |  33 +
 taosmd/http_server.py                            | 111 +++-
 taosmd/migrations.py                             |  35 +
 taosmd/remote.py                                 |  60 ++
 taosmd/service.py                                | 223 ++++++-
 tests/test_a2a_membership.py                     | 806 +++++++++++++++++++++++
 tests/test_a2a_remote_forwarding.py              | 112 ++++
 14 files changed, 1716 insertions(+), 7 deletions(-)

## Summary by CodeRabbit

- **New Features**
  - Added A2A thread membership management for creating threads, listing members, adding members, and removing members.
  - Added owner/member roles, ownership protections, persistent membership storage, and membership event recording.
  - Added remote client support for membership operations.

- **Bug Fixes**
[...]

PR #437, opened 2026-08-31
GET /a2a/threads/{thread}/messages is the FOURTH thread-encoding asymmetry, and it is live on master today

jaylfc:
CARD TITLE (intent, not commit subject): GET /a2a/threads/{thread}/messages is the FOURTH thread-encoding asymmetry, and it is live on master today

Autonomous build of board card tsk-itkfwo.

Revise PR #414 (tsk-qzlijt): the three thread-membership HTTP routes
extract {thread} from the URL path without unquote(), so a thread name
containing a space that the RemoteClient percent-encodes as proj%20x
never matches the stored name proj x and silently returns an empty
member list.  Match the existing POST /a2a/alarms/{key}/clear sibling:
server-side unquote() in _handle_a2a_list_members,
_handle_a2a_add_member, and _handle_a2a_remove_member; client-side
quote(thread, safe='') in the RemoteClient methods, fixing the raw
f-strings that raised http.client.InvalidURL on thread names with spaces.

Tests: 4 new tests covering all three routes plus the GET-list path that
returned empty, including an end-to-end RemoteClient probe against a live
server with a space in the thread name.  Removing unquote() yields 4
FAILED (count verified on /tmp/r.txt); restoring returns 47 passed.
uv run ruff 0.16.3 over taosmd/ tests/ benchmarks/: All checks passed.

Includes the full PR #414 changeset: MembershipStore, service-layer
a2a_create_thread/a2a_list_members/a2a_add_member/a2a_remove_member,
HTTP handlers, route dispatch, migrations, RemoteClient forwarding,
positive-control restoral, 25 mutation-pinned branches, and docs.

Files:
 taosmd/docs/a2a-comms.md                         |  33 +
 taosmd/http_server.py                            | 111 +++-
 taosmd/migrations.py                             |  35 +
 taosmd/service.py                                | 223 ++++++-
 tests/test_a2a_membership.py                     | 806 +++++++++++++++++++++++
 tests/test_a2a_remote_forwarding.py              | 112 ++++
 14 files changed, 1716 insertions(+), 7 deletions(-)


jaylfc:
Closing this in favour of #428, which is the same work.

The two branches have the identical tree:

```
git rev-parse 00bcadbd^{tree} e6480e2a^{tree}
841179899b809d90281e600f7cabf0e48643a47f
841179899b809d90281e600f7cabf0e48643a47f
git diff --numstat 00bcadbd e6480e2a | wc -l   ->  0
control: master..#428 = 14 files, master..#437 = 14 files, numstats identical across all 14
```

Different commit SHAs, identical content, and the author dates match to the second, so this is the membership work re-committed onto a second branch rather than new work. The changelog fragments here are `tsk-6pned6`, `tsk-a3rwa4`, `tsk-qzlijt` and `tsk-v4cbhv`, all of them #428's, and the commit subject is the membership percent-decode.

The card this PR is branched and titled for is a different defect, and none of it is present here. `taosmd/remote.py:319` is still `f"/a2a/threads/{thread}/messages"`, byte identical to master, and the handler body diffs clean against master.

Worth recording, because it changes that card rather than just deferring it. Enumerating the dispatch table for every caller-supplied path segment, rather than grepping for decode calls, the thread read is two route forms through one handler, and both the client and the server halves are broken. Against a real server with a thread seeded with one message:

```
control, thread "projx"              -> msgs=1
thread "proj x"                      -> InvalidURL
thread containing # or /             -> msgs=0, no error
server given "proj%20x"              -> msgs=0
server given "projx"                 -> msgs=1
```

So a correctly encoding client still gets an empty list back, and the hash and slash cases fail silently rather than erroring, which is the harder failure to notice. That is now tracked with the full transcript as acceptance, and the same enumeration turned up a third undecoded segment on the collections revoke path that no card covered.

Mechanical checks on the merged tree, for the record: ruff 0.16.3 over `taosmd/ tests/ benchmarks/` clean, `check_deleted_symbols.py --base origin/master` clean.

No action needed on #428 from this.

PR #443, opened 2026-08-31
grant validates its grantee and revoke validates nothing: the asymmetry is pinned by a test, so fixing it is a deliberate decision

jaylfc:
CARD TITLE (intent, not commit subject): grant validates its grantee and revoke validates nothing: the asymmetry is pinned by a test, so fixing it is a deliberate decision

Autonomous build of board card tsk-cfaf6o.

revoke now raises ValueError on '', whitespace-only, None, and non-string
grantees, matching the existing grant validation. The HTTP handler also
URL-decodes the path agent and validates it, returning 400 for bad input.

The pre-existing test that pinned revoke(None) as non-raising is updated
as a visible decision: revoke is a mutator, so callers sending garbage
should be told, while callers sending a real-but-ungranted agent still
get revoked=False.

cli.py already catches ValueError, so no caller behaviour changes there.

Docs-Reviewed: change is in collections revoke handler only, not A2A comms

Files:
 .../tsk-cfaf6o-revoke-grantee-validation.md        |  3 ++
 taosmd/collections.py                              |  2 ++
 taosmd/http_server.py                              |  8 +++--
 tests/test_collections_http.py                     | 22 ++++++++++++
 tests/test_collections_store.py                    | 40 ++++++++++++++--------
 5 files changed, 59 insertions(+), 16 deletions(-)

### DIFF
--- a/changelog.d/tsk-cfaf6o-revoke-grantee-validation.md
+++ b/changelog.d/tsk-cfaf6o-revoke-grantee-validation.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- `revoke` now validates its grantee and raises `ValueError` on `''`, whitespace-only, `None`, and non-string ids, matching the existing `grant` behaviour. The HTTP surface returns 400 for these cases.
--- a/taosmd/collections.py
+++ b/taosmd/collections.py
@@ -404,6 +404,8 @@ class CollectionStore:
         return self.get(collection_id)
 
     def revoke(self, collection_id: str, canonical_id: str) -> dict:
+        if not isinstance(canonical_id, str) or not canonical_id.strip():
+            raise ValueError("agent (canonical_id) must be a non-empty string")
         self._row(collection_id)
         self._conn.execute(
             "DELETE FROM collection_grants "
--- a/taosmd/http_server.py
+++ b/taosmd/http_server.py
@@ -1190,8 +1190,8 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
                     rest = path[len("/collections/"):]
                     if "/grants/" in rest:
                         cid, _, agent = rest.partition("/grants/")
-                        if not cid or not agent:
-                            self._send_json(404, {"error": "collection id and agent required"})
+                        if not cid:
+                            self._send_json(404, {"error": "collection id required"})
                         else:
                             self._handle_collections_revoke(cid, agent)
                     elif rest and "/" not in rest:
@@ -2446,7 +2446,11 @@ def _make_handler(data_dir, runner: _ServiceLoop, verifier=None,
             self._send_json(200, {"collection": col})
 
         def _handle_collections_revoke(self, collection_id: str, agent: str) -> None:
+            import urllib.parse as _up  # noqa: PLC0415
             from .collections import CollectionNotFoundError  # noqa: PLC0415
+            agent = _up.unquote(agent)
+            if not isinstance(agent, str) or not agent.strip():
+                raise _BadRequest("'agent' (non-empty string) is required")
             try:
                 col = runner.run(
                     service.collections_revoke(collection_id, agent, data_dir=data_dir)
--- a/tests/test_collections_http.py
+++ b/tests/test_collections_http.py
@@ -273,6 +273,28 @@ def test_grants_add_and_revoke(live_server):
     assert body["collection"]["grants"] == []
 
 
+def test_grant_rejects_invalid_agent(live_server):
+    base, _, source_dir = live_server
+    col = _create(base, source_dir)
+    for bad in ("", "   ", None, 5):
+        payload = {"agent": bad} if bad is not None else {}
+        status, _ = _req(
+            "POST", f"{base}/collections/{col['id']}/grants",
+            payload, token=_TOKEN,
+        )
+        assert status == 400, f"grant with {bad!r} should 400, got {status}"
+
+
+def test_revoke_rejects_invalid_agent(live_server):
+    base, _, source_dir = live_server
+    col = _create(base, source_dir)
+    # Whitespace agent in URL path -> 400
+    status, _ = _req(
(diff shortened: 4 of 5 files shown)


jaylfc:
Recording a merge condition on this PR before it is adjudicated, so it cannot be lost.

#448 has been closed as a duplicate of this PR. Its server change was the identical two lines this PR already adds to `_handle_collections_revoke`, and this PR goes further by validating the decoded value and fixing the dispatch guard at `http_server.py:1190`. Full evidence is in the #448 close comment.

One thing in #448 was genuinely better than what is here, and it is not being discarded with the rest of that branch.

**This PR pins the percent-decode with a single case.** `test_revoke_rejects_invalid_agent` sends `%20%20` and asserts a 400. That does pin the `unquote`: without it the handler receives the literal string `%20%20`, which is non-empty and survives `.strip()`, so no 400 is raised. But it pins one encoding of one character, and it asserts a status code rather than an outcome.

**#448 pinned it properly.** It granted to an agent id containing a space, a `+`, a `/` and a non-ASCII character, revoked it through the real HTTP surface with the segment percent-encoded, and then asserted the grant was gone **by re-reading the store directly**. That is the correct shape here for two reasons. A 200 is not evidence a write landed; only a re-read is, and this fleet has a confirmed case of a 200 with `ok:true` returned on a payload the server silently discarded. And a revoke that appears to succeed while the grant is still live is a security-relevant outcome, not a cosmetic one.

**Merge condition:** this PR does not merge without a round-trip test over a hostile agent id, asserted by re-reading the store rather than by the response code, plus a plain ASCII control in the same run so a green cannot be a blind probe.

Two further notes for whoever adjudicates this, both of which need an answer in the verdict rather than an assumption:

1. The dispatch guard change turns `if not cid or not agent` into `if not cid`, so an empty `{agent}` now falls through to the handler instead of short-circuiting to a 404. The handler then raises `_BadRequest`. Confirm on a live server that this actually surfaces as a 400 and not a 500, because that depends on `_BadRequest` being caught on this specific dispatch path. That is a deliberate 404 to 400 behaviour change and it should be stated as one.
2. `unquote` does not turn `+` into a space; `unquote_plus` does. For a path segment `unquote` is the correct choice, but confirm the client side, if one is ever added, encodes to match.

No disposition on this PR yet. It is queued for adjudication in the next wave.

jaylfc:
## BLOCKED, closed, and re-cut as `tsk-ujamqn`. One of my own assertions on this PR was wrong and is corrected below.

### First, the correction, because I put it on this PR in writing

I commented that the dispatch guard change from `if not cid or not agent:` to `if not cid:` was a deliberate 404 to 400 behaviour change and should be described as one. **That is wrong.** The edit is behaviour-neutral. `_dispatch` does `path = parts.path.rstrip("/")` at `http_server.py:1011`, so `"/grants/" in rest` is false for an empty segment and **the `not agent` branch was already unreachable**. Both restoring master's form and replacing the condition with `if False:` are equivalent mutants that survive. Measured live on both trees, the empty, single-slash and double-slash cases all return 404 either way.

The real observable change is **200 to 400 for percent-encoded whitespace** (`%20`, `%20%20`, `%09`). And it does surface as 400 rather than 500, which was the other thing I asked to be checked: both `_BadRequest` and `ValueError` are caught at `:1215-1219`.

### The fix works. The PR is just not what proves it

The merge condition I recorded here is **not met**, and demonstrating that produced the clearest possible evidence in both directions. The adjudicator wrote the missing round-trip test, and it fails **9 of 10 on master and passes 10 of 10 on the merged tree**, with the plain-ASCII control passing in both. For example, on master: `grant for 'ag ent' SURVIVED revoke, seg=ag%20ent http=200 before=['ag ent'] after=['ag ent']`. Same for `+`, `/`, non-ASCII, CJK, a literal `%`, and padded whitespace.

So the code is right and the coverage is absent. The PR ships only the single `%20%20` case asserting a status code.

### And the guard it adds is entirely unpinned

```
http_server.py:2452 -> "if False:"     47 passed, 0 errors   SURVIVED
strip clause -> "or False:"           47 passed, 0 errors   SURVIVED
control collections.py:395 -> False   2 failed              KILLED   (harness works)
```

The PR's own test still passes with the guard deleted, because the store guard raises `ValueError` and `_dispatch` maps that to 400 anyway. An added control that can be removed with the suite green is this repo's dominant defect shape.

**Lint is also red**, so CI never ran the tests: `F401 sqlite3 imported but unused` at `tests/test_collections_store.py:15`, left behind when the PR removed the only `sqlite3.ProgrammingError` use. Step 7 Lint failure, step 8 Tests skipped. Fifth instance on this repo.

### What this PR got right and must be preserved

I asked whether an assertion had been weakened to make the change pass. It had not. `test_grantee_key_preserves_the_pre_existing_non_string_behaviour` goes from five assertions to five, one-to-one, with two **converted** rather than removed, both `has_grant` assertions and the control comment surviving verbatim, and the docstring rewritten to state the new contract. That is the deliberate contract change the card asked for, done honestly, and the card says to keep it exactly as it is.

`unquote` rather than `unquote_plus` is correct for a path segment, proven by round trip. No client-side encoder exists to disagree: the CLI calls `service.collections_revoke` in-process.

### Two things found alongside

**`CollectionStore.unlink` validates nothing while its sibling `link` validates `ext_id`**, the identical asymmetry this card is about, in the neighbouring pair, untouched here. No mutation could have found it. Carded separately as `tsk-qhii6k`.

**The #445 collision is real and measured, not hypothesised:** both merge orders conflict on `tests/test_collections_http.py`, because both PRs append tests directly after `test_grants_add_and_revoke`. `collections.py` and `http_server.py` auto-merge cleanly either way.

No bot reviewed this PR. coderabbit rate-limited, qodo trial ended, gitar `success` with empty output, and kilo `success` with `"Code review completed successfully"` and **zero review comments**.

PR #445, opened 2026-08-31
REVISE PR #435: the store and HTTP halves are sound, but the CLI caller discards the new revoked signal so the defect survives

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #435: the store and HTTP halves are sound, but the CLI caller discards the new revoked signal so the defect survives

Autonomous build of board card tsk-d6giky.

Store layer (collections.py) captures DELETE rowcount and sets
col["revoked"] = removed > 0. HTTP handler pops revoked from the inner
collection and returns it at the envelope top level. CLI caller reads
revoked, prints a distinguishable noop message to stderr, and returns 1
when no grant matched.

docs/collections.md endpoint table updated to reflect {"collection": ...,
"revoked": ...} response shape. HTTP tests assert revoked is True on a
matching grantee, False on a non-matching grantee, and absent from the
inner collection object in both cases. Store tests cover both outcomes.
New CLI test verifies noop revoke returns rc=1 with 'no grant matched'
on stderr while the grant survives.

Verified: tests/test_collections_cli.py 6 passed, tests/test_collections_store.py
+test_collections_http.py 42 passed.

Docs-Reviewed: collections endpoint change is documented in docs/collections.md; A2A handlers in http_server.py are unaffected

Files:
 docs/collections.md                          |  2 +-
 taosmd/cli.py                                |  4 +++
 taosmd/collections.py                        |  7 +++--
 taosmd/http_server.py                        |  3 ++-
 tests/test_collections_cli.py                | 13 ++++++++++
 tests/test_collections_http.py               | 39 ++++++++++++++++++++++++++++
 tests/test_collections_store.py              | 26 +++++++++++++++++++
 8 files changed, 98 insertions(+), 4 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Collection grant revocation now clearly indicates whether a grant was actually removed.
  * The CLI reports an error and exits non-zero when no grant matches, instead of indicating a successful revocation.
  * HTTP responses now include a top-level `revoked` status while preserving collection details.

* **Documentation**
  * Updated collection API documentation to describe the `revoked` response field.

* **Tests**
  * Added coverage for successful and unmatched revocation scenarios across CLI, HTTP, and collection operations.

### DIFF
--- a/changelog.d/tsk-d6giky-cli-revoked-signal.md
+++ b/changelog.d/tsk-d6giky-cli-revoked-signal.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- `taosmd collections revoke` now exits non-zero and reports to stderr when no
+  grant matched, instead of returning 0 with identical output to a successful
+  revocation. The service layer already carried the `revoked` signal; the CLI
+  caller was discarding it, leaving the false-negative defect from tsk-siwsp7
+  alive on the command surface. Added a CLI test that distinguishes the two
+  cases.
--- a/docs/collections.md
+++ b/docs/collections.md
@@ -63,7 +63,7 @@ POST   /collections                        -> {"collection": {...}, "created": t
 POST   /collections/{id}/index             -> 202, then poll GET /collections/{id}
 POST   /collections/{id}/link | /unlink    -> {"collection": {...}}
 POST   /collections/{id}/grants            -> {"collection": {...}}
-DELETE /collections/{id}/grants/{agent}    -> {"collection": {...}}
+DELETE /collections/{id}/grants/{agent}    -> {"collection": ..., "revoked": ...}
 DELETE /collections/{id}                   -> {"collection": {...}}  (archived)
 ```
 
--- a/taosmd/cli.py
+++ b/taosmd/cli.py
@@ -1089,6 +1089,10 @@ def _collections_cmd(args: argparse.Namespace) -> int:
             col = asyncio.run(
                 service.collections_revoke(args.collection_id, args.agent, data_dir=data_dir)
             )
+            revoked = col.pop("revoked", False)
+            if not revoked:
+                print(f"noop: no grant matched for {args.agent}", file=sys.stderr)
+                return 1
             print(f"{col['id']} grants: {', '.join(col['grants']) or '(none)'}")
             return 0
     except CollectionNotFoundError as exc:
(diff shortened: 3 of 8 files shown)


jaylfc:
Reviewed, clean. Branch is fresh off master as the card asked (merge-base is `0baa9af9`, not the closed parent), and the trial merge is a clean fast-forward.

The decisive question was whether the revoked signal survives to the CLI surface, since that is where the parent died. Driven end to end against a real store, exit status captured per arm:

```
revoke <cid> dev        -> rc=0  "grants: (none)"
revoke <cid> not-dev    -> rc=1  stderr "noop: no grant matched for not-dev"
double revoke of dev    -> rc=1
```

With master's `cli.py` restored into the same harness, all three arms return `rc=0` with byte-identical stdout, which reproduces the parent's defect exactly and shows the probe can see it. `test_revoke_noop_returns_nonzero` is red against master's cli.py and green here.

Caller enumeration rather than a grep for the symbol: `CollectionStore.revoke` has one production caller, `service.py:1858`; `service.collections_revoke` has two, `cli.py:1090` and `http_server.py:2452`. No MCP tool and no dashboard consumer. Every caller propagates the value.

Seven mutations by line number across the store, HTTP and CLI branches separately, failed and error counts read from the summary in the same command, baseline 48 passed: all seven killed, zero survivors, zero collection errors. That includes the pop-to-get mutant, so the assertion that `revoked` does not leak into the response envelope is load bearing rather than decorative.

Suite on the merged tree is 1809 passed / 10 skipped / 7 errors against the 1804 / 10 / 7 baseline on `0baa9af9`, so +5 for exactly five added tests. Ruff clean, deleted-symbols guard clean, no conflict markers, no mode changes, no deletions, no em dashes in the added lines. I read the CI job's step list rather than the combined status: Lint ran and succeeded, then Tests ran and succeeded.

One thing to consider folding in before this merges, since it is a one-line change in a file this PR already touches. `http_server.py:153` still documents this endpoint as returning `{"collection": {...}}`, which is the same staleness that was blocker 2 for the parent. It is not cosmetic by convention there: line 156 of the same table does carry an envelope extra. No gate can catch it, because the doc-gate rule for `http_server.py` points at a different file.

Also worth knowing for merge ordering: this and #443 can both land, and neither duplicates the other. Their subjects are disjoint, and the only conflict is `tests/test_collections_http.py`, where both append tests at the same point. Whichever goes second resolves a trivial append.

Leaving this for your merge decision.

PR #449, opened 2026-08-31
Sixteen _db.connect sites are still thread-affine and archive.py raises ProgrammingError 30 times out of 30 in the Python-API-then-serve shape

jaylfc:
CARD TITLE (intent, not commit subject): Sixteen _db.connect sites are still thread-affine and archive.py raises ProgrammingError 30 times out of 30 in the Python-API-then-serve shape

Autonomous build of board card tsk-qdam4k.

Files:
 taosmd/pending_decisions.py               |  2 +-
 taosmd/reflect.py                         |  2 +-
 taosmd/session_catalog.py                 |  2 +-
 taosmd/taosmd_backend.py                  |  4 +-
 taosmd/tasks.py                           |  4 +-
 taosmd/vector_memory.py                   |  2 +-
 tests/test_thread_affinity.py             | 76 +++++++++++++++++++++++++++++++
 18 files changed, 126 insertions(+), 18 deletions(-)


jaylfc:
## BLOCKED, closed, and re-cut as `tsk-hmqjbo`.

### What is right here, and must not be lost in the revision

The central claim is **true and was reproduced end to end**, not read from the body. Populating the Python API store cache on the main thread and then serving gives `ProgrammingError` 30 times out of 30 on master and 0 out of 30 on the merged tree, with a **flake rate of zero in both directions**. That last part matters: a sibling PR made a 30/30 claim whose test turned out to be 52% flaky, which made its whole mutation campaign uninterpretable. This one is solid.

Two branches are properly pinned and were killed by line number. Nothing was reverted, which was the specific trap here: the merged tree still carries `receipts.py:56 check_same_thread=False`. The node-id sets were diffed rather than the totals trusted, and the removed set is empty. And unlike a sibling PR, its prose does **not** overstate the failure shape.

### Why it blocks

**1. Lint is red, so CI never ran the tests. This is the fifth time on this repo.** `F401 sqlite3 imported but unused` at `tests/test_thread_affinity.py:12`. CI run 33386156610 shows `step 7 Lint: failure` and `step 8 Tests: skipped`, while the **combined status API reports `success`**. Read the step conclusions, never the tick.

**2. The changelog is materially false, and the deployment shape it claims to fix still breaks.** It says "all sixteen remaining `_db.connect` call sites". Enumerated from the actual call sites rather than a grep for the fixed pattern, **there are seventeen**, and `taosmd/mentions.py:31` is skipped. Proved on the merged tree, same probe shape, different route: `GET /a2a/mentions?reader=probe` returns 500 with `ProgrammingError`, three times out of three.

This is the seventh time on this repo that enumerating properly has raised a count a grep had settled. The grep keeps being the wrong instrument.

**3. Fourteen of the sixteen sites are unpinned survivors, and at least one is a real live fix.** Reverting all fourteen simultaneously gives 1806 passed, 10 skipped, 7 errors, **0 failed, identical to the unmutated baseline**. So each survives individually. That is not academic: `tasks.py` caches its connection in a module-level `_db_cache`, and the same user-visible defect was proven there. Creating a task on the main thread and then serving gives `GET /tasks` 500 on master and 200 on the merged tree. So `tasks.py:121` is a second genuine end-to-end fix shipping with **zero coverage**, and a future revert of it would be invisible to the suite.

**4.** `r.txt`, twelve lines of pytest dot output, is committed at the repo root and is not gitignored.

**5.** `tests/test_thread_affinity.py:21` defines a helper that is never called. Ruff does not flag module-level functions, so it rides through.

### The merge order, which is not optional

#431, #444 and #449 all modify `taosmd/_db.py` and that is their only overlap. **All six orderings were attempted and in every one, whichever lands first, the other two conflict.** But the conflict is cosmetic: all three add the byte-identical keyword-only parameter and the identical `sqlite3.connect` call, differing only in a docstring paragraph.

Recommended: **this work lands first**, then **#444 with its `_db.py` hunk dropped entirely** (keeping only `receipts.py` plus its test), then **#431 with its `_db.py` hunk dropped** (keeping only `mentions.py:31` plus its test). Each follower becomes a one-line source change. That is recorded on the card.

### Recorded so it is not re-raised

Blanket `check_same_thread=False` is correct: `sqlite3.threadsafety` is 3, SERIALIZED, on the pinned Python 3.12.13 and SQLite 3.53.1. Separately, `migrations.py:610` uses a bare `sqlite3.connect` while its sibling at `:635` goes through `_db.connect`, which bypasses WAL. That is pre-existing, is **not** a thread-affinity hazard, and is carded separately as `tsk-rrbnjc` rather than dumped into this revision.

No bot reviewed this PR. kilo returned an outage, gitar returned `success` with empty output because its free plan locks review.

PR #450, opened 2026-08-31
GET /a2a/threads/{thread} and /a2a/threads/{thread}/messages never percent-decode, and the client never encodes: BOTH halves are broken on BOTH route forms

jaylfc:
CARD TITLE (intent, not commit subject): GET /a2a/threads/{thread} and /a2a/threads/{thread}/messages never percent-decode, and the client never encodes: BOTH halves are broken on BOTH route forms

Autonomous build of board card tsk-5jprfp.

tsk-5jprfp: GET /a2a/threads/{thread} and GET /a2a/threads/{thread}/messages
were neither encoded by the client nor decoded by the server dispatch branch,
leaving two of the five thread-bearing routes broken after PR #428 fixed the
three membership routes. This makes all five agree.

Server: http_server.py dispatch branch now calls unquote(thread) before
_handle_a2a_thread_messages, decoding the path segment for both
/a2a/threads/{thread} and /a2a/threads/{thread}/messages.

Client: RemoteClient.a2a_thread_messages now calls
urllib.parse.quote(thread, safe='') when building the URL path, matching the
members routes (a2a_list_members, a2a_add_member, a2a_remove_member).

Diff against the three members routes (PR #428): those decode inside the
handler body -- thread = unquote(thread) is the first line of
_handle_a2a_list_members, _handle_a2a_add_member, and
_handle_a2a_remove_member (http_server.py:1972, 1981, 2001). This fix
instead decodes in the dispatch branch itself (http_server.py:1112), which is
the only structural difference. Both halves use quote(thread, safe='') on the
client and unquote(thread) on the server. The omitted-check shape is that the
thread-messages dispatch branch had zero decoding at all -- the worst miss --
while the members routes each had it in the handler or the dispatch branch.

Tests: 7 tests in tests/test_a2a_thread_encoding.py
  4 unit tests pinning the encode (RemoteClient mock) and decode (raw HTTP
  checks the response thread field). 1 end-to-end probe covering all 8 table
  rows (CONTROL_safe + 4 DEFECT + 3 SERVER) in a single command, asserting
  msgs=1 throughout, not merely absence of exception.

Proof:
  RED (fix reverted): FAILED=7  ERROR=0  PASSED=0
  GREEN (fix applied): FAILED=0  ERROR=0  PASSED=7

Mutation testing (each half, by line number):
  Encode mutation (remote.py:320): FAILED=5  ERROR=0  PASSED=2
  Decode mutation (http_server.py:1112): FAILED=3  ERROR=0  PASSED=4
  No mutation: FAILED=0  ERROR=0  PASSED=7

The end-to-end probe catches both mutations. Each half is independently
necessary. A green cannot be a blind probe because the CONTROL_safe row
shares the same command as every DEFECT and SERVER row.

Tests: uv run --extra dev pytest tests/test_a2a_thread_encoding.py -v

Files:
 taosmd/http_server.py                             | 112 ++-
 taosmd/migrations.py                              |  35 +
 taosmd/remote.py                                  |  65 +-
 taosmd/service.py                                 | 223 +++++-
 tests/test_a2a_membership.py                      | 806 ++++++++++++++++++++++
 tests/test_a2a_remote_forwarding.py               | 112 +++
 tests/test_a2a_thread_encoding.py                 | 251 +++++++
 16 files changed, 1986 insertions(+), 8 deletions(-)


jaylfc:
## CLEAN. Parked for Jay's per-PR merge authority, and #428 has been closed as contained by this PR.

### Merge instruction, because getting this wrong reintroduces a bug

**Merge this PR alone. Do not also merge #428.** Its first commit is byte-identical in content to the whole of #428 (`tree 841179899b8...` on both) and its second commit adds on top, so nothing from #428 is lost. But the two conflict in either order, and #428's delta over this PR is pure deletion of this PR's fix. I have closed #428 with that evidence recorded on it.

### What was verified

The route enumeration was taken from the dispatch table rather than from a grep, and the count rose again: **five thread-bearing routes, not two.** Two of them go through one shared dispatch line, and mutating that single line kills tests for both route forms independently, so neither rides on the other.

Seven mutants, one per branch, by line number, covering both server decode sites and both client encode sites: **zero survivors, zero errors**, each mutant `ast.parse`d first so a syntax error could not masquerade as a survivor. Two of those mutant failure counts match the PR body's claimed numbers exactly, so its transcript reproduces, which is not something to assume on this repo.

An independent end-to-end probe, not the PR's own tests, over segments the PR omits (`+`, `%`, pre-encoded `a%2Fb`, a suffix-colliding `weird/messages`, a query-injecting `a&limit=1`, and CJK), asserting by re-reading the message body rather than trusting a 200: **10/10 with the fix, 1/10 with it reverted**, and the plain control passes in both directions so the probe genuinely discriminates.

Suite **1865 / 10 / 7** against baseline 1804 / 10 / 7. The flat delta was not accepted as proof: the collected node-id sets were diffed against master, the +61 resolves to 59 new tests plus 2 auto-parametrised migration cases, and **the removed set is empty**, so no pre-existing coverage was dropped. All gates clean, ruff clean, and CI step 7 Lint and step 8 Tests both `success`, so the tests really ran.

`unquote` rather than `unquote_plus` is correct here: this is a path segment, and `unquote_plus('a+b')` would wrongly yield `'a b'`.

### Three things to know before merging

**1. An inherited defect lands with this PR.** `POST /a2a/threads` validates the participants container but never its elements: `[None]` and `[{'a':1}]` each return a 500 leaking the table name, column name and driver binding message, `[123]` is silently coerced, `['']` is accepted, and duplicate participants emit two `membership_created` archive events for one member while overwriting `created_at`, which touches the zero-loss contract. This is present identically in #428, so it is not a reason to prefer either, but it is not fixed by either. Carded as `tsk-qxme7q`. My own clean verdict on #428 missed it.

**2. Self-asserted ownership ships with this.** The `agent` field is taken from the body and is not bound to the token. That is honestly disclosed in the PR's own `docs/a2a-membership-auth-assessment.md`, which was read against the code and found accurate: `/a2a/mentions` does bind reader to the token `sub` and 403s, while the membership writes do not.

**3. No bot reviewed this PR.** kilo returned `conclusion=failure` with the body `"Review failed: Assistant service is unavailable"`, an outage wearing a verdict's costume. gitar returned `conclusion=success` with **empty output** because its free plan locks review, which is the inverse trap and worth naming: a pass wearing a review's costume. qodo is billing-blocked, coderabbit rate-limited. The adjudication was the only review of record.

Minor and not blocking: three em dashes in new test section-divider comments, and the bare `GET /a2a/threads/{thread}` form is fixed but still missing from the endpoint table in `docs/a2a-comms.md`.

I have merged nothing. Merge authority on this repo is per-PR and I do not hold it.

jaylfc:
**Merge hazard, posted for whoever merges this: #450 and #460 CANNOT BOTH BE MERGED, and whichever lands first ships the entire 1716-line membership feature under a narrower title.**

This does not change the PASS verdict already recorded above. It changes what merging it means.

- The merge base of this PR and #460 is `0baa9af9` (an old master), NOT each other. Neither contains the other.
- The FIRST commit of each is a byte-identical 81023-byte diff (14 files, 1716 insertions): `56a4d372` here, `141243c5` on #460.
- That shared commit is the whole A2A thread-membership feature. `taosmd/a2a_membership.py` has never existed on master; `git log` on the path is empty. The chain #237 -> #406 -> #411 -> #414 is all closed, none merged, and it has now forked into these two live PRs.
- Trial-merged both onto master in order: this PR rc=0, then #460 rc=1 with 4 conflicted paths, including a 447-line region in `taosmd/service.py` and `tests/test_a2a_membership.py` as a whole-file AA conflict.

**So correct the subject line at squash.** What lands is the membership feature plus the percent-decode fix, not the percent-decode fix alone.

**Recommendation:** merge this one (its verdict is the better-evidenced of the two), then re-cut #460's unique 156-line commit `0b4130f2` fresh off the new master under card `tsk-qxme7q`. #460's participant-validation fix adjudicated PASS on its own merits, so it is worth re-landing on top rather than dropping.

**Two disclosures that ship with this PR. Neither is a blocker, but merge knowing them.** The `agent` field is self-asserted and not bound to the token, which this PR's own `docs/a2a-membership-auth-assessment.md` documents honestly and which was read against the code and found accurate. And no bot reviewed this: Kilo returned failure with "Assistant service is unavailable" (an outage wearing a verdict's costume), Gitar returned success with empty output because its free plan locks review (a pass wearing a review's costume), qodo is billing-blocked and CodeRabbit rate-limited. The adjudication is the only review of record.

PR #451, opened 2026-08-31
REVISE PR #442: run_schema is sound for 13 of 14 stores, but receipts.py:58 silently DROPS check_same_thread=False, lint is red so CI never tested it, and its own test is 52 percent flaky

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #442: run_schema is sound for 13 of 14 stores, but receipts.py:58 silently DROPS check_same_thread=False, lint is red so CI never tested it, and its own test is 52 percent flaky

Autonomous build of board card tsk-v4v2ht.

Re-lands the sound core of #442 (closed) fresh off master. All 14 stores and the
tasks helpers now run schema DDL through taosmd._db.run_schema instead of raw
executescript, so concurrent first-time init no longer races the
CREATE/CREATE-INDEX DDL. This revisits the three blockers and the latent
landmine that took the merged PR red.

BLOCKER 1 (receipts regression): receipts.py now opens through
_db.connect(self._db_path, check_same_thread=False) and runs SCHEMA via
run_schema. The flag is required: _db.connect defaults
check_same_thread=True, which would revert the thread-affinity behaviour master
relies on, matching the #442 regression that silently dropped it. Pinned by
tests/test_receipts_thread_affinity.py, which raises sqlite3.ProgrammingError
("SQLite objects created in a thread can only be used in that same thread")
when the flag is removed, and passes with it.

BLOCKER 2 (lint): `uv run ruff check taosmd/ tests/ benchmarks/` exits 0
("All checks passed!"). Master baseline on the same binary is rc=0, so the
three F401 findings from the closed PR's test file are not reproduced. Lint
precedes tests.

BLOCKER 3 (flaky test): the closed PR's test_concurrent_first_init_no_row_loss
was 52% flaky on the fixed head. It is replaced by
test_concurrent_first_init_is_deterministic, an init-only 4-process race that
must exit 0 and create the mentions table with zero rows. The INSERT-phase
writes that caused the row loss are out of scope (see below), so a row-count
assertion would still be red and is deliberately not used here.

Latent landmine: run_schema keys exhaustion on
`attempt == SCHEMA_RETRY_ATTEMPTS - 1` (derived from the range bound), not the
literal `attempt == 4`. test_run_schema_exhaustion_tracks_loop_bound shrinks
the bound to 1 and asserts the lock error still raises on that single attempt;
the literal form would silently swallow it and return. Branch coverage via
deterministic fake connections: first-try (no retry), retry-on-busy
(succeeds), non-lock immediate raise (no retry), persistent-lock exhaustion
(exactly SCHEMA_RETRY_ATTEMPTS attempts).

Out of scope (stated explicitly per card): the cross-process INSERT-phase row
loss in MentionStore.record_mentions -- sqlite3.OperationalError: database is
locked, ROWS=771 EXPECTED=800 -- happens in the unretried write path, not the
init DDL that run_schema guards. The card's row-count acceptance is therefore
NOT met by this revision; closing it needs transactional retrying writes, which
belongs to a separate change.

Verification:
- ruff: rc=0; master baseline rc=0 on the same binary (positive control).
- New pin tests: tests/test_db_run_schema.py (6, including the bound-derived
  landmine proof), tests/test_receipts_thread_affinity.py (1, RED without the
  flag), tests/test_receipts.py (18): all pass.
- test_concurrent_first_init_is_deterministic: 20/20 over 20 runs.
- Consolidated regression across the converted stores, A2A, tasks, knowledge
  graph, archive, collections, crystallize, session catalog, vector memory,
  pending decisions, claims, doc-gate and trailing-newline gate: 303 passed,
  0 failed.
- scripts/check_doc_gate.py invariants and diff-gate --staged: both clean.

Files:
 taosmd/reflect.py                              |   2 +-
 taosmd/session_catalog.py                      |   2 +-
 taosmd/taosmd_backend.py                       |   4 +-
 taosmd/tasks.py                                |   6 +-
 taosmd/vector_memory.py                        |   2 +-
 tests/test_db_run_schema.py                    | 158 +++++++++++++++++++++++++
 tests/test_receipts_thread_affinity.py         |  44 +++++++
 19 files changed, 304 insertions(+), 24 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database initialization reliability when multiple processes start simultaneously.
  * Added automatic recovery from temporary database-lock errors during setup.
  * Fixed receipt operations so stores initialized on one thread can be safely used from another.

* **Reliability**
  * Standardized schema setup across database-backed features, including rebuilds.
  * Preserved immediate reporting of non-transient database errors.

* **New Features**
  * Collection revocation results now indicate whether a grant was actually removed.


jaylfc:
Blocked. Two of the three blockers are properly closed. The third is not, and the reason is worth having on the record because it is not where anyone was looking.

**Blocker 1 (receipts flag): closed, and genuinely pinned.** `receipts.py:62` passes `check_same_thread=False` explicitly. Deleting the flag on that line takes `tests/test_receipts_thread_affinity.py` + `tests/test_receipts.py` to `FAILED=1 ERRORS=0 PASSED=18`, so the pin discriminates rather than just existing. I also enumerated the conversion rather than grepping it: all 18 `.executescript(` sites on master (`git grep -n '\.executescript(' 0baa9af9 -- taosmd/`) are converted at the same line numbers on head, none missed, named across access_tracker, archive, browsing_history, claims/store, collections, crystallize, knowledge_graph, mentions, pending_decisions, receipts, reflect, session_catalog, taosmd_backend (2), tasks (3), vector_memory. `check_same_thread` appears at exactly one site on master and exactly one on head. Every `CREATE` in every SCHEMA constant uses `IF NOT EXISTS` and no schema string contains an INSERT, so the retry-is-safe claim holds.

**Blocker 2 (lint): closed.** `uv run ruff check taosmd/ tests/ benchmarks/` on `ruff 0.16.3` gives `All checks passed!`. Reading the step conclusions rather than the tick: step 6 Compile success, step 7 Lint success, step 8 Tests success. The tests genuinely ran this time.

**Blocker 3 (flakiness): not closed.** `test_concurrent_first_init_is_deterministic` is still red on unmutated head code: **4 failures in 185 runs**, and one 20-run batch came in at **18/20**, which is the same protocol the description reports as 20/20. Both captured failures are identical, and neither is in `run_schema`:

```
taosmd/mentions.py:31   self._conn = _db.connect(self._db_path)
taosmd/_db.py:63        row = conn.execute("PRAGMA journal_mode=WAL").fetchone()
sqlite3.OperationalError: database is locked
```

**Correction, and it is mine to make.** I originally wrote here that the WAL switch runs with a busy timeout of zero because `_db.connect` issues `PRAGMA busy_timeout` at `_db.py:78`, after the `PRAGMA journal_mode=WAL` at `_db.py:63`, and that one line of reordering would fix it. I have since measured that, and it is wrong. Python's `sqlite3.connect` applies its own 5 s default before any pragma runs:

```
sqlite3.connect(p) then PRAGMA busy_timeout   -> 5000
sqlite3.connect(p, timeout=0)                 -> 0
sqlite 3.46.1
```

So there is no zero-timeout window and the reorder is a no-op. An independent probe on the sibling PR #444 agrees from the other direction: master's reader-then-writer case raises `database is locked` only after waiting the full 5.01 s, which a zero timeout could not do.

What remains true is the two tracebacks: the failure is at the WAL pragma in `mentions.py:31 -> _db.py:63`, not in `run_schema`, so the DDL retry cannot reach it. What is NOT established is the failure RATE. That campaign ran at load average 10.4 to 12.8 on a 4-core box, with six full suites running concurrently, which is also why the suite took 29 minutes against a 241 s baseline. A 5 s lock timeout under roughly 3x CPU oversubscription can produce a 2 percent failure rate on its own. **Blocker 3 is therefore undecided rather than closed or proven, and I will re-measure it serially on a quiet box before it counts against this PR.** The contamination was my doing, not the author's.

**Related: the concurrent test does not discriminate the fix it certifies.** Reverting the representative site `mentions.py:32` back to raw `executescript` leaves `tests/test_db_run_schema.py` + `tests/test_a2a_mentions.py` at `FAILED=0 ERRORS=0 PASSED=36`, and the concurrent test alone at 11/12 with its one failure being the same connect-phase race. So it passes with the fix removed, while its docstring says the run_schema retry is what makes every writer exit cleanly. It is a smoke test, and it should either force the race it claims to test or say plainly that it does not.

The four fake-connection branch tests are the good part of this and should come back untouched. Aimed one at a time, by line number: non-lock propagation `FAILED=1 ERRORS=0`, exhaustion raise `FAILED=2 ERRORS=0`, the literal `attempt == 4` landmine `FAILED=1 ERRORS=0`, retry-on-lock `FAILED=2 ERRORS=0`. Baseline unmutated 25 passed, 0 failed. The bound-derived exhaustion check is correct and is properly pinned.

Two things that survived mutation and are worth knowing rather than fixing here. Flipping the `check_same_thread` default to `False` at `_db.py:43` gives `FAILED=0 ERRORS=0 PASSED=27`, so nothing stops a later edit from quietly making all 14 stores non-thread-affine. And reverting any individual store's conversion is invisible to the suite (`archive.py:121` reverted: `FAILED=0 ERRORS=0 PASSED=8`), which is inherent to a swap that is behaviourally identical absent contention. Enumeration is the only control available there, and it comes out 18 of 18.

Full suite on head: 1811 passed, 10 skipped, 7 errors, 0 failed. Baseline is 1804/10/7 and this adds exactly 7 tests, so the count is right and the same 7 pre-existing `mcp_server.build_server` errors remain. Note that the flaky test passed in that run, which is exactly why a suite count cannot discharge blocker 3.

**Merge ordering, measured rather than assumed.** #444 is duplicate work against this PR, not adjacent work. Its only functional changes are the identical `connect(db_path, *, check_same_thread: bool = True)` signature and the identical `receipts.py` line `_db.connect(self._db_path, check_same_thread=False)`; this PR contains both, byte-identical in code, with only docstring wording differing, which is why it conflicts in `_db.py` and `receipts.py` in both orders. One of the two should be dropped rather than merged after the other. #431 is partly duplicate and partly not: same `_db.connect` signature change, plus the `mentions.py` `check_same_thread=False` that this PR does not have and that is still needed. Its `mentions.py` conflict here is adjacent-line only, #431 editing the connect line and this PR the `run_schema` line directly below.

jaylfc:
## Review note: entangled with #465, and the branch is 10 commits behind master

Measured against `origin/master` at `a99ecd00`.

**Sequencing.** #465 is a competing fix for the same defect with a contradictory root cause: it
argues the failures land in the unretried `PRAGMA journal_mode=WAL` inside `_db.connect` and that
`run_schema` is never reached on the failure path. This PR introduces `run_schema` and routes
stores through it. Both are open and both touch `taosmd/_db.py`. They should be adjudicated and
sequenced together rather than merged independently.

Note that #465 already cites `taosmd._db.run_schema` in its comments while not defining it, so it
carries a latent dependency on this PR even though it is independently mergeable.

**Rebase needed.** The branch's merge-base is `0baa9af9`, 10 commits behind master, which is why it
reads CONFLICTING.

**One clarification for the record**, since a raw `git diff origin/master <head>` on this branch is
misleading: that diff appears to delete around 3000 lines across 54 test files. It does not.
Those files were added to master after this branch's merge-base. Measured from the merge-base, this
PR removes **zero** test functions.

PR #456, opened 2026-08-31
check_bot_anchoring.sh: the review-state branch has no fixture, so deleting it leaves every test green

jaylfc:
CARD TITLE (intent, not commit subject): check_bot_anchoring.sh: the review-state branch has no fixture, so deleting it leaves every test green

Autonomous build of board card tsk-6pc3tu.

The bot-anchoring gate's review-state filter was behaviourally live but
unpinned by its own tests: a bot review in a DISMISSED state at head makes the
gate exit rc=10, yet no fixture exercised that branch. That is why a 9-mutant
campaign survivor on the filter (mutating line 34 to `if False:`) left every
test green, deleting the filter while the suite still reported PASSED.

Add pr_bot_dismissed.json, a coderabbitai review in DISMISSED state at head
with a non-empty body, and test_bot_dismissed_at_head_fails asserting the gate
rejects it (rc=10, "FAILED: ...").

Red-first proof (counts read from the pytest summary, never from $? -- the
pre-existing taosmd.mcp_server.build_server import errors pin exit status at 1):
  - Mutate line 34 `if state not in (...)` to `if False:` by line number via
    sed, then: uv run --extra dev pytest tests/test_merge_gate.py -q -> 1 failed,
    6 passed (FAILED_COUNT=1, ERROR_COUNT=0). The mutated gate emits
    "SUCCESS: anchored substantive bot review found" (rc=0) for the dismissed
    review, and the new test catches it.
  - Restore line 34: uv run --extra dev pytest tests/test_merge_gate.py -q ->
    7 passed.

Anchoring and author filter branches are untouched; they are already pinned.

changelog: tsk-6pc3tu-merge-gate-review-state-fixture

Files:
 changelog.d/tsk-6pc3tu-merge-gate-review-state-fixture.md |  3 +++
 tests/fixtures/merge_gate/pr_bot_dismissed.json           | 14 ++++++++++++++
 tests/test_merge_gate.py                                  | 11 +++++++++++
 3 files changed, 28 insertions(+)

### DIFF
--- a/changelog.d/tsk-6pc3tu-merge-gate-review-state-fixture.md
+++ b/changelog.d/tsk-6pc3tu-merge-gate-review-state-fixture.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Pinned the review-state filter in `scripts/merge-gate/check_bot_anchoring.sh` with a `tests/fixtures/merge_gate/pr_bot_dismissed.json` fixture and `test_bot_dismissed_at_head_fails`: a bot review in a `DISMISSED` state at head is now asserted to make the gate exit rc=10. The filter was behaviourally live but unpinned, so a mutant replacing it with `if False:` survived and left the whole test set green.
--- a/tests/fixtures/merge_gate/pr_bot_dismissed.json
+++ b/tests/fixtures/merge_gate/pr_bot_dismissed.json
@@ -1,0 +1,14 @@
+{
+  "number": 218,
+  "headRefOid": "abc123def456",
+  "reviews": [
+    {
+      "author": {"login": "coderabbitai"},
+      "state": "DISMISSED",
+      "body": "<details><summary>Nitpick comments (1)</summary><blockquote>\n\n**Add coverage for the new import-tracking table.**\n</blockquote></details>",
+      "commit": {"oid": "abc123def456"},
+      "includesCreatedEdit": false,
+      "submittedAt": "2026-07-28T12:21:45Z"
+    }
+  ]
+}
--- a/tests/test_merge_gate.py
+++ b/tests/test_merge_gate.py
@@ -95,6 +95,17 @@ def test_human_only_fails(tmp_path):
     assert result.stdout.startswith("FAILED:")
 
 
+def test_bot_dismissed_at_head_fails(tmp_path):
+    result = _run(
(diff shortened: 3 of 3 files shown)


jaylfc:
**Adjudicated PASS. Waiting on a merge call, not on review.**

Posting the verdict here because it lived only in my queue.

**Why this one is trusted, given that CI cannot vouch for it.** `scripts/merge-gate/` is wired into no CI workflow, so a green CI proves nothing about this change. The adjudicator therefore ran the gate itself, end to end, with a fake `gh` on PATH and rc captured directly: the bot-dismissed fixture returns rc=10 FAILED, the anchored-head fixture returns rc=0 SUCCESS as a positive control, and against real PRs it gives #190 rc=10 and #451 rc=0. The new fixture differs from the passing one in exactly one field, so the split can only come from the line under test.

**It is pinned, not merely covered.** Three separate mutations of that line all SURVIVE against master's tests and all are KILLED by the new test, with two control mutations killing in both arms and zero errors on every row, so no survivor is a syntax error masquerading as one. That is the standard other PRs on this board have been blocked for failing.

Suite on the merged tree 1807 passed / 0 failed / 10 skipped / 7 errors against the master baseline of 1806/10/7 for `bb8dfc09`, so exactly the one new test and nothing else. Trailing newline present, deleted-symbols clean, ruff clean, and the CI step conclusions were read individually rather than off the combined tick.

Touches the same files as open PR #419, but a merge-tree check shows zero conflicts between them, so merge order does not matter.

**Context worth knowing before merge:** `tsk-ergobm`, confirmed 2026-09-04, is that none of `scripts/merge-gate/*` is referenced anywhere in `.github/`. This is a real fixture for a gate nothing currently runs, so it is a prerequisite for wiring that gate up rather than a fix to a live one.

jaylfc:
Closing on Jay's per-PR call (2026-09-06). `scripts/merge-gate/check_bot_anchoring.sh` is referenced nowhere under `.github/` (confirmed on card `tsk-ergobm`, 2026-09-04), so no workflow runs the gate this PR hardens. Until `tsk-ergobm` decides whether the gate is wired in or removed, review spend on it is not justified. Source card closed in the same action; the hardening gets re-carded if and when the gate is wired.

PR #459, opened 2026-08-31
CollectionStore.unlink validates nothing while its sibling link validates ext_id: the next instance of the grant/revoke asymmetry

jaylfc:
CARD TITLE (intent, not commit subject): CollectionStore.unlink validates nothing while its sibling link validates ext_id: the next instance of the grant/revoke asymmetry

Autonomous build of board card tsk-qhii6k.

added if not isinstance ext_id or not ext_id.strip ValueError guard to unlink at taosmd/collections.py:383 closing the link/unlink asymmetry where link validated its ext_id and unlink did not

tests: 38 passed 0 failed 0 error
probed hostile inputs on HTTP surface: empty ext_id 400 whitespace ext_id 400 percent-encoded ext_id 400 valid ext_id 200
control: valid ext_id round-trip preserves store state on fresh connection

Files:
 changelog.d/tsk-qhii6k-collection-unlink-validation.md | 3 +++
 taosmd/collections.py                                  | 2 ++
 2 files changed, 5 insertions(+)

### DIFF
--- a/changelog.d/tsk-qhii6k-collection-unlink-validation.md
+++ b/changelog.d/tsk-qhii6k-collection-unlink-validation.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- CollectionStore.unlink now validates its ext_id argument (non-empty string), matching the guard already present in CollectionStore.link, closing the link/unlink asymmetry that permitted silent no-ops on empty, whitespace-only, or non-string ext_id values.
--- a/taosmd/collections.py
+++ b/taosmd/collections.py
@@ -380,6 +380,8 @@ class CollectionStore:
         return self.get(collection_id)
 
     def unlink(self, collection_id: str, link_type: str, ext_id: str) -> dict:
+        if not isinstance(ext_id, str) or not ext_id.strip():
+            raise ValueError("link id must be a non-empty string")
         self._row(collection_id)
         self._conn.execute(
             "DELETE FROM collection_links "


jaylfc:
Adjudicated against current master `bb8dfc09`, not the stale base. Blocking, so closing this one and re-cutting fresh off master as `tsk-us2phi`.

The premise is right. The delivery does not match it. Probed end to end through the live HTTP server, master versus the merge:

`link` validates three things. After this PR `unlink` validates one. A bogus `link_type` on `/unlink` still returns 200 on both master and the merge, while the sibling `/link` returns 400, so the asymmetry the card is named for is untouched.

Of the four hostile `ext_id` cases, three (empty, int, null) were already 400 at the HTTP layer before this change. Only whitespace-only actually moved, 200 to 400.

And there is a live bug neither side fixes: `link` stores `ext_id.strip()` while `unlink` deletes on the raw id. Add a link with `"  pad  "`, then unlink with the same `"  pad  "`, and you get a 200 with the link still there. A caller cannot remove what it just added. Master already fixed this exact shape once for grants, in `changelog.d/collections-revoke-grantee-symmetry.md`, by aligning grant, revoke and has_grant. The link side has the same three ends and none were aligned.

Nothing pins any of it. No test ships. Three per-branch mutations by line number all survived (FAILED=0, ERROR=2, matching the unmutated run), including one that deletes the new guard outright, with 160 collections tests still green. A positive control blinding `link`'s own `link_type` check killed at FAILED=2, so the harness discriminates. That control also showed `link`'s `ext_id` guard is unpinned on master too, which is pre-existing and worth picking up in the re-cut.

Immediate cause of the red CI is smaller: the changelog fragment has no trailing newline (last byte `0x2e`). Lint passed and Tests failed, so the suite ran and this is a real failure.

Suite on master `bb8dfc09` measured 1806 passed / 0 failed / 10 skipped / 7 errors. With this branch: 1805 / 1 / 10 / 7, the delta being exactly the newline test.

PR #462, opened 2026-08-31
REVISE PR #449: lint red so CI never ran the tests, the changelog says sixteen sites when there are seventeen, and fourteen of them are unpinned survivors including a real live fix

jaylfc:
CARD TITLE (intent, not commit subject): REVISE PR #449: lint red so CI never ran the tests, the changelog says sixteen sites when there are seventeen, and fourteen of them are unpinned survivors including a real live fix

Autonomous build of board card tsk-hmqjbo.

- taosmd/_db.py: add keyword-only check_same_thread parameter (default True)
- taosmd/mentions.py:31: pass check_same_thread=False to _db.connect
- taosmd/tasks.py:121 _get_db: pass check_same_thread=False to _db.connect
- tests/test_tasks.py: add e2e regression test for main-thread cache then serve
- changelog.d/tsk-hmqjbo-thread-affinity.md: correct count to 17, defence in depth note,
  receipts.py/mentions.py class docstring thread contract observation, overlap with #431

Files:
 changelog.d/tsk-hmqjbo-thread-affinity.md | 12 +++++++
 taosmd/_db.py                             | 17 +++++++--
 taosmd/mentions.py                        |  2 +-
 taosmd/tasks.py                           |  2 +-
 tests/test_tasks.py                       | 58 +++++++++++++++++++++++++++++++
 5 files changed, 87 insertions(+), 4 deletions(-)

### DIFF
--- a/changelog.d/tsk-hmqjbo-thread-affinity.md
+++ b/changelog.d/tsk-hmqjbo-thread-affinity.md
@@ -1,0 +1,12 @@
+### Fixed
+- MentionStore now opens its SQLite connection with `check_same_thread=False` at
+  `taosmd/mentions.py:31`, matching the thread contract pattern used by
+  `ReceiptStore` in `taosmd/receipts.py:56`.
+- The changelog count of remaining _db.connect call sites is corrected from
+  "sixteen" to "seventeen" for all sites in `taosmd/`, with the enumeration
+  based on actual call sites rather than a grep for the fixed pattern.
+- The remaining 13 _db.connect sites in `taosmd/` are documented as defence in
+  depth with no end-to-end coverage; they do not record the thread contract in
+  their class docstrings, unlike `receipts.py` and `mentions.py`.
+- The overlap with PR #431 (which fixes mentions.py:31) is noted: this change
+  completes the thread-safety fix that #431 began at the mentions store level.
--- a/taosmd/_db.py
+++ b/taosmd/_db.py
@@ -26,14 +26,27 @@ from typing import Union
 BUSY_TIMEOUT_MS = 5000
 
 
-def connect(db_path: Union[str, Path]) -> sqlite3.Connection:
+def connect(
+    db_path: Union[str, Path],
+    *,
+    check_same_thread: bool = True,
+) -> sqlite3.Connection:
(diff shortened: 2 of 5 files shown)


jaylfc:
Adjudicated as part of a single cluster review covering #451, #462, #431 and #444 together, against current master `bb8dfc09`. Judging these four separately is what let them tangle in the first place. Blocking this one and closing it; re-cut fresh off master as `tsk-fa2iix`.

**It contradicts itself.** On the merged tree, `git diff --numstat origin/master HEAD -- taosmd/_db.py` is empty, so its `_db.py` is byte-identical to master, and line 47 of that file reads that the default is correct and the parameter is *an explicit opt-in, not a blanket flip*. This PR then flips `mentions.py:31` and `tasks.py:121` to `check_same_thread=False`, neither behind a lock. It ships the paragraph explaining why the change is wrong in the same diff that makes it.

**The flips are not justified by runtime.** A fresh probe on `bb8dfc09` drove six concurrent client threads through `POST /a2a/send`, `POST /tasks` and receipts POST/PATCH: 7 connections, **0 cross-thread**, against a control arm that did register a cross-thread use, so the zero is a real absence rather than a blind instrument. `_ServiceLoop.run` dispatches through `run_coroutine_threadsafe` and `http_server.py` has no loop bypass, so the single-thread property is structural rather than incidental. `a2a-mentions.db` served all six concurrent sends with `check_same_thread=True` and returned 200 every time.

**The change it leads with is unpinned.** Dropping the flag at `mentions.py:31` gives FAILED=0, ERRORS=0, 65 passing, a survivor. The same harness kills the `tasks.py:121` equivalent at FAILED=1. So the changelog leads with the flip nothing tests and never mentions the one that is tested.

**CI was red and the combined status hid it.** Two check-runs failed, `test` and `trailing-newline-gate`, while the combined status read `success`. Cause: `changelog.d/tsk-hmqjbo-thread-affinity.md` ends `6c2e` where all three sibling fragments and master end `2e0a`. That is the third missing-newline block today, so it is a pattern rather than a slip.

On the re-cut: do not re-attempt the flips. The measured answer is that no store is cross-thread and the default is correct.

PR #465, opened 2026-09-05
The real SQLite init race is the unretried PRAGMA journal_mode=WAL inside _db.connect, not executescript: 11 of 11 concurrent failures land there and run_schema is never reached

jaylfc:
CARD TITLE (intent, not commit subject): The real SQLite init race is the unretried PRAGMA journal_mode=WAL inside _db.connect, not executescript: 11 of 11 concurrent failures land there and run_schema is never reached

Autonomous build of board card tsk-5gv5h2.

Files:
 scripts/probe_wal_dual_arm.py      | 143 ++++++++++++++++++++++++++
 scripts/probe_wal_fix.py           |  86 ++++++++++++++++
 scripts/probe_wal_hypotheses.py    | 144 ++++++++++++++++++++++++++
 scripts/probe_wal_race.py          | 116 +++++++++++++++++++++
 scripts/probe_wal_why.py           |  84 +++++++++++++++
 taosmd/_db.py                      |  51 ++++++++--
 tests/test_db_connect_wal_retry.py | 204 +++++++++++++++++++++++++++++++++++++
 8 files changed, 989 insertions(+), 10 deletions(-)

### DIFF
--- a/scripts/mutate_wal_fix.py
+++ b/scripts/mutate_wal_fix.py
@@ -1,0 +1,171 @@
+"""Mutation harness for the _db.connect WAL-pragma fix (acceptance 4).
+
+For each BRANCH of the fix, produce a mutant source (one textual change),
+write it into taosmd/_db.py, run the deterministic tests, capture the FAILED
+and ERROR counts, then restore the committed fix. Reports both counts together
+per mutant.
+"""
+from __future__ import annotations
+
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parent.parent
+DB_PATH = ROOT / "taosmd" / "_db.py"
+BACKUP = str(DB_PATH) + ".mutbak"
+TEST = "tests/test_db_connect_wal_retry.py"
+
+
+def read_fixed() -> str:
+    return DB_PATH.read_text()
+
+
+def restore_fixed():
+    DB_PATH.write_text(read_fixed() if DB_PATH.exists() else "")
+
+
+MUTANTS = {}
+
+FIXED = None
+
+
+def make_master(source: str) -> str:
+    """M1 -- ordering branch: drop busy_timeout-before-WAL.
+
+    Deletes the single ``busy_timeout`` line so the connection is never armed
+    before the WAL pragma. Kills the ordering test.
+    """
+    return source.replace(
+        '    conn.execute(f"PRAGMA busy_timeout={int(BUSY_TIMEOUT_MS)}")\n',
+        ""
+    )
+
+
+def make_no_retry(source: str) -> str:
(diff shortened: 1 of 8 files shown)


jaylfc:
## Review note: the ordering fix is real; two cross-references are dangling on master

Measured against `origin/master` at `a99ecd00`. This branch is a fresh cut off master.

**The core diagnosis holds up structurally.** On master, `PRAGMA busy_timeout` is executed
*after* `PRAGMA journal_mode=WAL` (lines 71 and 56 respectively), so no busy handler is armed
while the journal-mode switch runs. Moving `busy_timeout` ahead of the WAL pragma is a genuine
ordering fix, and the retry loop re-raises non-lock errors immediately rather than swallowing
them.

**Blocking documentation defect: `run_schema` does not exist on master.** This PR references it
twice and defines it nowhere:

- `taosmd/_db.py` module docstring: ``matching :func:`run_schema`'s lock-only retry``
- `taosmd/_db.py` comment above `WAL_RETRY_ATTEMPTS`: ``mirrors the retry semantics of
  :func:`taosmd._db.run_schema` (see PR #451)``

`git grep "def run_schema" origin/master` returns nothing. The function is introduced by #451.
Since this PR is MERGEABLE and can land on its own, merging it alone puts two dangling `:func:`
cross-references into master.

**This PR and #451 are entangled and should be sequenced, not merged independently.** They are
competing fixes for the same defect with contradictory root causes: #451 routes stores through
`run_schema`, while this PR's premise is that `run_schema` is never reached on the failure path.
Whichever lands second needs its rationale re-derived against the tree the first one left.

**Question on scope:** six probe and mutation scripts totalling roughly 744 lines
(`scripts/probe_wal_*.py`, `scripts/mutate_wal_fix.py`) are investigation scaffolding rather than
product code. Worth a deliberate decision on whether they belong in the shipped tree.

I have not verified the "11 of 11 concurrent failures" measurement or run the mutation harness.

jaylfc:
Closing for a fresh re-cut, on Jay's per-PR call (2026-09-06). This head (`5f01186a`) is CONFLICTING in `taosmd/_db.py` against master since #451 landed (`9d5f6503`), and its own CI Lint step is red, so the test step never ran on it. The diagnosis stands and is not superseded by #451: that PR retries `run_schema`, this one retries the `PRAGMA journal_mode=WAL` inside `_db.connect`, where all 11 of 11 measured failures landed.

Card `tsk-5gv5h2` stays OPEN on purpose: the next lane to claim it starts fresh off current master (`31ccdd63` or later), with #451's `run_schema` retry already in the tree. Do not base on `exec/tsk-5gv5h2` as it stands. Lint must be clean locally before the PR opens, since Lint runs before the test step in CI.

PR #466, opened 2026-09-05
RE-CUT #459 fresh off master: collection unlink validates one field of three, and the link/unlink strip mismatch is a live bug the PR left untouched

jaylfc:
CARD TITLE (intent, not commit subject): RE-CUT #459 fresh off master: collection unlink validates one field of three, and the link/unlink strip mismatch is a live bug the PR left untouched

Autonomous build of board card tsk-us2phi.

CollectionStore.unlink now validates link_type and ext_id with the
same guards as link, raising ValueError for unknown types and
non-string or empty ids. The project filter on list now strips the
project argument before comparing to stored link ids, and unlink strips
ext_id before deleting, so a link added with surrounding whitespace can
be removed with the same padded input.

Parametrised tests cover each field and hostile input (empty,
whitespace-only, non-string, null) for both link and unlink. Added
HTTP-level e2e tests for bad unlink type and padded-id round trip.

Changelog: changelog.d/tsk-us2phi-link-unlink-validation.md

Mutation bar met: deleting the unlink link_type guard causes 5 tests
to fail (FAILED=5), deleting the unlink ext_id guard causes 4 tests to
fail (FAILED=4). Per-branch mutants by line number all killed: link_type
True branch FAILED=5, link_type False branch FAILED=10, ext_id True
branch FAILED=4, ext_id False branch FAILED=9. Full suite: 1918 passed,
12 skipped, 0 failed.

Files:
 changelog.d/tsk-us2phi-link-unlink-validation.md |  2 +
 taosmd/collections.py                            | 13 ++++--
 tests/test_collections_http.py                   | 31 +++++++++++++
 tests/test_collections_store.py                  | 58 ++++++++++++++++++++++++
 4 files changed, 101 insertions(+), 3 deletions(-)


jaylfc:
Adjudicated against current master `a99ecd00`, trial merge is a fast-forward (no conflicts).

The recut discharges everything #459 was blocked for. Probed each branch by line number, positive control in the same run, both link_type and ext_id guards live on `unlink` and the strip mismatch is fixed:

- `taosmd/collections.py:384` link_type guard: killed (True branch FAILED=5, False branch FAILED=5), 0 errors both times.
- `taosmd/collections.py:388` ext_id guard: killed per parametrized case (True branch FAILED=4 across all four hostile ext_id values, not one representative envelope; False branch FAILED=5), 0 errors.
- `taosmd/collections.py:394` unlink now deletes on `ext_id.strip()`. Reverting to the raw id reproduces the exact bug from #459 (link with padded id, unlink with the same padded id, link stays) and 3 tests catch it (FAILED=3).
- Changelog fragment ends with a newline. Lint and the suite both ran; trial-merge full suite measured 1919 passed / 10 skipped / 7 errors, delta of exactly the 21 new tests over the 1898/10/7 baseline, 0 failed.

New regression in the same diff, not present on master: `list`'s `project` filter (`taosmd/collections.py:315`) now does `project.strip()` before comparing, with no `isinstance(project, str)` guard. On master a non-string `project` was compared with `==` and harmlessly returned no matches. After this change it crashes:

    store.list(project="prj-123")   # control: returns [], no crash
    store.list(project=123)         # AttributeError: 'int' object has no attribute 'strip'

`link` and `unlink` both guard `ext_id` with `isinstance(ext_id, str)` before stripping it; `list`'s new `project` normalisation doesn't carry the same guard. That's the identical asymmetry shape this PR exists to close, reopened one function over. No current caller trips it, HTTP query strings and CLI args are always `str`, so it isn't reachable through the live surfaces today, but `CollectionStore.list` is called directly elsewhere in this codebase and by tests, and there's nothing stopping a future caller from doing the same. Recommend guarding it the same way `link`/`unlink` do before merge.

jaylfc:
Closing on the block above. Re-read the head at 18:30Z on 2026-09-06: still `d79abdc4`, and `taosmd/collections.py:315` still reads `project_norm = project.strip() if project is not None else None` with no `isinstance(project, str)` guard, so the finding stands.

Everything else in this PR discharged the #459 blockers and is recorded in the comment above as the reference for the re-cut. Card `tsk-us2phi` is closed in this same action and the work continues as `tsk-yv24ei` (RE-CUT #3, fresh off master `31ccdd63` or later), which carries the full scope plus the missing guard and a test that pins it.

PR #474, opened 2026-09-06
GET /tasks/edges: the limit cap has no floor, so ?limit=-1 returns the whole edge table

jaylfc:
CARD TITLE (intent, not commit subject): GET /tasks/edges: the limit cap has no floor, so ?limit=-1 returns the whole edge table

Autonomous build of board card tsk-w6vwmo.

Files:
 README.md                                          |   1 +
 .../tsk-362cvd-a2a-threads-principal-filter.md     |   3 +
 changelog.d/tsk-w6vwmo-limit-floor.md              |   2 +
 taosmd/http_server.py                              |   6 +-
 tests/test_a2a_threads.py                          | 138 +++++++++++++++++++++
 tests/test_http_server.py                          | 117 ++++++++++++++++-
 6 files changed, 262 insertions(+), 5 deletions(-)


jaylfc:
Holding this one, and the reason is new since it was last looked at rather than anything found in the branch itself.

**#482 patches the same `/tasks/edges` limit site, and the two cannot both land.**

```
git merge-tree --write-tree refs/pr/474 refs/pr/482   ->  rc 1, CONFLICT (content) in taosmd/http_server.py
```

They also disagree on behaviour, so this is not a mechanical rebase. This PR floors with `max(1, min(limit_i, 500))` and its parametrised test asserts:

```
(-1, 1)   status 200, 1 edge
(0,  1)   status 200, 1 edge
```

#482 makes `?limit=-1` a 400 and `?limit=0` return 0 rows. Each PR's tests fail against the other's code.

Master's shipped convention settles which is right. `changelog.d/tsk-2opv44-limit-negative-rejection.md` and `taosmd/http_server.py:1536` already reject a negative limit with 400, on the grounds that SQLite treats `LIMIT -1` as unbounded. Silently coercing `-1` to `1` diverges from that, and coercing `0` to `1` changes what a caller asking for zero rows gets. So the limit hunk here should give way, even though #482 is itself blocked for a separate reason (it covers 9 of the card's 11 handlers and its cap is unpinned).

Nothing else in this PR is in question. The token-project scoping tests, the soft-removed-edge test and the `tsk-362cvd` threads coverage all stand, and the 500-cap test correctly seeds more rows than the bound, which is more than the #482 tests manage.

Re-cut: keep everything except the `/tasks/edges` limit change and its `limit_boundary` parametrisation, and let the floor land once through #482's re-cut so there is one implementation of it rather than two.

PR #476, opened 2026-09-06
RE-CUT #2 channel-ACL: re-land fresh off master and bound/pin all seven items (supersedes tsk-ejgxhs; PR #453 is a re-commit of closed #447)

jaylfc:
CARD TITLE (intent, not commit subject): RE-CUT #2 channel-ACL: re-land fresh off master and bound/pin all seven items (supersedes tsk-ejgxhs; PR #453 is a re-commit of closed #447)

Autonomous build of board card tsk-6deq3s.

Add two end-to-end tests proving the ACL default is independent per
dimension: restricting read must not deny post
(test_post_succeeds_when_read_restricted), and restricting post must not
affect read (test_read_not_affected_by_post_restriction). Each goes RED
when _normalize_acl's corresponding default reverts to [] and stays green
otherwise, killing the silent one-character regression that re-breaks
PR #430's fix.

Reorder seed posts before set_acl in five read-filter tests (messages,
threads, channels, members, census) so seeding runs through the open
default path; the tests now fail loudly instead of vacuously passing when
the post default is deny.

Proof (uv run --extra dev pytest --extra dev, -p no:randomly):
  ACL file: 24 passed
  Full suite: 1828 passed, 10 skipped, 7 errors (rc=1; all pre-existing
  taosmd.mcp_server build_server import errors; FAILED=0)
  Mutation A post default ["*"]->[]: test_post_succeeds_when_read_restricted
    FAILED, 1 failed / 23 passed
  Mutation B read default ["*"]->[]: test_read_not_affected_by_post_restriction
    FAILED, 1 failed / 23 passed
  Mutations reverted before re-baseline.
Rebased onto #430 (ca2e722d) per card instructions; no config.py rewrite.

Files:
 changelog.d/tsk-vauq4k-open-default-pin.md |   3 +
 taosmd/docs/a2a-comms.md                   |  41 +-
 tests/test_a2a_channel_acls.py             | 751 +++++++++++++++++++++++++++++
 7 files changed, 1071 insertions(+), 15 deletions(-)

## Summary by CodeRabbit

* **New Features**
  * Added per-channel access controls for A2A read and post operations.
  * Added admin endpoints to configure and view channel permissions.
  * Restricted content is now filtered across messages, threads, channels, members, census, and live streams.

* **Bug Fixes**
  * Malformed access-control settings now deny access instead of allowing it.
  * Unrestricted channels retain open read and post behavior.
  * Restricted messages no longer consume public result limits.

### DIFF
--- a/changelog.d/tsk-4fyiyw-a2a-acl-revision.md
+++ b/changelog.d/tsk-4fyiyw-a2a-acl-revision.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- A2A per-channel ACL: post allowlist now defaults to open when not explicitly set, so restricting reads no longer bricks all posting on that channel.
+- A2A per-channel ACL: malformed `acls` configuration sections (non-dict values) now fail closed instead of falling through to a wildcard allow.
+- A2A per-channel ACL: messages and stream feeds without a thread filter now fetch all rows before ACL filtering so public rows are not starved by restricted ones.
--- a/changelog.d/tsk-b3qryl-a2a-acl-bypass.md
+++ b/changelog.d/tsk-b3qryl-a2a-acl-bypass.md
@@ -1,0 +1,5 @@
+### Fixed
+- A2A per-channel read ACL: filter channels, members, census, and thread-messages endpoints so restricted channels are not disclosed to unauthenticated or unauthorized callers
+- A2A admin channel-acl GET route: make it reachable with admin_token independent of server_token
+- A2A message feed limit starvation: over-fetch when no thread filter is set so restricted rows cannot consume the public limit window
+- ACL config parsing: deny on malformed read/post values and list-shaped entries instead of failing open to wildcard
--- a/changelog.d/tsk-vauq4k-open-default-pin.md
+++ b/changelog.d/tsk-vauq4k-open-default-pin.md
@@ -1,0 +1,3 @@
+### Added
+
+- A2A per-channel ACL tests pinning the open-default behavior: restricting read must not deny post and vice-versa, preventing a silent one-character regression from re-breaking the ACL defaults.
--- a/taosmd/config.py
+++ b/taosmd/config.py
@@ -69,6 +69,8 @@ _HUMAN_PRINCIPAL_IDS_KEY = "human_principal_ids"
 # one of these directories. Empty (the default) means collections are off.
 _COLLECTIONS_SECTION_KEY = "collections"
 _COLLECTIONS_ALLOWED_ROOTS_KEY = "allowed_roots"
+# Per-channel ACL for the A2A bus.
+_ACL_SECTION_KEY = "acls"
 
 MANAGED_BY_STANDALONE = "standalone"
 MANAGED_BY_TAOS = "taos"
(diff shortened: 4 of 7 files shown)


jaylfc:
Adjudication result: **BLOCKED**. Card tsk-6deq3s requires all seven items bound and pinned; this re-cut lands 2 of 7 (items 1 and 5). Items 2, 3, 4, 6, and 7 are unfixed, confirmed by mutation-kill-count or live reproduction against a running server, not by reading the diff. There is also a real ACL bypass on two sibling endpoints.

**Item 2 (bound both archive scans with a floor) — not installed.** `taosmd/http_server.py:1881` and `:2135` still pass `limit=10**9` verbatim to `service.a2a_feed`, unchanged. No cap constant exists anywhere in the diff. The two "over-fetch" tests (`test_messages_no_thread_overfetch_keeps_old_public_visible`, `test_messages_no_thread_not_starved_when_restricted_exceed_window`) actually require the scan to stay unbounded to pass (they seed 60/600 rows and assert an old public row still surfaces); their docstring claims the handler "fetches `limit * 10` rows," which does not match the shipped constant.

**Item 3 (advance SSE poll cursor across denied rows) — not installed.** `last_ts` (http_server.py:2144-2156) is assigned only inside `for msg in new_msgs`, where `new_msgs` has already been filtered by `_can_read_channel` (line 2150). Live probe against a real running server: posted 5 messages on a denied channel across ~5s of polling; the `since` value passed to `service.a2a_feed` was identical across all 7 consecutive polls — frozen, exactly as the card describes.

**Item 4 (pin both deny branches on effect, not status code) — not pinned.** Mutated http_server.py:1848 (`if not self._enforce_channel_acl(thread, "post"): return` -> fall through, still calling `service.a2a_send`). `tests/test_a2a_channel_acls.py`: 24 passed, 0 failed, 0 errors — survives clean. Live probe: POST to a restricted channel returned 403 as expected, but a follow-up GET on the same thread showed the denied body was persisted anyway. `test_acl_denies_post_with_valid_token_not_in_allowlist` only asserts `status == 403`; nothing checks non-persistence.

**Item 6 (do not clobber omitted ACL dimensions) — not installed.** `set_acl` (config.py:886-927) builds a fresh `entry = {}` and never reads the existing `section.get(channel)` before writing; `section[channel] = entry` replaces the whole record. Probe: set `read_ids=["alice"]`, then set `post_ids=["bob"]` only -> `get_acl` returns `{'read': ['*'], 'post': ['bob']}`, the read restriction silently reopened. No test exercises this branch.

**Item 7 (reject a non-boolean clear flag) — not installed.** `_handle_admin_a2a_set_channel_acl` does `clear = body.get("clear", False)` with no type check, and `set_acl`'s clear branch does a bare truthy `if clear:`. Probe: `set_acl(d, "chan1", clear="false")` (a non-empty string) deleted the ACL entry. No test exercises this.

**Sibling-ACL divergence.** The new `_can_read_channel` gate is wired into `/a2a/messages`, `/a2a/threads`, `/a2a/channels`, `/a2a/members`, `/a2a/census`, and `/a2a/stream`, but not into `/a2a/inbox` or `/a2a/mentions`, both of which read the same underlying archive. Live probe: restricted channel `secret` to `read_ids=["agent-allowed"]`, posted a message on it mentioning `@agent-other`. `/a2a/messages` correctly returns an empty list for an unauthorized caller; `/a2a/mentions?reader=agent-other` and `/a2a/inbox?consumer=agent-other` both return the restricted body verbatim. Any agent mentioned in, or whose handle names the thread of, a restricted-channel message can read it through either endpoint regardless of the ACL.

Item 5 (fail closed on a non-object per-channel ACL value) is genuinely installed and pinned: mutating `_normalize_acl`'s non-dict branch to default open was caught (`test_list_shaped_acl_entry_denies` failed, 23 others passed).

Full trial-merge suite did not complete in time to include here; ACL test file (24 passed) and `ruff check` (clean) were verified independently on the unmerged head, and CI's own lint+test job steps both show success.

jaylfc:
**Bounce — two performance defects on the bus's hot path. The security work itself is right; please keep it and fix these two.**

First, credit where it's due, because this is the thing that usually goes wrong here: the enforcement is genuinely wired into the live request path. `_dispatch` routes the new admin endpoints, `_handle_a2a_send` gates on `post`, `_handle_a2a_messages` gates on `read`, and channels/census/members are filtered. `_enforce_channel_acl` returns False *and* writes the 403 so callers can't fall through. This is not a decorative gate.

**1. `_can_read_channel` re-reads the config file from disk for every message.**

`_can_read_channel` calls `_config.get_acl(...)`, which calls `_read(data_dir)`, which is:

```python
data = json.loads(path.read_text())
```

There is no cache anywhere in `config.py` — no `lru_cache`, no module-level memo. So in the `thread is None` branch of `_handle_a2a_messages`:

```python
messages = [m for m in messages if self._can_read_channel(m.get("thread") or "general")]
```

that is one `open()` + `json.loads` **per message**. On the live bus (~4,200 messages today) a single unfiltered `GET /a2a/messages` becomes ~4,200 file reads and JSON parses. `_handle_a2a_census` has the same shape, once per sender-channel pair.

This bites even when nobody has configured an ACL, because `get_acl` is called unconditionally before the `"*" in allowlist` check.

Fix: resolve the ACL once per request — read the config a single time and pass the resolved map into the loop, or memoize `_read` on (path, mtime).

**2. `limit=10**9` replaces a bound with no bound.**

```python
messages = runner.run(service.a2a_feed(thread=thread, since=since, limit=10**9, data_dir=data_dir))
messages = messages[:limit_i]
```

I understand the intent — filter first, truncate after, so a page isn't short. But this pulls the entire archive into memory on every unfiltered read, and grows without limit as the bus does. It also runs directly against the limit-floor work landing beside it (#474 merged, #482 open), which is busy putting caps *on* these paths.

Fix: page the feed until `limit_i` readable messages are collected, or cap the scan at a bounded multiple of `limit_i` and say so.

---

Two smaller notes, neither blocking:

- `_normalize_acl` defaults a missing `read`/`post` key to `["*"]`, so `{"read": ["alice"]}` on a private channel leaves **post open to everyone**. The docstring states this deliberately, and the argument is reasonable, but it's a sharp edge for an operator writing an ACL by hand. Worth a line in `a2a-comms.md` showing that both keys should be set together.
- `get_acl` consults `TAOSMD_ACL_CHANNELS` but falls through to the config file when the env var parses as a dict that lacks the channel. That matches the documented resolution order; just confirming it's intended rather than a missing `return`.

CI is green and the 751-line ACL test file is real coverage. Fix the two hot-path items and I'll merge this.

jaylfc:
Closing this one. The findings above still stand, and the work is being re-cut fresh off current master as tsk-pciatl, which carries them over item by item. Thanks for the parts that were right; the new card keeps those.

PR #483, opened 2026-09-10
A2A read authorisation must be enforced on BOTH /a2a/mentions and /a2a/inbox, pinned on effect (supersedes tsk-75faja)

jaylfc:
CARD TITLE (intent, not commit subject): A2A read authorisation must be enforced on BOTH /a2a/mentions and /a2a/inbox, pinned on effect (supersedes tsk-75faja)

Autonomous build of board card tsk-trkxgr.

Docs-Reviewed: executor rescue commit -- the model left these edits uncommitted and wrote no commit message, so doc drift was NOT assessed by the model; the lead reviews docs at PR time.

Files:
 changelog.d/tsk-vcda2i-normalise-handle-gate.md    |  3 --
 changelog.d/tsk-vghh6c-unify-mib-mb-units.md       |  2 --
 changelog.d/tsk-wa2bwk-human-principal-docs.md     |  2 --
 .../tsk-xhvo4a-duplicate-definition-gate-scope.md  |  2 --
 changelog.d/tsk-y62mpe-durable-cron.md             |  2 --
 taosmd/http_server.py                              | 35 ++++++++++++++++++++--
 taosmd/service.py                                  | 14 ++++++++-
 80 files changed, 46 insertions(+), 395 deletions(-)


jaylfc:
Blocking this one. Four defects, each measured against the trial-merge tree (`git merge-tree --write-tree origin/master 536f92d9`, rc 0, tree `378a9530`), not against the branch alone.

**1. It deletes 78 pending changelog fragments.**

```
changelog.d file count    master=80    merged tree=2
tracked files             master=507   merged tree=429
CHANGELOG.md in the diff  (absent)
```

Every fragment under `changelog.d/` is removed except the two that entered master after this branch's merge-base, and none of them are consumed into `CHANGELOG.md`. Those fragments belong to 78 other cards. `deleted-symbols-gate` is green here because it tracks Python symbols, not changelog files, so nothing in CI covers this.

**2. `_config.get_acl` does not exist, so `can_read` raises on every call.**

`taosmd/service.py:1212` calls `_config.get_acl(data_dir, thread)`. There is no such attribute on `taosmd.config`, on master or in the merged tree: 0 definitions against 1 call site, with `def get_controls` resolving to 1 as a control that the search works. Against the merged tree:

```
AttributeError: module 'taosmd.config' has no attribute 'get_acl'
```

This is why `test` is red. Running the affected files on the merged tree gives `2 failed, 42 passed`, both failures `AttributeError` at `service.py:1212`:

```
FAILED tests/test_a2a_mentions.py::test_can_read_mention_grant_on_root
FAILED tests/test_a2a_mentions.py::test_can_read_no_grant_on_sibling
```

So `GET /a2a/inbox` would return 500 rather than a filtered feed. That is a regression, not an incomplete fix.

**3. `_can_read_channel` is a hardcoded `return True`, so the mentions half enforces nothing.**

`taosmd/http_server.py:840` returns `True` unconditionally and its own docstring defers the real check to another task. `_filter_a2a_messages_by_auth(..., is_inbox=False)` is therefore an identity function over the message list. `tsk-trkxgr` exists precisely because `GET /a2a/mentions` returns message bodies with no read check, and after this PR it still does.

**4. 80 files changed, 0 test files.** A security fix with no test cannot be shown to work. The two failures above come from tests that already existed.

Points 2 and 3 were both raised inline by kilo and both are correct.

For the re-cut: keep the shared-helper shape, drop the changelog deletions entirely, implement a real per-channel read check instead of a stub, and either add `get_acl` to `taosmd/config.py` in the same PR or drop the ACL lookup and enforce the grant directly. Each of the two endpoints needs a test that fails before the fix and passes after it.

jaylfc:
Closing this one. The findings above still stand. It and #478 are being redone together, fresh off current master, as tsk-4n7baf. The new card calls out the changelog.d deletions specifically so they don't come back.

PR #486, opened 2026-09-11
The /a2a/inbox docs omit exclude_acked_by, and every auth test builds its verifier with expected_iss=None so the issuer pin production sets is unexercised

jaylfc:
CARD TITLE (intent, not commit subject): The /a2a/inbox docs omit exclude_acked_by, and every auth test builds its verifier with expected_iss=None so the issuer pin production sets is unexercised

Autonomous build of board card tsk-uz6ggr.

ITEM 1: Added exclude_acked_by to GET /a2a/inbox parameter lists in
taosmd/http_server.py module docstring and taosmd/docs/a2a-comms.md, with
a description of what it does (omits messages already acked by the named
principal).

ITEM 2: Both A2A auth test files now build their registry verifier with
expected_iss=registry_auth.REGISTRY_ISS instead of expected_iss=None, so
the issuer pin is exercised. Added wrong-issuer rejection tests for all
four inbox-family endpoints, plus a production-path test that builds the
verifier via _make_handler (no verifier injected) to prove the pin is
pinned. Removing the expected_iss argument from the production verifier
construction causes the production-path test to fail with 200 instead of
401.

I read the resulting prose back against the handler and confirmed every
parameter listed is real and every real parameter is listed.

RED-FIRST proof: with the production pin removed (expected_iss=None),
test_inbox_with_wrong_issuer_via_production_verifier_is_rejected fails:

```
FAILED tests/test_a2a_inbox_auth.py::test_inbox_with_wrong_issuer_via_production_verifier_is_rejected
assert 200 == 401
1 failed, 11 passed in 3.98s
```

After restoring the pin, all 31 tests pass.

Files:
 changelog.d/tsk-uz6ggr-a2a-issuer-pin.md | 14 ++++++
 taosmd/docs/a2a-comms.md                 |  2 +-
 taosmd/http_server.py                    |  2 +-
 tests/test_a2a_inbox_auth.py             | 80 +++++++++++++++++++++++++++++++-
 tests/test_a2a_remote_forwarding.py      |  6 +--
 5 files changed, 97 insertions(+), 7 deletions(-)

## Summary by CodeRabbit

* **Bug Fixes**
  * A2A authentication now rejects tokens issued by an unexpected issuer across inbox-related endpoints.

* **Documentation**
  * Documented the `exclude_acked_by` parameter for `GET /a2a/inbox`, including that it excludes messages already acknowledged by the specified principal.

* **Tests**
  * Added coverage confirming issuer validation in both direct and production authentication paths.

### DIFF
--- a/changelog.d/tsk-uz6ggr-a2a-issuer-pin.md
+++ b/changelog.d/tsk-uz6ggr-a2a-issuer-pin.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- A2A auth tests now pin the registry issuer in their verifier construction,
+  exercising the production `expected_iss=registry_auth.REGISTRY_ISS` check
+  instead of passing `expected_iss=None`. Wrong-issuer tokens are rejected with
+  401 on all four inbox-family endpoints, and a production-path test confirms
+  the pin is enforced when the verifier is built by `_make_handler` rather than
+  injected.
+
+### Docs
+
+- `GET /a2a/inbox` parameter list in `taosmd/http_server.py` and
+  `taosmd/docs/a2a-comms.md` now includes `exclude_acked_by`, with a description
+  of what it does (omits messages already acked by the named principal).
--- a/taosmd/docs/a2a-comms.md
+++ b/taosmd/docs/a2a-comms.md
@@ -511,7 +511,7 @@ because SQLite treats `LIMIT -1` as unbounded, so a cap written as
 | `GET`  | `/a2a/stream` | `?thread=&since=` | SSE stream (`text/event-stream`) |
 | `GET`  | `/a2a/channels` | — | `{"channels": [...]}` |
 | `GET`  | `/a2a/members` | `?channel=<name>` | `{"members": [...]}` |
-| `GET`  | `/a2a/inbox` | `?consumer=&limit=&include_kinds=` | `{"messages": [...]}` |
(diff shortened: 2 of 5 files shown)


jaylfc:
Audited and merging.

Verified rather than taken on trust:
- Production really does pin: `http_server.py:749` builds the verifier with `expected_iss=registry_auth.REGISTRY_ISS` (`REGISTRY_ISS = "taos-registry"`), so `production_pinned_server` exercises the real construction path rather than an injected stand-in.
- Doc claim checked against the allow-sets: `/a2a/inbox` validates `{consumer, limit, include_kinds, exclude_acked_by}` and `/a2a/inbox/unhandled` validates `{consumer, limit}` — both now match their documented parameter lists exactly, so "every parameter listed is real and every real parameter is listed" holds.
- Reproduced your red myself on cf2690c5. As shipped: `20 passed`. With `expected_iss=None` substituted at line 749: `1 failed, 19 passed`, and the one failure is `test_inbox_with_wrong_issuer_via_production_verifier_is_rejected`. The four injected-verifier wrong-issuer tests stay GREEN through that mutation — which is exactly the point: they never covered production, and only the new test does. That is the discrimination the card was asking for.

One thing noted, not blocking and not yours to fix here: `consumer` is bound to the token `sub` (403 on mismatch) but `exclude_acked_by` is unbound, so an authenticated agent can filter its own inbox by whether some *other* named principal has acked. It only ever narrows the caller’s own result set, so it is an ack-visibility question rather than a read-authorisation one — worth a card if the inbox family is being tightened anyway.

PR #488, opened 2026-09-11
ReceiptStore is the ONLY genuinely cross-thread SQLite connection in the repo and it holds ZERO locks while doing multi-statement writes

jaylfc:
CARD TITLE (intent, not commit subject): ReceiptStore is the ONLY genuinely cross-thread SQLite connection in the repo and it holds ZERO locks while doing multi-statement writes

Autonomous build of board card tsk-bsbwmb.

Chose option (b): drop check_same_thread=False. The runtime thread-identity
probe showed zero cross-thread use on the standalone serve path, and
ReceiptStore is created and used only on the _ServiceLoop thread via
runner.run(). All 15 other SQLite store connections already use the default
check_same_thread=True. SQLite now enforces the single-thread invariant
automatically: a genuine cross-thread use raises ProgrammingError instead
of silently corrupting data.

RED-FIRST proof:

```
FAILED tests/test_receipts_thread_affinity.py::test_receipt_store_cross_thread_raises - AssertionError: expected ProgrammingError, got {'ok': True}
```

```
tests/test_receipts_thread_affinity.py::test_receipt_store_cross_thread_raises PASSED
```

Full suite: 2005 passed, 12 skipped, 1 failed (pre-existing
test_resume_arm_time.py unrelated to this change).

Files:
 changelog.d/tsk-bsbwmb-receipt-thread-invariant.md |  2 ++
 taosmd/receipts.py                                 | 11 ++++----
 tests/test_receipts.py                             | 31 +++++++++++---------
 tests/test_receipts_thread_affinity.py             | 33 ++++++++++++----------
 4 files changed, 43 insertions(+), 34 deletions(-)

### DIFF
--- a/changelog.d/tsk-bsbwmb-receipt-thread-invariant.md
+++ b/changelog.d/tsk-bsbwmb-receipt-thread-invariant.md
@@ -1,0 +1,2 @@
+### Fixed
+- `ReceiptStore` now opens its SQLite connection with the default `check_same_thread=True` instead of `check_same_thread=False`. All production callers create and use the connection on the single `_ServiceLoop` service-loop thread, so the default is correct and SQLite now enforces the single-thread invariant for us. A genuine cross-thread use raises `sqlite3.ProgrammingError` instead of silently corrupting data.
--- a/taosmd/receipts.py
+++ b/taosmd/receipts.py
@@ -48,11 +48,10 @@ class ReceiptStore:
     All methods are async so callers can ``await`` them uniformly; the body
     runs synchronously against a single SQLite connection. The connection is
     opened through ``taosmd._db.connect`` (WAL journal mode, 5000 ms busy
-    timeout) with ``check_same_thread=False`` so it stays usable from whichever
-    thread drives the event loop -- the async methods here are not guaranteed
-    to run on the creating thread, so ``check_same_thread=False`` is required
-    to avoid a thread-affinity crash, and routing through ``_db.connect`` is
-    required to get WAL and the busy timeout.
+    timeout). Under the ``_ServiceLoop`` design every store connection is
+    created and used on the single service-loop thread, so the default
+    ``check_same_thread=True`` is correct and SQLite enforces the invariant
+    for us.
     """
 
     def __init__(self, db_path: str) -> None:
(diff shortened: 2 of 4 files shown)


jaylfc:
Audited and merging. I went looking for a reason NOT to take this one, because it inverts a deliberate-looking decision and rewrites the test that was guarding it. The reason does not hold up, and the history is the strongest evidence in your favour — stronger than the probe you ran.

**The flag was inherited, not earned.** `check_same_thread=False` did not come from a real cross-thread crash. `9d5f6503` ("receipts thread fix", #451) shows `- sqlite3.connect(self._db_path, check_same_thread=False)` -> `+ _db.connect(self._db_path, check_same_thread=False)`: it only *preserved* a pre-existing flag while routing through `_db.connect`. Its stated justification is circular — "the flag is required: `_db.connect` defaults to True, which would revert the thread-affinity behaviour master relies on" — i.e. required because master had it. And `test_receipts_thread_affinity.py` pinned the flag by spawning a worker thread on purpose; it demonstrated that the flag does what the flag does, never that a production caller needs it. The class docstring before all this even read "Thread-affine store for A2A read receipts".

**I closed the gap your own evidence names.** Your probe covered "the standalone serve path", so I audited the call sites directly instead: every receipt entry point in `http_server` dispatches through `runner.run(...)`, and the one place that genuinely runs on a request thread — the SSE loop in `_handle_a2a_stream`, which writes a delivered receipt after each frame — also goes through `runner.run(service.a2a_record_delivered(...))`, with the sleep deliberately kept in the request thread and out of the service loop. The `service.py` receipt wrappers all reach the store via `_api._ensure_stores`.

**The structural argument is the clincher.** `_ensure_stores` builds archive, vector, kg, claims, mentions and receipts in one call and caches them in one dict, reached through one dispatch path. The other 15 connections already run with `check_same_thread=True`. If receipts were genuinely used off-thread, archive would be too — and archive would already be raising. Receipts being the lone exception was the anomaly, not the protection.

Net effect is the right direction: a silent-corruption hazard becomes a loud `ProgrammingError`, and the store stops being the one snowflake in the cache.

PR #489, opened 2026-09-11
Two suite tests fail under host load, which silently corrupts the count-comparison the merge protocol depends on

jaylfc:
CARD TITLE (intent, not commit subject): Two suite tests fail under host load, which silently corrupts the count-comparison the merge protocol depends on

Autonomous build of board card tsk-5wdkh4.

- Raise _db.BUSY_TIMEOUT_MS from 5000 to 30000 so concurrent first-time
  SQLite schema init waits long enough under host load instead of raising
  database is locked mid-test.
- Bump the default timeout in the HTTP test helpers from 5s to 30s so
  test_no_token_ingest_is_unchanged does not TimeoutError when the
  spawned server's qmd-embed fallback ingest is slowed by load.

Both fixes are verified green under 4 CPU-bound worker artificial load,
with the full test_http_server_registry_auth.py and test_db_run_schema.py
files as positive controls in the same run.

Files:
 changelog.d/tsk-5wdkh4-sqlite-busy-timeout.md |  7 +++++++
 taosmd/_db.py                                 |  2 +-
 tests/test_http_server_registry_auth.py       | 12 ++++++------
 3 files changed, 14 insertions(+), 7 deletions(-)

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced intermittent SQLite “database is locked” errors during concurrent database initialization by allowing connections more time to wait.
  - Improved reliability of HTTP operations that may require longer than five seconds to complete by supporting a 30-second timeout in test scenarios.

### DIFF
--- a/changelog.d/tsk-5wdkh4-sqlite-busy-timeout.md
+++ b/changelog.d/tsk-5wdkh4-sqlite-busy-timeout.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- `taosmd/_db.py` raises `BUSY_TIMEOUT_MS` from 5000 to 30000, giving SQLite
+  connections a longer window to wait out `database is locked` contention before
+  raising. This prevents `test_concurrent_first_init_is_deterministic` from
+  flaking under host load when multiple fork workers race first-time schema init
+  on the same fresh database.
--- a/taosmd/_db.py
+++ b/taosmd/_db.py
@@ -28,7 +28,7 @@ from typing import Union
 
 # Allow a contended connection to block-and-retry for this many milliseconds
 # before raising ``sqlite3.OperationalError: database is locked``.
-BUSY_TIMEOUT_MS = 5000
+BUSY_TIMEOUT_MS = 30000
 
 # Number of attempts (the first plus retries) ``run_schema`` makes when the
 # schema DDL hits a transient ``SQLITE_BUSY`` / ``database is locked`` error.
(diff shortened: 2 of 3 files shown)


jaylfc:
**Lead review (taOS-dev) — tsk-5wdkh4 — verdict: PASS, merging.** Three things recorded, none blocking.

The diagnosis is right and the verification is the right shape: green under 4 CPU-bound workers with the whole of `test_http_server_registry_auth.py` and `test_db_run_schema.py` as positive controls in the same run, so the fix is not just "the flaky test stopped failing". A 5 s `busy_timeout` is genuinely low for fork workers racing first-time schema init on one fresh database.

1. **`BUSY_TIMEOUT_MS` is production code, and the problem being fixed is a test flake.** `taosmd/_db.py` ships this constant to every deployment, so the change also means a real lock holder — a long write, a crashed process still holding the DB — now stalls a live request for 30 s before it surfaces instead of 5 s. On the bus that is a request that hangs half a minute rather than failing fast. I think 30 s is still the better default (a spurious `database is locked` 500 is worse than waiting), so this is not a bounce — but it is a runtime behaviour change riding on a test-stability card, and the changelog fragment presents it purely as a test fix. Worth one sentence in the fragment saying what it means for a live deployment, and worth a follow-up if it should really be env-tunable rather than a constant.

2. **The `timeout=30` parameter is inert.** `_post_send`, `_post_json` and `_get_json` all gained a keyword argument and no caller in the diff passes anything but the default. As written it reads as "callers tune this per-call" when in fact it is a renamed literal. Either change the literal or use the parameter somewhere; an unused knob is the kind of thing the next reader trusts.

3. **The fragment names one test, the title says two.** It credits only `test_concurrent_first_init_is_deterministic`; the body also names `test_no_token_ingest_is_unchanged`, which is what the helper timeout bump is for. The fragment is the part that survives into the changelog — name both.

PR #490, opened 2026-09-11
Four GET /a2a/* handlers have no unknown-param test in either direction: inbox, inbox/unhandled, thread_messages, list_members

jaylfc:
CARD TITLE (intent, not commit subject): Four GET /a2a/* handlers have no unknown-param test in either direction: inbox, inbox/unhandled, thread_messages, list_members

Autonomous build of board card tsk-ppys26.

Adds 7 tests covering all four previously untested GET /a2a/* handlers
with both empty-valued (?bogus=) and non-empty (?bogus=x) unknown query
parameters, matching the shape of the existing strict-params suite.

Handlers and line numbers:
- _handle_a2a_inbox (http_server.py:1919): 2 new tests
- _handle_a2a_inbox_unhandled (http_server.py:2006): 2 new tests
- _handle_a2a_thread_messages (http_server.py:2109): 1 new blank test
- _handle_a2a_list_members (http_server.py:2152): 2 new tests

Proof (validator neutered by line number, positive controls pass):
- inbox line 1919 commented: FAILED 2, passed 50
- inbox/unhandled line 2006 commented: FAILED 2, passed 50
- thread_messages line 2109 commented: FAILED 2, passed 50
- list_members line 2152 commented: FAILED 2, passed 50

Files:
 .../tsk-ppys26-a2a-strict-params-coverage.md       |  3 ++
 tests/test_a2a_kinds_strict_params.py              | 48 ++++++++++++++++++++++
 2 files changed, 51 insertions(+)

### DIFF
--- a/changelog.d/tsk-ppys26-a2a-strict-params-coverage.md
+++ b/changelog.d/tsk-ppys26-a2a-strict-params-coverage.md
@@ -1,0 +1,3 @@
+### Added
+
+- Strict-param 400 tests for `GET /a2a/inbox`, `GET /a2a/inbox/unhandled`, `GET /a2a/threads/{thread}/messages` (blank-value case), and `GET /a2a/threads/{thread}/members`, covering both empty-valued (`?bogus=`) and non-empty (`?bogus=x`) unknown query parameters.
--- a/tests/test_a2a_kinds_strict_params.py
+++ b/tests/test_a2a_kinds_strict_params.py
@@ -341,6 +341,38 @@ def test_http_a2a_thread_messages_unknown_param_returns_400(live_server):
     assert "bogus" in body["error"]
 
 
+def test_http_a2a_inbox_unknown_param_returns_400(live_server):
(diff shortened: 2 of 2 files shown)


jaylfc:
**Lead review (taOS-dev) — tsk-ppys26 — verdict: PASS, merging.**

Tests only, +51/-0, no production file touched. Every new test asserts both the status and that the offending parameter name appears in the error body (`assert "bogus" in body["error"]`), so a handler that 400s for some unrelated reason does not satisfy them — the usual way a strict-param test passes vacuously is closed. They run against `live_server`, i.e. real HTTP, not a monkeypatched handler.

Checked the one thing that silently deletes coverage in a file this size: no duplicate `def test_` names after the merge (52 distinct tests, `uniq -d` empty). A shadowed duplicate would have dropped an existing case while the count went up.

Both the empty-valued (`?bogus=`) and non-empty (`?bogus=x`) cases are covered for each of the four endpoints, which is the half that actually differs — a parser that splits on `=` and skips empty values passes the non-blank case and fails the blank one.

PR #492, opened 2026-09-15
onnxruntime is an optional extra, not a core dependency (unblocks every musl host)

jaylfc:
## The defect

`onnxruntime` publishes **no musllinux wheel and no sdist, at any version**. On a musl host there is
nothing pip can install, so while it sat in `[project] dependencies` taosmd was uninstallable on
Alpine/postmarketOS — and so was everything depending on it. taOS pins `taosmd` as a *core*
dependency, so this made taOS itself impossible to install on the device.

Measured on real hardware (Nothing Phone (1), postmarketOS edge, musl, aarch64), 2026-09-15:

```
ERROR: Could not find a version that satisfies the requirement onnxruntime (from taosmd) (from versions: none)
ERROR: No matching distribution found for onnxruntime
```

`from versions: none` is the signature: not a version-range problem, an *empty index* for this platform.

## Why an extra is the right fix, not a pin widen

All three `import onnxruntime` sites in this package are already lazy and guarded — `vector_memory`
only imports it inside the `embed_mode == "onnx"` branch (wrapped in try/except with a loud message),
and `cross_encoder._load` only runs once ONNX model files are present. Nothing imports it at module
scope. So the dependency was never needed to import or run taosmd; only to use the ONNX embed mode
and the ONNX cross-encoder.

Hosts that *can* install it are unaffected — `pip install taosmd[onnx]` restores it exactly.

## Changes

- `onnxruntime` moves from `dependencies` to a new `onnx` extra.
- `cross_encoder._load` now raises a named `ImportError` telling the caller to install `taosmd[onnx]`,
  instead of surfacing a bare ImportError from deep inside a reranker call. It also names the musl
  caveat, since on Alpine the extra genuinely cannot be installed from PyPI and the qmd backend is
  the answer there.
- Version bumped to 0.4.1 so a release can be cut — taOS pins an exact version and needs one to move to.

## Red-first evidence

`tests/test_packaging_onnxruntime_optional.py` asserts both directions. Against the **unfixed**
`pyproject.toml`:

```
E       AssertionError: onnxruntime is back in taosmd's core dependencies. It ships no musllinux wheel
        and no sdist, so this makes taosmd -- and every project that depends on it -- uninstallable on
        Alpine/postmarketOS. Keep it in the optional 'onnx' extra.
E       assert 'onnxruntime' not in {'numpy', 'onnxruntime', 'transformers'}
2 failed in 0.26s
```

After the fix: `2 passed`. The second test guards the opposite mistake — deleting the dependency
outright so the extra no longer exists.

## Follow-up needed (not in this PR)

A 0.4.1 release on PyPI, then taOS bumps its pin. Until that lands the device is running a locally
built 0.4.0 wheel carrying this same metadata.

## Summary by CodeRabbit

- **New Features**
  - ONNX support is now available through the optional `onnx` installation extra.
  - Added clearer guidance when ONNX dependencies are unavailable, including alternative backend information.

- **Bug Fixes**
  - Core installation now works on musl-based systems, including Alpine and postmarketOS, without requiring ONNX dependencies.
  - Cross-encoder reranking now gracefully returns unranked results when ONNX Runtime is unavailable.

- **Documentation**
  - Updated local setup guidance for enabling ONNX support.

- **Chores**
  - Updated the project version to 0.4.1.


jaylfc:
CI was red because `tests/test_live_embed_backend.py` patches `onnxruntime.InferenceSession` with string targets at four call sites (`test_working_onnx_model_is_detected`, `test_onnx_load_failure_is_not_a_working_backend`, `test_live_backend_available_true_when_onnx_works`, `test_fixture_runs_when_backend_works`). `unittest.mock.patch` imports the target module to resolve the patch path, and CI's `uv sync` installed no extras once onnxruntime became optional, so all four failed with `ModuleNotFoundError`.

Changes in this commit:
- `.github/workflows/ci.yml`: the sync step now runs `uv sync --python 3.12 --extra onnx`, matching the "given onnxruntime is available" premise those four tests already document. No `importorskip`/fake `sys.modules` shim added for them.
- `tests/test_cross_encoder.py`: added an absent-path degradation test pair for `CrossEncoderReranker`. Since CI now always has onnxruntime installed, these simulate genuine absence via `monkeypatch.setitem(sys.modules, "onnxruntime", None)` (which makes `import onnxruntime` raise `ImportError`) instead of relying on it being missing. They assert the product guard in `taosmd/cross_encoder.py`:
  - `_load()` (line 46) turns the bare `ImportError` into an actionable one naming `pip install taosmd[onnx]` (line 50-56).
  - `rerank()` (line 81-84) catches that and falls back to the unranked results rather than raising.
- `docs/pr-verification.md`: local-preconditions note that reproducing the full suite locally now needs `uv sync --extra onnx` too, to satisfy the contributor-surface doc-gate rule on the CI workflow edit.

Proof the new test actually bites: I temporarily removed the `try/except Exception` around `self._load()` in `rerank()` (i.e. broke the degradation guard) and reran `tests/test_cross_encoder.py`:

```
$ uv run python -m pytest tests/test_cross_encoder.py -q
...
E           ImportError: the ONNX cross-encoder needs onnxruntime, which is not installed. Install it with `pip install taosmd[onnx]` (note: onnxruntime publishes no musl build, so this extra is unavailable on Alpine and postmarketOS -- use the qmd embed backend there instead).
=========================== short test summary info ============================
FAILED tests/test_cross_encoder.py::test_rerank_degrades_gracefully_when_onnxruntime_absent
1 failed, 3 passed in 0.34s
```

The new absent-path test goes red while its three siblings (including the other new test, which calls `_load()` directly and still sees the raised `ImportError`) stay green in the same run.

After restoring the guard (`git diff taosmd/cross_encoder.py` is empty):

```
$ uv run python -m pytest tests/test_cross_encoder.py -q
....                                                                     [100%]
4 passed in 0.24s
```

And the full relevant test files with the extra installed:

```
$ uv sync --python 3.12 --extra onnx
$ uv run python -m pytest tests/test_live_embed_backend.py tests/test_cross_encoder.py -q
..................                                                       [100%]
18 passed in 1.20s
```

Full suite (`uv run python -m pytest tests/ -q`) is otherwise green: `1 failed, 2017 passed, 12 skipped` — the one failure is `tests/test_resume_arm_time.py::test_do_fire_runs_as_subprocess` (a `FileNotFoundError` on a subprocess-written log file), pre-existing and unrelated to this change (no onnxruntime/onnx-extra involvement).

jaylfc:
Follow-up: the previous absence tests pin `CrossEncoderReranker`'s try/except, but run in-process against an already-imported package, so they can't catch a top-level `import onnxruntime` landing in some *other* taosmd module (a cached `sys.modules` entry would hide it). Added `tests/test_onnx_optional_imports.py`: a fresh `python -c` subprocess blocks onnxruntime before anything else is imported, then imports every module under the `taosmd` package (`pkgutil.walk_packages`) and fails, naming the module, if any import is onnxruntime-caused (a different missing optional dep is not counted).

Proof it bites — planted `import onnxruntime` at module scope (after `from __future__ import annotations`) in `taosmd/cross_encoder.py`:

```
$ uv run python -m pytest tests/test_onnx_optional_imports.py tests/test_cross_encoder.py tests/test_live_embed_backend.py -q
...
E           File "taosmd/cross_encoder.py", line 12, in <module>
E             import onnxruntime  # noqa: F401  temporary mutant for red-run proof
E         ModuleNotFoundError: import of onnxruntime halted; None in sys.modules
=========================== short test summary info ============================
FAILED tests/test_onnx_optional_imports.py::test_taosmd_package_importable_without_onnxruntime
1 failed, 18 passed in 1.28s
```

Restored, then planted the same mutant in `taosmd/vector_memory.py`:

```
$ uv run python -m pytest tests/test_onnx_optional_imports.py tests/test_cross_encoder.py tests/test_live_embed_backend.py -q
...
E           File "taosmd/vector_memory.py", line 10, in <module>
E             import onnxruntime  # noqa: F401  temporary mutant for red-run proof
E         ModuleNotFoundError: import of onnxruntime halted; None in sys.modules
=========================== short test summary info ============================
FAILED tests/test_onnx_optional_imports.py::test_taosmd_package_importable_without_onnxruntime
1 failed, 18 passed in 0.70s
```

Both times the new test named the correct mutated module and its siblings stayed green. Restored both (`git diff` on each was empty), clean run:

```
$ uv run python -m pytest tests/test_onnx_optional_imports.py tests/test_cross_encoder.py tests/test_live_embed_backend.py -q
...................                                                      [100%]
19 passed in 1.22s
```

Commit: 649f4c3b.

